PR #28, opened 2022-09-16
v1.12.0 - Fix stylelint errors.

kevinrodrigues:
### Added
- Code of conduct file

### Fixed
- `stylelint` errors.


apps/pie-microsite/src/assets/styles/components/_notification.scss
```diff
@@ -1,11 +1,12 @@
 @use '../fozzie-config' as f;
 
+$combinedIconAndPaddingWidth:       calc($icon-width + $notification-padding);
```
jamieomaguire, line 3: I didn't think it would be happy with this using other SCSS variables before they're defined?

kevinrodrigues: You're right. It errors (i skipped the hook). Reverted.

apps/pie-microsite/src/assets/styles/base/_layout.scss
```diff
@@ -3,19 +3,11 @@
...
-    grid-template-rows: 64px 1fr max-content;
+    grid-template: 64px 1fr max-content / 260px 1fr;
+    grid-template-areas: 'header header' 'nav main' 'nav footer';
 
     @include f.media('<=narrowMid') {
-        grid-template-areas:
-            "header header"
-            "main main"
-            "footer footer";
+        grid-template-areas: 'header header' 'main main' 'footer footer';
```
jamieomaguire, line 10: I personally disagree with this rule. I think we should override it as you lose the main benefit in my opinion which is an almost ascii-art style representation of the layout. Just my opinion though!

kevinrodrigues: I've not changed this rule it seems to come from the defaults. We could change it / override it so it allows breaks.

ashleynolan: I think this might be one of those weird stylelint areas where it's hard to get exactly what we want. I would have through this would be valid syntax with the rule you just changed though, with multi-line spacing being allowed?

kevinrodrigues: So this one is a different [declaration](https://github.com/justeat/stylelint-config-fozzie/blob/master/index.js#L36) in stylelint. The one I changed was around scss variables. I can unify it tho?

ashleynolan: Ah, ok. Yeh, may be good to take a look at it and make it the same potentially

kevinrodrigues: Added a fix here: https://github.com/justeat/stylelint-config-fozzie/pull/26

kevinrodrigues: Fixed.

jamieomaguire (approved):

ashleynolan (approved):

PR #31, opened 2022-09-22
pie-microsite@v1.16.1 - add CNAME file

jamieomaguire:
v1.16.1
*September 22, 2022*

### Added
- CNAME file to allow domain settings to persist between deployments


siggerzz (approved):

kevinrodrigues (approved):

apps/pie-microsite/.eleventy.js
```diff
@@ -21,6 +21,9 @@ module.exports = eleventyConfig => {
   // Shortcodes
   shortcodes.addAllShortCodes(eleventyConfig);
 
+  // Pass CNAME file into dist output (needed for domain to work correctly)
+  eleventyConfig.addPassthroughCopy("CNAME");
```
kevinrodrigues, line 25: Super minor but single quotes here. Also it might be good to start thinking about grouping the passthroughs / config changes soon so they are separated out into logical directories?

jamieomaguire: Sorry I only just noticed this and I completely agree - next change I can group them. Didn't actually spot the pre-existing one!

PR #53, opened 2022-11-25
PIE microsite @v1.23.0 - Create tokenTable

FayeCarter:
v1.23.0
*November 25, 2022*
### Added
- `tokenTable` shortcode to display design tokens as list
- design tokens to colour pages

## Video demo
https://user-images.githubusercontent.com/24740015/204016214-d68b4316-41a3-45b7-bc70-96d7664d7b39.mov


apps/pie-microsite/src/_11ty/shortcodes/tokensTable.js
```diff
@@ -0,0 +1,172 @@
+const pieDesignTokens = require('@justeat/pie-design-tokens/dist/tokens.json');
+const { stringHelpers, objectHelpers } = require('../../utilities/helpers');
+const tokenTypes = require('../../_data/tokenTypes');
+const { isColorDark } = require('../../utilities/colors');
+
+/**
+ * Creates the SCSS token name such as '$color-black'
```
ashleynolan, line 7:
```suggestion
 * Takes the token key and token type and
 * creates an SCSS token name such as '$color-black'
```

apps/pie-microsite/src/_11ty/shortcodes/tokensTable.js
```diff
@@ -0,0 +1,172 @@
...
+
+/**
+ * Creates a token pill element to display the SCSS token name in the token list item.
+ * @param {string} tokenScssName - the text to display
+ * @returns {string} the token pill HTML string
+ */
+const createTokenPill = tokenScssName => `<span class="c-tokensTable-token">${tokenScssName}</span>`;
+
+/**
+ * Creates a token list item element to to add to the tokens list.
```
ashleynolan, line 92:
```suggestion
 * Creates a token list item element to add to the tokens list.
```

apps/pie-microsite/src/_11ty/shortcodes/tokensTable.js
```diff
@@ -0,0 +1,172 @@
...
+        <span class="c-tokensTable-displayName">${tokenDisplayName}</span>
+        <span></span>
+        <span></span>
+      </div>
+      ${tokenPill}
+    </li>`;
+};
+
+/**
+ * Creates a list of design token list items and column headers to display above
```
ashleynolan, line 121:
```suggestion
 * Takes a list of design token list items and outputs these with appropriate column headers to be displayed above them
```

apps/pie-microsite/src/_11ty/shortcodes/tokensTable.js
```diff
@@ -0,0 +1,172 @@
...
+  <span>Example</span>
+  <span>Description</span>
+  <span>Token name</span>
+</div>
+<ul class="c-tokensTable-list">
+  ${listElements.join('')}
+</ul>`;
+
+/**
+ * Throws an error listing which, if any, parameters are missing/invalid from the configuration object
```
ashleynolan, line 135: This doesn't make sense atm – doesn't read quite right

apps/pie-microsite/src/assets/styles/components/_contentPage.scss
```diff
@@ -25,7 +25,7 @@ $content-max-width: 80ch;
 }
 
 .c-contentPage-header > *,
-.c-contentPage-content > *:not(.c-contentPage-img, hr, .c-contentPage-notification, section),
+.c-contentPage-content > *:not(.c-contentPage-img, hr, .c-contentPage-notification, section, div),
```
ashleynolan, line 28: This selector seems like it could be getting a bit unwieldy (just wondering if there might be a better way to do this potentially). Just worried that this not list could grow as it's already a bit hard to comprehend

apps/pie-microsite/src/assets/styles/components/_tokensTable.scss
```diff
@@ -0,0 +1,121 @@
...
+
+.c-tokensTable-list {
+    list-style: none;
+    padding: 0;
+    margin: 0;
+    width: 100%;
+
+    @include f.media('<=wide') {
+        margin-top: f.spacing(f);
+    }
```
ashleynolan, line 14: I'd probably make your media queries consistent as to what is the default. I'd probably go mobile-first and put this in the base selector and then have:

```suggestion
    @include f.media('>wide') {
        margin: 0;
    }
```

apps/pie-microsite/src/assets/styles/trumps/_utilities.scss
```diff
@@ -0,0 +1,7 @@
+@use '../fozzie-config' as f;
+
+.u-hideBelowOrAtWide {
```
ashleynolan, line 3: Would be good to get this into fozzie at some point (we have a lot of these type of [trumps here](https://github.com/justeat/fozzie-components/blob/master/packages/tools/fozzie/src/scss/trumps/_rwd.scss)).

apps/pie-microsite/src/assets/styles/components/_contentPage.scss
jamieomaguire: yeah fully agree - not quite sure the best approach if we want to keep all the max-widths declared together in this file

apps/pie-microsite/src/assets/styles/trumps/_utilities.scss
jamieomaguire: yep can do in the near future! Was surprised it wasn't already in there :D

apps/pie-microsite/src/assets/styles/components/_contentPage.scss
jamieomaguire: Maybe we could always agree that custom components on a MD page are wrapped in specific div/class and just exclude that?

jamieomaguire: or if its only going to apply to typographic elements, we could just list those elements

jamieomaguire (dismissed):

apps/pie-microsite/src/_11ty/shortcodes/tokensTable.js
```diff
@@ -0,0 +1,173 @@
...
+}) => {
+    const tokenPill = createTokenPill(tokenScssName);
+    const tokenExampleElement = createTokenExampleElement(token, tokenType);
+
+    return `<li class="c-tokensTable-row c-tokensTable-item">
+      ${tokenExampleElement}
+      <div class="c-tokensTable-content">
+        <span class="c-tokensTable-displayName">${tokenDisplayName}</span>
+        <span></span>
+        <span></span>
```
dandel10n, line 115: What are those empty spans for?

dandel10n (commented):
One tiny question, other than that looks great!

apps/pie-microsite/src/_11ty/shortcodes/tokensTable.js
FayeCarter: We had some copy in there initially but chose to take it out and move it into another PR.
These were left in as placeholders, but aren't needed. I've removed them now
Thanks 🙂

dandel10n (approved):

jamieomaguire (approved):

PR #56, opened 2022-12-08
Colour categories

jamieomaguire:


apps/pie-microsite/src/_11ty/shortcodes/tokensTable.js
```diff
@@ -1,4 +1,7 @@
+/* eslint-disable no-trailing-spaces */
 const pieDesignTokens = require('@justeat/pie-design-tokens/dist/tokens.json');
+const pieTokensMetadata = require('../../pieDesignTokensMetadata.json');
+const pieTokenCategories = require('../../pieDesignTokenCategories.json');
```
ashleynolan, line 4: I don't think we necessarily need to prefix all these with `pie`. I think they make as much sense without that. I would probably name these `designTokens`, `tokensMetadata` & `tokensCategories`

apps/pie-microsite/src/_11ty/shortcodes/tokensTable.js
```diff
@@ -130,6 +149,138 @@ const createTokensList = listElements => `<div class="c-tokensTable-row u-spacin
   ${listElements.join('')}
 </ul>`;
 
+/**
+ * Creates a tokens list for a given category
+ * @param {*} tokens 
+ * @param {string} path 
+ * @param {string} category 
+ * @param {string} tokenType 
+ * @returns
```
ashleynolan, line 158: Would be good to detail the type of thing this returns and brief comment for each param

apps/pie-microsite/src/_11ty/shortcodes/tokensTable.js
```diff
@@ -130,6 +149,138 @@ const createTokensList = listElements => `<div class="c-tokensTable-row u-spacin
...
+    // all 'chunks' of the tokens table page HTML in a single string
+    return categoryTokenLists.join('<hr />');
+};
+
+/**
+ * Gets all the metadata associated with tokens of a given type such as colour.
+ * @param {string} path 
+ * @returns {object}
+ */
+const getTokenTypeMetadata = path => objectHelpers.getObjectPropertyByPath(pieTokensMetadata, path);
```
ashleynolan, line 203: As this file is getting pretty big, I would look to split it out into separate files where logical.

Looking over it, it seems like there are creation/build functions and functions that interact directly with the data/metadata, so I'd probably split it based on that as a start to separate out a bit.

apps/pie-microsite/src/_11ty/shortcodes/tokensTable.js
```diff
@@ -130,6 +149,138 @@ const createTokensList = listElements => `<div class="c-tokensTable-row u-spacin
...
+const getTokensForCategory = (category, tokenTypeMetadata) => Object
+    .keys(tokenTypeMetadata)
+    .filter(token => tokenTypeMetadata[token].category === category);
+
+
+/**
+ * Gets all subcategory keys for a given parent category
+ * @param {object} tokenTypeCategories 
+ * @param {string} parentCategoryKey 
+ * @returns
```
ashleynolan, line 220: Need to add to the return here (to explain what is returned and in what format)

apps/pie-microsite/src/_11ty/shortcodes/tokensTable.js
```diff
@@ -130,6 +149,138 @@ const createTokensList = listElements => `<div class="c-tokensTable-row u-spacin
...
+ * @returns 
+ */
+const getSubcategoriesForParentCategory = (tokenTypeCategories, parentCategoryKey) => Object
+          .keys(tokenTypeCategories)
+          .filter(categoryKey => tokenTypeCategories[categoryKey].parentCategory === parentCategoryKey);
+
+/**
+ * Builds a list of tokens that are categorised by parent and subcategory
+ * @param {object} config 
+ * @returns {string}
```
ashleynolan, line 229: An example of the return would be useful – looks like it's a list of tokens separate by `<hr />` tags, but an example here would be useful.

apps/pie-microsite/src/_11ty/shortcodes/tokensTable.js
```diff
@@ -130,6 +149,138 @@ const createTokensList = listElements => `<div class="c-tokensTable-row u-spacin
...
+ */
+const buildCategoryListsWithParents = ({
+    parentCategories, path, tokenType, isGlobal, tokens
```
ashleynolan, line 232: As this function is fairly big, could probably benefit from some comments to break it up a little and explain each step.

apps/pie-microsite/src/_11ty/shortcodes/tokensTable.js
```diff
@@ -130,6 +149,138 @@ const createTokensList = listElements => `<div class="c-tokensTable-row u-spacin
...
+    return parentCategoryLists.join('<hr />');
+};
+
+
+/**
+ * Builds all of the token lists for a given token type
+ * @param {string} pathToTokens 
+ * @param {string} path 
+ * @param {string} tokenType 
+ * @returns {string}
```
ashleynolan, line 264: Would be good to have comments/examples for these params and return value

apps/pie-microsite/src/_11ty/shortcodes/tokensTable.js
```diff
@@ -130,6 +149,138 @@ const createTokensList = listElements => `<div class="c-tokensTable-row u-spacin
...
+ * @param {string} path 
+ * @returns {object}
+ */
+const getTokenTypeMetadata = path => objectHelpers.getObjectPropertyByPath(pieTokensMetadata, path);
+
+/**
+ * Gets all tokens for a given category such as 'orange'
+ * @param {string} category 
+ * @param {object} tokenTypeMetadata 
+ * @returns {string[]}
```
ashleynolan, line 209: Comments and example of the return value

apps/pie-microsite/src/_11ty/shortcodes/tokensTable.js
```diff
@@ -130,6 +149,138 @@ const createTokensList = listElements => `<div class="c-tokensTable-row u-spacin
...
+    });
+
+    // all 'chunks' of the tokens table page HTML in a single string
+    return categoryTokenLists.join('<hr />');
+};
+
+/**
+ * Gets all the metadata associated with tokens of a given type such as colour.
+ * @param {string} path 
+ * @returns {object}
```
ashleynolan, line 201: Comments and example of return value

apps/pie-microsite/src/_11ty/shortcodes/tokensTable.js
```diff
@@ -130,6 +149,138 @@ const createTokensList = listElements => `<div class="c-tokensTable-row u-spacin
...
+        tokenType,
+        tokenMetadata: tokenTypeMetadata[key]
+    }));
+
+    return buildTokensList(tokenListElements);
+};
+
+/**
+ * Builds a categorised list of tokens
+ * @returns {string}
```
ashleynolan, line 178: Again – an example of the string returned would be useful

apps/pie-microsite/src/assets/styles/components/_tokensTable.scss
```diff
@@ -88,8 +89,15 @@ $tokensTable-border-width: 1px;
 
     @include f.media('>=wide') {
         padding: f.spacing(e) 0;
-        border-width: 0 0 $tokensTable-border-width 0;
         border-radius: 0;
+        border-width: 0 0 $tokensTable-border-width 0;
+
+        // Remove bottom border for final item in non-section token lists and the final item in section token lists
```
ashleynolan, line 95: This comment doesn't make a huge amount of sense (as someone not familiar with the code).

I'm guessing it's something to do with the subcategories listings, but I'd make this comment more verbose and as clear as possible.

apps/pie-microsite/src/assets/styles/components/_tokensTable.scss
```diff
@@ -122,3 +130,35 @@ $tokensTable-border-width: 1px;
...
+        margin-top: f.spacing(i);
+        position: relative;
+        &:before {
+            content: '';
+            position: absolute;
+            left: 0;
+            right: 0;
+            top: -#{f.spacing(f)};
+            background-color: f.$color-border-default;
+            height: $tokensTable-border-width;
```
ashleynolan, line 153: Is this to apply a border? If so, would be useful to have a comment before the `&:before` just saying that (or what it is styling)

apps/pie-microsite/src/pieDesignTokenCategories.json
```diff
@@ -0,0 +1,104 @@
...
+        "content": {
+          "displayName": "Content",
+          "description": "Alias colour tokens used for text and icons only."
+        },
+        "reactive": {
+          "displayName": "Reactive",
+          "description": "Alias colour tokens used to represent state changes and loading animations in components."
+        }
+      },
+      "pageBackgrounds": {
```
ashleynolan, line 54: Just a thought around structure here. Would it make sense to group these inside the parentCategories?

I think it probably makes the list a bit more scannable then (not sure if it would a massive pain to rewrite some of the category logic though in this format).

So something like:

```json
"alias": {
      "parentCategories": {
        "elementary": {
          "displayName": "Elementary",
          "description": "Basic alias colour tokens for the basic building blocks of our products, including backgrounds, borders and dividers."
          "subCategories": [
              {
                 "pageBackgrounds": {
                     "displayName": "Page Backgrounds"
                 }
              },
              {
                 "containerBackgrounds": {
                    "displayName": "Container Backgrounds"
                 }
              },
              { "etc" }
          ]
        },
```

ashleynolan: Just looking at the metadata, it may be a nightmare to not have the categories all in the root of the alias list – if so, then obviously don't worry about it. Just thought it was a nice way of grouping them potentially, but don't want to make the logic more complicated in the code. Take a look and see what you think anyway!

FayeCarter: Do you mind if we leave this change until we add another token type? This way we can see how spacing/box-shadow affects all these files.

apps/pie-microsite/src/_11ty/shortcodes/tokensTable.js
FayeCarter: I've separated out the two types of functions. I think both of these will end up changing a fair amount as we add new token types so I haven't changed much at this point.

apps/pie-microsite/src/pieDesignTokenCategories.json
ashleynolan: Yeh, that's fine. Happy to leave it as-it-is for now

ashleynolan (approved):

FayeCarter (approved):

PR #58, opened 2022-12-09
pie-microsite@v1.26.0 - mvvp home page styles and mobile screen adjustment

jamieomaguire:
v1.26.0
*December 14, 2022*

### Added
- Background image to hero section of content pages and home page
- Home page image

### Changed
- Increased small screen styling media query from `narrowMid` to `mid` to fix awkward tablet styling


apps/pie-microsite/src/assets/styles/base/_layout.scss
```diff
@@ -6,14 +6,15 @@
     grid-template: 64px 1fr max-content / 260px 1fr;
     grid-template-areas:
         'header header'
-        'nav main'
-        'nav footer';
+        'main main'
```
xander-marjoram, line 9: Is this the same?
```
grid-template-areas:
    'header'
    'main'
    'footer'
```

apps/pie-microsite/src/index.md
```diff
@@ -1,5 +1,7 @@
 ---
+site_title: PIE
 layout: layouts/page-layout.njk
 title: Principles for Interfaces and Experiences
+description: PIE is the global design system for Just Eat Takeaway, powering all our products.
```
xander-marjoram, line 5: Should it be this?
```suggestion
description: PIE is the global design system for Just Eat Takeaway.com, powering all our products.
```

xander-marjoram:
The home hero SVG looks to have big transparent bars at the top and bottom, is that intentional?

apps/pie-microsite/src/index.md
jamieomaguire: I'm just going by the copy on the designs. I'll check with the designers

apps/pie-microsite/src/assets/styles/base/_layout.scss
jamieomaguire: Not quite. It currently still uses to a 2 column grid on mobile and desktop (your code is a single column grid). We can change this in future if we need to but I'd prefer to leave it for this PR if that's ok

jamieomaguire:
> The home hero SVG looks to have big transparent bars at the top and bottom, is that intentional?

Nope, good spot! Also the SVG doesn't seem to place nice with SVGOMG annoyingly. I've put the correct SVG in without the spaces.

apps/pie-microsite/src/assets/styles/base/_layout.scss
xander-marjoram: Ah right, I thought this was just being used for the layout for the top level components (header, main, footer), and we would specify the column layout inside those using `grid-template-columns`, for example.

apps/pie-microsite/src/index.md
jamieomaguire: updated! thanks for catching this

xander-marjoram (approved):

FayeCarter (approved):

PR #65, opened 2023-01-05
pie-microsite@v1.30.0 - Iconography overview page

FayeCarter:
v1.30.0
*January 6, 2023*

### Added
- Iconography page folder and overview page content
- Iconography page assets (not all SVGs have been optimised because we are likely going to update images soon with properly optimised ones)
- New `list` shortcode to allow render 'pill' and 'ordered' lists

### Changed
- Notification shortcode to render markdown
- Markdown filter to render inline markdown


apps/pie-microsite/src/_11ty/filters/markdown.js
```diff
@@ -4,4 +4,4 @@ const md = new MarkdownIt({
     html: true
 });
 
-module.exports = content => md.render(content);
+module.exports = (content, inline = false) => (inline ? md.renderInline(content) : md.render(content));
```
jamieomaguire, line 7: we should probably add jsdoc annotations here to explain what `inline` does

apps/pie-microsite/src/_11ty/shortcodes/notification.js
```diff
@@ -65,6 +66,6 @@ module.exports = function (config) {
 
     return `<aside class="${contextClass} c-notification" style="--bg-colour: ${bgColour}">
       ${svg}
-      <p class="c-notification-message">${config.message}</p>
+      <p class="c-notification-message">${markdownFilter(config.message, true)}</p>
```
jamieomaguire, line 69: can we update the jsdoc annotation for `config.message` to say it supports markdown? There's an example in the `contentPageImage.js` shortcode

apps/pie-microsite/src/_11ty/shortcodes/list.js
```diff
@@ -0,0 +1,20 @@
+const markdownFilter = require('../filters/markdown');
+
+/**
+ * A List HTML component
+ * @param {string} type - Type of list: orderd, pill
+ * @param {string} items - An array of list items
```
jamieomaguire, line 6: I think the type for `items` should be `string[]` as it's an array of strings. Also I notice we use capitalised types in some JSDoc comments and not others. Probably worth us seeing which is valid and keeping it consistent

apps/pie-microsite/src/_11ty/shortcodes/list.js
```diff
@@ -0,0 +1,20 @@
+const markdownFilter = require('../filters/markdown');
+
+/**
+ * A List HTML component
+ * @param {string} type - Type of list: orderd, pill
```
jamieomaguire, line 5: typo: ordered

apps/pie-microsite/src/_11ty/shortcodes/list.js
```diff
@@ -0,0 +1,20 @@
...
+ * @param {string} items - An array of list items
+ * @returns {string}
+ */
+// eslint-disable-next-line func-names
+module.exports = function ({ type, items }) {
+    const listItems = items.map(item => `<li class="c-list-item"><span>${markdownFilter(item, true)}</span></li>`).join('');
+    const listClasses = [
+        'c-list',
+        `c-list--${type}`
```
jamieomaguire, line 14: not critical but this allows the author to inject anything into the css class. We could always create a map of valid types or something and try to retrieve the value they provide from it (so they give the shortcode a key to look up). Then we can provide a default or fail the build if its an invalid type.

apps/pie-microsite/src/assets/styles/components/_list.scss
```diff
@@ -0,0 +1,50 @@
+@use '../fozzie-config' as f;
+
+.c-list {
+    list-style: none;
+    padding-left: 0;
+}
+
+.c-list--ordered {
```
jamieomaguire, line 8: Bit of a nit, and might be worth seeing what @ashleynolan thinks. But I wonder if it makes more sense for the selector to be:

```
.c-list.c-list--ordered {
    // styles
}
```

That way the class modifier _has_ to be used in combination with the root/block class.

Totally subjective opinion!

apps/pie-microsite/src/assets/styles/components/_list.scss
```diff
@@ -0,0 +1,50 @@
...
+    counter-reset: item;
```
jamieomaguire, line 9: what is `item` here? I can't see it on the MDN docs

apps/pie-microsite/src/_11ty/shortcodes/list.js
```diff
@@ -0,0 +1,20 @@
...
+    ];
+
+    return `<ul class="${listClasses.join(' ')}">
```
jamieomaguire, line 17: If the type is ordered, we should use an `<ol>` here :)

apps/pie-microsite/src/assets/styles/components/_list.scss
```diff
@@ -0,0 +1,50 @@
...
+
+    .c-list-item {
```
jamieomaguire, line 11: Screenshot 2023-01-05 at 15 52 39
the pill items to not appear to wrap when their container shrinks

apps/pie-microsite/src/assets/styles/components/_list.scss
```diff
@@ -0,0 +1,50 @@
...
+            align-items: center;
+            justify-content: center;
+            flex: 0 0 auto;
+        }
+    }
+}
+
+.c-list--pill {
+    display: flex;
+    gap: 8px;
```
jamieomaguire, line 41: can this be linked to the `spacing()` function?

apps/pie-microsite/src/assets/styles/components/_list.scss
```diff
@@ -0,0 +1,50 @@
...
+
+    .c-list-item {
+        @include f.font-size('body-s');
+        background-color: f.$color-container-strong;
+        padding: 2px 4px;
+        border-radius: 8px;
```
jamieomaguire, line 47: same here regarding the `spacing` thing

apps/pie-microsite/src/content/pages/foundations/iconography/overview.md
```diff
@@ -0,0 +1,184 @@
...
+{% contentPageImage {
+    src:"../../../../../assets/img/foundations/iconography/icon-anatomy.svg",
+    width: "199px"
+} %}
+
+{% list {
+    type: "ordered",
+    items: [
+        "**Icon area:** (Small (Default size): 14x14px / Large (Default size): 28x28px).",
+        "**Bounding box:** (Small (Default size): 16x16px / Large (Default size): 32x32px)."
```
jamieomaguire, line 47: just checking - do we guarantee that the inputted item order is preserved?

apps/pie-microsite/src/content/pages/foundations/iconography/overview.md
```diff
@@ -0,0 +1,184 @@
...
+## Appearance
+
+There are two types of appearance options for our icons:
+
+### Default
+
+This is the default look for our icons and should always be considered as the first option to use when placing icons in your designs.
+
+{% contentPageImage {
+    src:"../../../../../assets/img/foundations/iconography/appearance-default.svg",
```
jamieomaguire, line 100: nit: should probably have a space after the colon

jamieomaguire:
Fantastic stuff - just a few minor bits I've commented on 👍

apps/pie-microsite/src/content/pages/foundations/iconography/overview.md
jamieomaguire: same for all the onces below too haha

jamieomaguire:
small request - please could you add the inline arg to the markdown filter in `contentPageImage.js`? :)

apps/pie-microsite/src/assets/styles/components/_list.scss
FayeCarter: As far as I understand, this is where we create a counter called 'item' then on line 12 we increase the counter by 1 and apply that number to the before.
I'll add a comment in
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Counter_Styles/Using_CSS_counters

apps/pie-microsite/src/_11ty/shortcodes/list.js
FayeCarter: I will give it a go :)

apps/pie-microsite/src/content/pages/foundations/iconography/overview.md
FayeCarter: Yep, as discussed, we have a snapshot test for it

apps/pie-microsite/src/assets/styles/components/_list.scss
FayeCarter: As there's not a token for these values I've left them as `px` values for now

raoufswe (approved):

LTurns (approved):

apps/pie-microsite/src/_11ty/shortcodes/contentPageImage.js
```diff
@@ -7,7 +7,7 @@ const markdownFilter = require('../filters/markdown');
  * @returns {String} a <figcaption> element containing the caption text, or an empty string if no caption is provided.
  */
 const createCaption = config => (config.caption
-    ? `<figcaption class="c-contentImage-caption">${markdownFilter(config.caption)}</figcaption>`
+    ? `<figcaption class="c-contentImage-caption"><span>${markdownFilter(config.caption, true)}</span></figcaption>`
```
ashleynolan, line 10: More out of interest than anything – what's the span for here? (just as there's no styling in this PR to go along with adding it)

apps/pie-microsite/src/_11ty/shortcodes/list.js
```diff
@@ -0,0 +1,26 @@
+const markdownFilter = require('../filters/markdown');
+const listTypes = require('../../_data/listTypes');
+
+/**
+ * A List HTML component
```
ashleynolan, line 5:
```suggestion
 * A List HTML component – takes an array of list items and turns them into a marked-up list
```

apps/pie-microsite/src/__tests__/_11ty/shortcodes/list.test.js
```diff
@@ -0,0 +1,16 @@
...
+    it.each([
+        'ordered',
+        'pill'
+    ])('should return the expected HTML', type => {
+        // act
+        const result = list({ type, items });
+
+        // assert
+        expect(result).toMatchSnapshot();
+    });
```
ashleynolan, line 15: Would be good to include a test(s) that checks the different types you can pass in

apps/pie-microsite/src/assets/styles/components/_list.scss
```diff
@@ -0,0 +1,53 @@
+@use '../fozzie-config' as f;
+
+.c-list {
+    list-style: none;
+    padding-left: 0;
+}
+
+.c-list.c-list--ordered {
```
ashleynolan, line 8: I tend not to add the base to the modifier here, mainly as it makes the CSS more specific (just avoids hitting any specificity issues down the line). As it's a modifier, the name should imply that it shouldn't be used without the base class

```suggestion
.c-list--ordered {
```

apps/pie-microsite/src/assets/styles/components/_list.scss
```diff
@@ -0,0 +1,53 @@
...
+            color: f.$color-white;
+            display: flex;
+            align-items: center;
+            justify-content: center;
+            flex: 0 0 auto;
+        }
+    }
+}
+
+.c-list.c-list--pill {
```
ashleynolan, line 41: Same here – I'd just make this:

```suggestion
.c-list--pill {
```

ashleynolan (commented):
Couple of small comments – otherwise LGTM

apps/pie-microsite/src/_11ty/shortcodes/contentPageImage.js
FayeCarter: The markdown filter returns it without and enclosing tag which meant that the styling would break if it had another html tag rendered as part of the html. I've removed it now as I've round that I can get the markdown filter to return the span so we don't need to apply it to every shortcode.

apps/pie-microsite/src/_11ty/shortcodes/list.js
```diff
@@ -0,0 +1,26 @@
+const markdownFilter = require('../filters/markdown');
+const listTypes = require('../../_data/listTypes');
```
jamieomaguire, line 2: I may be wrong _but_ I think you can access any global 11ty data through `this` within the shortcode function (but not nested functions called inside it). If that works, you can remove the import and it would be something like `this.data.listTypes` (actual syntax may vary - I think the docs mention this)

jamieomaguire (approved):

LTurns (approved):

PR #66, opened 2023-01-09
pie-microsite@v1.33.0 - global tokens added to spacing page

LTurns:
### Added
- global design tokens to spacing page


apps/pie-microsite/src/_11ty/shortcodes/tokensTable/index.js
```diff
@@ -77,6 +77,17 @@ const buildColorExample = token => {
     return `<div class="${classes.join(' ')}" style="${cssVariable}";></div>`;
 };
 
+/**
+ * Builds the example spacing swatch to show on the token list item
+ * @param {string} token - the token value i.e. 
+ * @returns {string} - the color swatch example HTML string
+ */
+const buildSpacingExample = token => {
+    const cssVariable = `${token}px`;
```
ashleynolan, line 86: I'd make the name a bit more descriptive here – is this passing over the spacing token value?

If so, `spacingTokenValue` might be a better name? (or whatever you think describes it best!)

apps/pie-microsite/src/_11ty/shortcodes/tokensTable/index.js
```diff
@@ -77,6 +77,17 @@ const buildColorExample = token => {
...
+
+    return `<div class="c-tokensTable-example--spacing" style="--example-spacing: ${cssVariable}";></div>`;
```
ashleynolan, line 88: `--example-spacing` could be better named too. (happy to have a quick hangout and chat about, as probably easier if could see the context of it?)

ashleynolan: I can see it's used above too – I reckon move some of this out to the other variable above so it makes more sense in context. So:

const cssVariable = `--example-spacing: ${token}px`;

apps/pie-microsite/src/_11ty/shortcodes/tokensTable/index.js
```diff
@@ -178,6 +190,27 @@ const buildTokensListForCategory = (tokens, path, category, tokenType) => {
     return buildTokensList(tokenListElements);
 };
 
+/**
+ * Builds a categorised list of tokens
+ * @param {string} path - path to the category i.e.  'path:color.alias.default' / 'path:color.alias.dark'
```
FayeCarter, line 195: is this still needed?

apps/pie-microsite/src/assets/styles/components/_tokensTable.scss
```diff
@@ -51,6 +51,19 @@ $tokensTable-border-width: 1px;
     }
 }
 
+.c-tokensTable-example--spacing {
+    width: var(--example-spacing);
+    height: var(--example-spacing);
+    background-color: #d7d9da;
```
FayeCarter, line 57: I think this is `f.$color-content-default` on figma

apps/pie-microsite/src/utilities/helpers.js
```diff
@@ -5,14 +5,18 @@
...
-        result = result[pathSegments[i]];
-    }
+        for (let i = 0; i < pathSegments.length; i++) {
+            result = result[pathSegments[i]];
+        }
 
-    return result;
+        return result;
+    } catch {
+        return undefined;
```
FayeCarter, line 18: I think it might be worth throwing an error here?

apps/pie-microsite/src/assets/styles/components/_tokensTable.scss
```diff
@@ -51,6 +51,19 @@ $tokensTable-border-width: 1px;
...
+    display: block;
+    margin: 10%;
+    margin-left: auto;
+    margin-right: auto;
```
FayeCarter, line 61: Could probably do `margin: 10% auto;` not sure it should be a `%` or a px value though

FayeCarter: ooh actually you might be able to do 
```
  align-self: center;
  justify-self: center;
```

apps/pie-microsite/src/utilities/helpers.js
jamieomaguire: Agreed - in general we want the entire site to fail the build if we get errors like this @LTurns , otherwise we risk silent failings getting built. Visual regression _may_ catch them but we should make it as difficult as possible to allow them go through.

Like Faye suggested - throwing a descriptive error here would probably be preferable.

apps/pie-microsite/src/assets/styles/components/_tokensTable.scss
jamieomaguire: I'm not quite sure what this is for but I'm unsure whether we have any reason to use a percentage based margin - what was the need for this?

apps/pie-microsite/src/utilities/helpers.js
jamieomaguire: @FayeCarter I can see your point, but in this case we actually do want to return undefined so it matches the API you'd expect when querying an object for an unknown property. This means the consume would throw an error when they hit undefined, rather than the helper function which might be used to test for the property's existence.

jamieomaguire:
Screenshot 2023-01-10 at 12 16 58
Small thing I noticed - we need a background colour on mobile for the examples, and the visual bit needs to be vertically and horizontally centered :)

jamieomaguire:
Screenshot 2023-01-10 at 12 17 46
How it should look (related to my comment above)

apps/pie-microsite/src/utilities/helpers.js
FayeCarter: I wish the homer hide emoji was available on github

apps/pie-microsite/src/assets/styles/components/_tokensTable.scss
LTurns: ah you can ignore this 😄 updated to just use flexbox after I checked figma 👍

jamieomaguire:
please also add a PR description and invite design to review the preview deployment

apps/pie-microsite/src/assets/styles/components/_tokensTable.scss
```diff
@@ -51,6 +51,20 @@ $tokensTable-border-width: 1px;
...
+    align-self: center;
+    justify-self: center;
+    background-color: f.$color-content-default;
+    opacity: 10%;
+    width: var(--example-spacing);
+    height: var(--example-spacing);
+    border-radius: 2px;
+
+    @include f.media('>=wide') {
+        justify-self: left;
```
FayeCarter, line 64: this one should be 
```
  justify-self: start;
  align-self: start;
```

apps/pie-microsite/src/assets/styles/components/_tokensTable.scss
```diff
@@ -104,13 +120,15 @@ $tokensTable-border-width: 1px;
...
     padding: $tokensTable-column-padding;
+    background-color: f.$color-container-default;
 
     > * {
         display: block;
     }
 
     @include f.media('>=wide') {
         padding: 0;
+        background-color: f.$color-mozzarella-5;
```
FayeCarter, line 131: same for this guy `f.$color-background-default`

FayeCarter (dismissed):

apps/pie-microsite/src/assets/styles/components/_tokensTable.scss
```diff
@@ -86,11 +101,14 @@ $tokensTable-border-width: 1px;
...
     border: $tokensTable-border-width solid f.$color-border-default;
     border-radius: f.$radius-03;
+    background-color: f.$color-background-subtle;
 
     @include f.media('>=wide') {
         padding: f.spacing(e) 0;
         border-radius: 0;
         border-width: 0 0 $tokensTable-border-width 0;
+        background-color: f.$color-background-default;
+        min-height: 128px;
```
jamieomaguire, line 111: I've definitely seen this number elsewhere, I think for the row heights - can it become a variable please? :)

raoufswe:
image

jamieomaguire (approved):

apps/pie-microsite/src/_11ty/shortcodes/tokensTable/index.js
```diff
@@ -135,15 +135,19 @@ const buildTokenListElements = ({
...
-
+    
     // TODO - description is just an example of how we might use the metadata
     // We would likely wanted to move them into a colour specific handler similar to how we build
     // the colour token example. Please consider them placeholder for now.
     const tokenDescription = tokenMetadata.description
         ? `<span class="c-tokensTable-tokenDescription">${tokenMetadata.description}</span>`
         : '';
+    
+    const cssVariable = tokenType === 'color'
```
FayeCarter, line 146: `tokenTypes.COLOR`

apps/pie-microsite/src/utilities/helpers.js
```diff
@@ -27,12 +31,18 @@ const capitaliseFirstLetter = string => {
     return capitalised;
 };
 
+// checks whether value can convert to number
```
jamieomaguire, line 34: Can we use JS docs here rather than double slash? See the other functions in this file for examples

apps/pie-microsite/src/_11ty/shortcodes/tokensTable/index.js
```diff
@@ -119,19 +131,23 @@ const buildTokenListElements = ({
...
+        ? '--token-row-height: 128px'
+        : '--token-row-height: 176px';
 
-    return `<li class="c-tokensTable-row c-tokensTable-item">
+    return `<li class="c-tokensTable-row c-tokensTable-item" style="${cssVariable}">
```
jamieomaguire, line 150: ah I see, I was wondering if in the SCSS we could default to 176px and if its colour then use 128. It would save us having to inject this to every node on the page

FayeCarter (approved):

FayeCarter (dismissed):

raoufswe (approved):

jamieomaguire (dismissed):

apps/pie-microsite/src/_11ty/shortcodes/tokensTable/index.js
```diff
@@ -77,6 +77,17 @@ const buildColorExample = token => {
     return `<div class="${classes.join(' ')}" style="${cssVariable}";></div>`;
 };
 
+/**
+ * Builds the example spacing swatch to show on the token list item
+ * @param {string} token - the token value i.e.
```
ashleynolan, line 82: Were you meant to put something after the `i.e.`?

apps/pie-microsite/src/_11ty/shortcodes/tokensTable/index.js
```diff
@@ -178,6 +190,26 @@ const buildTokensListForCategory = (tokens, path, category, tokenType) => {
     return buildTokensList(tokenListElements);
 };
 
+/**
+ * Builds a categorised list of tokens
```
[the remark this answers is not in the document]
ashleynolan: Doesn't look like this has updated?

apps/pie-microsite/src/_11ty/shortcodes/tokensTable/index.js
```diff
@@ -178,6 +190,30 @@ const buildTokensListForCategory = (tokens, path, category, tokenType) => {
...
```
ashleynolan, line 194: This comment seems at odds with the function name?

The function is called `buildUncategorisedLists`, but the comment says it builds a ategorised list?

apps/pie-microsite/src/_11ty/shortcodes/tokensTable/index.js
```diff
@@ -178,6 +190,30 @@ const buildTokensListForCategory = (tokens, path, category, tokenType) => {
...
+ * @param {string} tokenType - the type of token i.e. color, spacing, radius
+ * @param {object} tokens
+ * @returns - object of token categories with display names i.e.  whiteBlack: { displayName: 'White and Black' }
+ */
+const buildUncategorisedLists = ({
+    tokenType, tokens 
+}) => {
+    // if tokens are numbers (spacing / radius), sort is ascending order
```
ashleynolan, line 202:
```suggestion
    // if tokens are numbers (spacing / radius), sort in ascending order
```

apps/pie-microsite/src/_11ty/shortcodes/tokensTable/index.js
```diff
@@ -178,6 +190,30 @@ const buildTokensListForCategory = (tokens, path, category, tokenType) => {
     return buildTokensList(tokenListElements);
 };
 
+/**
+ * Builds a categorised list of tokens
+ * @param {string} tokenType - the type of token i.e. color, spacing, radius
+ * @param {object} tokens
+ * @returns - object of token categories with display names i.e.  whiteBlack: { displayName: 'White and Black' }
```
ashleynolan, line 197: Just to double-check – does this function return an object? It looks like it returns `buildTokensList()` which outputs a string (but not sure if I'm missing something)

apps/pie-microsite/src/__tests__/utilities/helpers.test.js
```diff
@@ -49,4 +49,18 @@ describe('helpers.js', () => {
             });
         });
     });
+
+    describe('numberHelpers', () => {
+        describe('isNumber', () => {
+            [['1', true], ['fooBarBaz', false]].forEach(example => {
+                it('correctly checks whether string can convert to number', () => {
```
ashleynolan, line 56: Out of interest – will this function pass true if you pass an actual number to it?

Main reason I ask is that the name of the function implies that it'd return true in that scenario. If it does pass, then the function name is all good, but if it doesn't for some reason, I'd rename the function to be more string specific.

I'd add a test here for passing in a number either way (so just `10` or something), as it'd be good to have that covered.

apps/pie-microsite/src/_11ty/shortcodes/tokensTable/index.js
LTurns: ah yep, sorry copy and paste issue! will fix 👍

apps/pie-microsite/src/assets/styles/components/_tokensTable.scss
```diff
@@ -91,6 +132,7 @@ $tokensTable-border-width: 1px;
         padding: f.spacing(e) 0;
         border-radius: 0;
         border-width: 0 0 $tokensTable-border-width 0;
+        background-color: f.$color-background-default;
```
ashleynolan, line 135: Don't think you need this line (looks like removing it in the preview has no effect)

ashleynolan (commented):
Few comments from me – nothing major though 😄

apps/pie-microsite/src/__tests__/utilities/helpers.test.js
LTurns: just checked and it works with a number too - I've added the test and will leave the function name as is 😄

apps/pie-microsite/src/utilities/helpers.js
```diff
@@ -27,12 +31,22 @@ const capitaliseFirstLetter = string => {
     return capitalised;
 };
 
+/**
+ * Checks whether string can convert to number
+ * @param {string} string
```
jamieomaguire, line 36: think this needs to accept a string OR number, based on your unit tests :)

jamieomaguire: Or I would rename to something like isStringANumber or something and remove the unit test you added. It's a little confusing otherwise.

LTurns: ah yep, will change the param to string or number, thank you! 😄

apps/pie-microsite/src/utilities/helpers.js
```diff
@@ -32,8 +32,9 @@ const capitaliseFirstLetter = string => {
 };
 
 /**
- * Checks whether string can convert to number
+ * Checks whether string can convert to number - returns number if number is passed through
  * @param {string} string
+ * @param {number} number
```
jamieomaguire, line 37: sorry - needs to be like this:
`* @param {string|number} value The value to test can be parsed as a number.`

(or something like that)

apps/pie-microsite/src/utilities/helpers.js
```diff
@@ -32,8 +32,9 @@ const capitaliseFirstLetter = string => {
 };
 
 /**
- * Checks whether string can convert to number
+ * Checks whether string can convert to number - returns number if number is passed through
```
jamieomaguire, line 35: I don't think this is correct - does this function return a number?

jamieomaguire (approved):

raoufswe (approved):

apps/pie-microsite/src/_11ty/shortcodes/tokensTable/index.js
```diff
@@ -77,6 +77,17 @@ const buildColorExample = token => {
...
 
+/**
+ * Builds the example spacing swatch to show on the token list item
+ * @param {string} token - the token value i.e. 24, 80
+ * @returns {string} - the spacing swatch example HTML string
+ */
+const buildSpacingExample = token => {
+    const cssVariable = `--example-spacing: ${token}px`;
+
+    return `<div class="c-tokensTable-example-spacing-container"><div class="c-tokensTable-example--spacing" style="${cssVariable}";></div></div>`;
```
xander-marjoram, line 88: This semicolon looks a bit out of place to me
```suggestion
    return `<div class="c-tokensTable-example-spacing-container"><div class="c-tokensTable-example--spacing" style="${cssVariable}"></div></div>`;
```

apps/pie-microsite/src/__tests__/utilities/helpers.test.js
```diff
@@ -49,4 +49,18 @@ describe('helpers.js', () => {
             });
         });
     });
+
+    describe('numberHelpers', () => {
+        describe('isNumber', () => {
+            [['1', true], ['fooBarBaz', false], [1, true]].forEach(example => {
```
xander-marjoram, line 55: Personally I find the cases a bit easier to read when each one is put on its own line

PR #72, opened 2023-01-11
pie-microsite@v1.34.0 Radius page - Add global tokens

fernandofranca:
### Added
-   Radius page for global design tokens 

### Changed
-   Tokens metadata for radius global design tokens 

#### Contains bits of code picked from spacing PR
Already rebased. ~~The code from the [first commit](https://github.com/justeattakeaway/pie/pull/72/commits/818fea7ad13f2793de158afa114dcc2a557c4444) should be ignored because as soon the spacing PR gets merged I'll rebase and remove these bits from this PR.~~

#### Extra padding on the top is missing

Solved. ~~How do we create an exception to change the height or padding of the example container? I'm asking because its value is set by `grid-template-rows`.~~


jamieomaguire (dismissed):

apps/pie-microsite/src/assets/styles/components/_tokensTable.scss
```diff
@@ -92,6 +92,32 @@ $tokensTable-item-min-height: 128px;
...
+    @include f.media('>=wide') {
+        display: block;
+        background-color: transparent;
+    }
+}
+
+.c-tokensTable-example-radius {
+    width: 80px;
+    height: 80px;
+    margin-top: 16px;
```
jamieomaguire, line 111: we can use `f.spacing(d)` here instead of `16px`

[reference](https://github.com/justeat/fozzie/blob/eb5775f47d30e21b4b60407da9fa5dd36a3ecbc9/src/scss/settings/_variables.scss#L98)

jamieomaguire (approved):

fernandofranca:
Sorry, had to add a small change, but it's for good.

jamieomaguire (dismissed):

apps/pie-microsite/src/assets/styles/components/_tokensTable.scss
```diff
@@ -92,6 +92,32 @@ $tokensTable-item-min-height: 128px;
     }
 }
 
+.c-tokensTable-example-radius-container {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    background-color: f.$color-background-subtle;
+    min-height: $tokensTable-spacing-container-height;
```
raoufswe, line 100: nitpick: I would recommend renaming the variable to `$tokensTable-container-height` if it's gonna be used by multiple tokens.

raoufswe (approved):

apps/pie-microsite/src/assets/styles/components/_tokensTable.scss
```diff
@@ -92,6 +92,32 @@ $tokensTable-item-min-height: 128px;
...
+
+    @include f.media('>=wide') {
+        display: block;
+        background-color: transparent;
```
raoufswe, line 104: I think we might need to change the min-height for wide screens as it doesn't really match the spacing we have in designs (80px). 

Screenshot 2023-01-13 at 09 44 44

apps/pie-microsite/src/assets/styles/components/_tokensTable.scss
fernandofranca: I think it's a good idea, however, I wasn't sure if all containers should or will have the same height.
Curious to know other opinions.

jamieomaguire: Sounds like a good idea to me too, for what it's worth 👍

fernandofranca: In Figma designs elevation uses 160, spacing and radius 176, and typography 184.

raoufswe: then I think we probably need to change the `min-height` value to match the elevation token height. 

```suggestion
    min-height: '160px';
    
    wide {
        min-height: '80px';
    }
```

raoufswe (approved):

raoufswe:
image

jamieomaguire (approved):

raoufswe (approved):

jamieomaguire (approved):

PR #83, opened 2023-01-17
pie-microsite@v1.37.0 radius page - alias tokens added

LTurns:
### Added
-   Alias tokens to radius page


apps/pie-microsite/src/_11ty/shortcodes/tokensTable/index.js
```diff
@@ -19,17 +19,22 @@ const {
  * @param {string} tokenType - the type of token i.e. color, spacing, radius
  * @returns {string} the SCSS variable name
  */
-const createScssTokenName = (tokenKey, tokenType) => `$${tokenType}-${tokenKey}`;
+const createScssTokenName = (tokenKey, tokenType, path = '') => {
+    const prefixExcludes = [path === 'radius.alias' ? tokenTypes.RADIUS : []];
```
jamieomaguire, line 23: I think what would be cleaner and more extensible here would be to branch the logic based on token type and have separate handler for picky tokens like radius, and a default for others. Similar to what we do the for create example element function :)

apps/pie-microsite/src/_11ty/shortcodes/tokensTable/index.js
```diff
@@ -19,17 +19,22 @@ const {
...
  * Creates a display name of the provided token. 'system-purple' would become 'System Purple'
  * @param {string} tokenKey - the token key i.e. 'support-positive-02'
  * @param {string} tokenType - the type of token i.e. color, spacing, radius
  * @returns {string} the display name of the token
  */
-const createTokenDisplayName = (tokenKey, tokenType) => {
+const createTokenDisplayName = (tokenKey, tokenType, path = '') => {
     // Some tokens don't require a prefix in front of their display names
-    const prefixExcludes = [tokenTypes.COLOR];
+    const prefixExcludes = [tokenTypes.COLOR, path === 'radius.alias' ? tokenTypes.RADIUS : []];
```
jamieomaguire, line 37: same here regarding splitting logic based on token type

apps/pie-microsite/src/assets/styles/components/_tokensTable.scss
```diff
@@ -97,7 +97,7 @@ $tokensTable-item-min-height: 128px;
     justify-content: center;
     align-items: center;
     background-color: f.$color-background-subtle;
-    min-height: 160px;
```
jamieomaguire, line 100: are alias and global the same here?

apps/pie-microsite/src/content/pages/foundations/radius/tokens/alias.md
```diff
@@ -4,3 +4,13 @@ eleventyNavigation:
     parent: Radius
     order: 3
 ---
+
+## Alias radius tokens
+To use radiuses consistently make sure you use the alias tokens listed below:
+
+---
+
+{% tokensTable {
```
jamieomaguire, line 13: 🔥

apps/pie-microsite/src/assets/styles/components/_tokensTable.scss
LTurns: yep 😄

apps/pie-microsite/src/_11ty/shortcodes/tokensTable/index.js
raoufswe: 📓 fyi `tokenMetadata` accepts `displayName` as a property, so if you think normalizing the token adds extra overhead, we can  just pass the value directly like [here](https://github.com/justeattakeaway/pie/blob/main/apps/pie-microsite/src/tokensMetadata.json#L643)

raoufswe (dismissed):

jamieomaguire (approved):

raoufswe (approved):

PR #103, opened 2023-01-23
pie-storybook@v0.1.0 - Storybook initial setup

fernandofranca:
### Added
- `pie-storybook` workspace
- Storybook setup with support to Web Components and TypeScript
- Sample `pie-button` story


apps/pie-storybook/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "dependencies": {
+    "@justeattakeaway/pie-button": "workspace:*"
+  },
+  "devDependencies": {
+    "@storybook/addon-essentials": "7.0.0-beta.31",
+    "@storybook/addon-links": "7.0.0-beta.31",
+    "@storybook/blocks": "7.0.0-beta.31",
+    "@storybook/web-components": "7.0.0-beta.31",
+    "@storybook/web-components-vite": "7.0.0-beta.31",
+    "react": "18.2.0",
```
fernandofranca, line 21: > `Starting in 7.0, react and react-dom are now required peer dependencies of Storybook.

https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#react-peer-dependencies-required`

apps/pie-storybook/stories/pie-button.stories.ts
```diff
@@ -0,0 +1,32 @@
+import { html } from 'lit';
+import type { Meta, StoryObj } from '@storybook/web-components';
+import { PieButton } from '@justeattakeaway/pie-button';
+
+const componentRef = [PieButton];
+
+const meta: Meta = {
+    title: 'pie-button',
+    component: 'pie-button',
+};
```
fernandofranca, line 10: ### Added component reference in story

Storybook 7 uses CSF 3 for writing stories, and now the field `component` uses a string instead of the component reference. 

Since the component is not referenced anywhere, it doesn't get initialized, therefore it doesn't render.

To circumvent the issue I added a constant to hold it's reference.

`const componentRef = [PieButton];`

Unfortunately I didn't found a canonical way to solve this, only found another user with a [similar issue](https://discord.com/channels/486522875931656193/490822506916347904/1052209727146307614).

fernandofranca:
### Issue with Storybook HMR and component updates

Problem:
- When a story file is updated, HMR works as expected ✅
- When the component source code is updated, the changes are not automatically reflected on Storybook 😲
	- However when we recompile the component, the expected change is displayed
- This is undesirable because it breaks the developer workflow 

The issue happens because:
- The component `main` field in `package.json` points to the `dist` folder, containing the code compiled from TS src
- As much as we would like, we can't expect the component to compile the dist by itself
- The most common setup for Storybook assumes it lives in the same package or workspace of the components
- Compilation of a monorepo components is not a concern of Storybook

Possible approaches towards a solution:

#### 1) Keep Storybook stories inside the component workspace
- Each story file could live near the its component counterpart and simply use relative imports to the source TS files, which are always up-to-date

That's the probably the simplest solution, however it doesn't scale well:
- as soon as we introduce a different component as dependency of the story or even of the component itself, it can became outdated
- one can argue if such case will exist (component or story depending on another component), however I can point several concrete cases in Snacks

📘 Reminder: If we add story files near its component, let's ensure the TS compiler won't be compiling stories and adding them to each package `dist` folder. The consumer doesn't need story files. 😉

#### 2) Use relative imports 
If we use relative imports, pointing to component's source code (instead of `dist`), Storybook will be always up-to-date.
That might imply in using relative imports everywhere, not only stories, but also in components. Molecule components can be composed of one or more atom components. The concern is, if we do that, how do we ensure not to bundle atoms source code into molecules? 
Therefore I don't think this is a proper solution.

#### 3) Point the component `main` to `src` instead of `dist`
It's a temporary hack, not a solid solution.

#### 4) Provide a watcher mechanism
This mechanism would run in parallel with Storybook, keeping each component `dist` folder up-to-date: as soon  some change is introduced in a component, it gets recompiled.

To be fair I didn't tested this solution as it requires effort outside the scope of this ticket. My main concern is how fast TS compilation will happen to avoid a lags in dev workflow.

#### I might be missing other solutions
Suggestions are welcomed. 😀

fernandofranca:
A few reference links, in case you're interested:

- https://github.com/vercel/turbo/tree/main/examples/design-system
- https://storybook.js.org/docs/web-components/configure/overview
- https://piyushsinha.tech/introduction-to-storybook-for-web-components

turbo.json
```diff
@@ -5,6 +5,10 @@
             "cache": true,
             "dependsOn": ["^build"]
         },
+        "storybook-build": {
```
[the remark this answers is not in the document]
siggerzz: Looks good to me! Just make sure a `storybook-dev` property is added to turbo, but ensure `cache:false` is set 👍🏼 

https://turbo.build/repo/docs/core-concepts/caching#turn-off-caching

siggerzz:
Really good work, @fernandofranca. Looks great!

I'm not sure if I follow though on why story files can't live within the directory of it's associated component?

We follow this approach in fozzie components:
f-button stories - https://github.com/justeattakeaway/fozzie-components/tree/master/packages/components/atoms/f-button/stories

f-button `main` package.json - https://github.com/justeattakeaway/fozzie-components/blob/master/packages/components/atoms/f-button/package.json#L5

AFAIK HMR worked as expected.


This approach would also allow us to hook into Turborepo's `--filter` functionality, to only load story files that've been changed. An example in Fozzie Components can be seen here - https://github.com/justeattakeaway/fozzie-components/blob/master/packages/storybook/config/storybook/story.config.js#L5-L27


Happy to jump on a call tomorrow if I'm failing to grasp the HMR issue!

fernandofranca:
> I'm not sure if I follow though on why story files can't live within the directory of it's associated component?

We can. However, even if we did, it's not a solution for HMR.

- A component might depend on another local component package (because package resolution will point to `/dist`, hence might be outdated)
- A component Story might depend on another local component package (same issue with package resolution)

A few instances of this relationship (Component->Component, Story->Multiple components) happen on Snacks:

- Card component depends on [IconButton](https://git.takeaway.com/design-system/snacks-design-system/-/blob/main/packages/snacks-design-system/src/components/card/card.js#L12)
- Dropdown story depends on [Dropdown, List and ListItem](https://git.takeaway.com/design-system/snacks-design-system/-/blob/main/packages/snacks-design-system/src/components/dropdown/stories/index.js#L8)

I don't have a preference for where stories should be stored.

jamieomaguire:
@fernandofranca thanks for the write up. I wonder if "**4) Provide a watcher mechanism**" is the solution. Haven't really worked with TS based components before so unsure. Happy to be a part of that call with @siggerzz though :)

fernandofranca:
I believe the issue with Storybook HMR originates from the way we are working with TypeScript.

If we reframe the situation, replacing Storybook with a Node.js app, the issue will persist. An app shouldn't be responsible for recompiling the source code of the dependencies it consumes by itself.

I did a bit of research and found a `--watch` flag for the TypeScript compiler: https://www.typescriptlang.org/docs/handbook/configuring-watch.html

It would be more appropriate to deal with the TypeScript issue in a follow-up ticket as it doesn't belong to this one.

wdyt? @ashleynolan @jamieomaguire @siggerzz

kevinrodrigues (approved):

fernandofranca:
- Added `storybook-serve` script and Turborepo config
- Updated README info
- Updated Storybook version

kevinrodrigues (approved):

kevinrodrigues:
> ### Issue with Storybook HMR and component updates
> Problem:
> 
> * When a story file is updated, HMR works as expected ✅
> * When the component source code is updated, the changes are not automatically reflected on Storybook 😲
>   
>   * However when we recompile the component, the expected change is displayed
> * This is undesirable because it breaks the developer workflow
> 
> The issue happens because:
> 
> * The component `main` field in `package.json` points to the `dist` folder, containing the code compiled from TS src
> * As much as we would like, we can't expect the component to compile the dist by itself
> * The most common setup for Storybook assumes it lives in the same package or workspace of the components
> * Compilation of a monorepo components is not a concern of Storybook
> 
> Possible approaches towards a solution:
> 
> #### 1) Keep Storybook stories inside the component workspace
> * Each story file could live near the its component counterpart and simply use relative imports to the source TS files, which are always up-to-date
> 
> That's the probably the simplest solution, however it doesn't scale well:
> 
> * as soon as we introduce a different component as dependency of the story or even of the component itself, it can became outdated
> * one can argue if such case will exist (component or story depending on another component), however I can point several concrete cases in Snacks
> 
> 📘 Reminder: If we add story files near its component, let's ensure the TS compiler won't be compiling stories and adding them to each package `dist` folder. The consumer doesn't need story files. 😉
> 
> #### 2) Use relative imports
> If we use relative imports, pointing to component's source code (instead of `dist`), Storybook will be always up-to-date. That might imply in using relative imports everywhere, not only stories, but also in components. Molecule components can be composed of one or more atom components. The concern is, if we do that, how do we ensure not to bundle atoms source code into molecules? Therefore I don't think this is a proper solution.
> 
> #### 3) Point the component `main` to `src` instead of `dist`
> It's a temporary hack, not a solid solution.
> 
> #### 4) Provide a watcher mechanism
> This mechanism would run in parallel with Storybook, keeping each component `dist` folder up-to-date: as soon some change is introduced in a component, it gets recompiled.
> 
> To be fair I didn't tested this solution as it requires effort outside the scope of this ticket. My main concern is how fast TS compilation will happen to avoid a lags in dev workflow.
> 
> #### I might be missing other solutions
> Suggestions are welcomed. 😀

From a brief look without looking at it in too much detail the watcher task seems like the easiest solution to me, maybe a spike ticket to test the TS compilation vs Dev experience might be worth looking into?

apps/pie-storybook/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "pie-storybook",
+  "version": "0.1.0",
```
siggerzz, line 3: Forgot to mention. I think the `version` can be removed. I don't think we'll be wanting to publish the storybook instance as it won't be used by consumers 👍🏼

fernandofranca: We don't need to publish Storybook as a package, however not having a version field seems to be a [bad practice](https://docs.npmjs.com/creating-a-package-json-file#:~:text=A%20package.json-,file%20must%20contain,-%22name%22%20and)

> A package.json file must contain "name" and "version" fields.

siggerzz: Cool, sounds good! Yeah that's fine with me. This will still continue to play along nicely with changesets 👍🏼

siggerzz (approved):

PR #107, opened 2023-01-24
pie-icons@v2.0.0-beta.3 – Updating to use new icons

ashleynolan:
### Added
- Support for subdirectories of assets, so that these subdirectories are also respected when they are moved to the `_optimised` and `dist` folders

### Changed
- **Breaking** – Icon names have been normalised against design and brand naming. This means a lot of icons (mainly naming) have been changed in this release (please contact the PIE team for help with migration).
- `assets/optimised` renamed to `assets/_optimised` so that it always appears at the top of the folder tree
- Convert `build.sh` to `build.js` to enable Windows users to build the project locally
- Updated NPM Scripts to work with Turborepo when run from root of repo

### Removed
- Old references to `icing-` prefixed icons (which no longer exist).


packages/tools/pie-icons/bin/build.js
```diff
@@ -1,14 +1,14 @@
...
 
-console.log('Making src/assets/optimised directory');
-execSync('mkdir "src/assets/optimised"');
+console.log('Making src/assets/_optimised directory');
+execSync('mkdir "src/assets/_optimised"');
 
 // Process SVG files (Optimise)
 console.log('Optimising SVGS');
-execSync('npx babel-node bin/process-svgs.js');
+execSync('npx babel-node bin/process-svgs.js', { stdio: 'inherit' });
```
jamieomaguire, line 11: out of interest what does this arg do?

jamieomaguire (approved):

packages/tools/pie-icons/bin/build.js
ashleynolan: It passes any log output up to the parent process, so you can see it in the shell output (otherwise it gets suppressed)

raoufswe (approved):

packages/tools/pie-icons/package.json
```diff
@@ -18,9 +18,9 @@
     "prepare": "yarn build && yarn test:coverage",
     "clean": "rimraf -rf ./dist/*",
     "prepublishOnly": "yarn run build",
-    "lint:scripts": "eslint .",
-    "lint:scripts:fix": "eslint . --fix",
-    "test": "jest",
+    "lint:scripts": "run -T eslint .",
```
fernandofranca, line 21: What does `run -T` do?

fernandofranca (approved):

packages/tools/pie-icons/package.json
xander-marjoram: @fernandofranca It's a little-known flag that we only recently discovered! https://yarnpkg.com/cli/run#options

In this case it looks for the eslint dependency in the root of the monorepo.

fernandofranca: That's awesome! 🔥

PR #110, opened 2023-01-24
v0.3.0 - Adds button styling.

kevinrodrigues:
### Added
- Button default styling (medium).
- JET font family to html file.

### Screenshot
Screenshot 2023-01-24 at 15 44 29


packages/components/pie-button/src/index.ts
```diff
@@ -1,10 +1,27 @@
...
+      button {
+        background-color: var(--dt-color-interactive-brand);
+        border-radius: var(--dt-radius-rounded-e);
+        border: none;
+        color: var(--dt-color-content-interactive-primary);
+        font-size: 20px; // A future ticket will pull in a helper for font size & line height.
+        line-height: 1.4;
+        font-family: var(--dt-font-interactive-m-family);
+        font-weight: var(--dt-font-interactive-m-weight);
+        padding: auto;
```
raoufswe, line 18: i think `padding: auto;` is not a valid CSS value

Screenshot 2023-01-24 at 16 47 16

xander-marjoram (approved):

packages/components/pie-button/src/index.ts
xander-marjoram: Hmm so I guess if we don't set vertical padding at all then it should still work?

raoufswe:
good stuff 🥇 

just a couple of questions/concerns

- we need to support multiple sizes/variants of the button, and the current approach doesn't provide this type of flexibility as far as I can tell. is this out of scope or I'm missing something?  

- the default button size in snacks is small, but I can't tell if the medium is the new default size on the Figma design. have we confirmed this with designers?

packages/components/pie-button/index.html
```diff
@@ -3,11 +3,58 @@
...
+                        fonts.forEach(function (font) {
+                            document.fonts.add(font);
+                        });
+                    })
+                    .then(function () {
+                        document.documentElement.classList.add   ('webfonts-loaded');
+                    });
+            }
+
+            // This next block is for IE11 and old Edge support, which don't    support the new Font loading API
```
fernandofranca, line 47: Given the context (component development) do we need this fallback code for IE11?

xander-marjoram:
@raoufswe Good questions.

One of the next tickets will be for supporting all of the sizes, we'll probably extract the functionality into a helper module to keep the component simple.

In fozzie, the default size is medium. If design agree that small should be the default then we could update fozzie to align but that would be a breaking change, so it might be worth keeping the difference for now but we can make sure this web component has the correct default value.

packages/components/pie-button/index.html
xander-marjoram: Good spot, yeah we don't need this

packages/components/pie-button/src/index.ts
```diff
@@ -1,10 +1,27 @@
...
+import { LitElement, html, css } from 'lit';
 import { customElement, property } from 'lit/decorators.js';
 
 @customElement('pie-button')
 export class PieButton extends LitElement {
     @property() actionType = 'submit';
 
+    static styles = [css`
+      button {
+        background-color: var(--dt-color-interactive-brand);
```
fernandofranca, line 10: As complexity can vary for different components, the amount of CSS definitions can grow as much. 

Since we're allowed to compose the component styles in the `styles` array, I wonder if it wouldn't be more readable and scalable to define the component CSS in a const outside the class, as the class can contain plenty of code supporting its behavior and lifecycle.

With this approach we benefit from some degree of segregation between style and behavior.

```
const localStyles = css`
    ...
`;

@customElement('pie-button')
    @property() actionType = 'submit';

    static styles = [utilityStyles, localStyles]

    render () {
        ...
     }
```

fernandofranca: Happy to see styling coming along! 🔥

kevinrodrigues: Good point. @xander-marjoram and myself discussed this too so we were thinking about importing it from file so we can keep it separated. Similar to what you have but within its own file instead then used with the styles array like you suggested.

packages/components/pie-button/index.html
```diff
@@ -3,11 +3,58 @@
...
+                Promise.all([bold.load(), extrabold.load()])
+                    .then(function (fonts) {
+                        fonts.forEach(function (font) {
+                            document.fonts.add(font);
+                        });
+                    })
+                    .then(function () {
+                        document.documentElement.classList.add   ('webfonts-loaded');
+                    });
+            }
```
raoufswe, line 45: nitpick, is not a big deal and can be addressed separately, Shall we use const and let here instead of vars 🥺 also a .catch at the end to log errors if they happen would be great

kevinrodrigues: I think we'd need to come back to this anyway. I've added a entry to the decision log about font loading. Right now it's using a snippet of code from the documentation we hold on it. Happy to update it here but for POC purposes I think this is fine?

packages/components/pie-button/index.html
kevinrodrigues: Fixed.

packages/components/pie-button/index.html
raoufswe: cool cool 😎

raoufswe (approved):

xander-marjoram (approved):

fernandofranca (approved):

PR #149, opened 2023-02-02
How we support you page

raoufswe:
- adds "How we support you" page


apps/pie-microsite/src/_11ty/shortcodes/list.js
```diff
@@ -1,19 +1,36 @@
...
 const listTypes = require('../../_data/listTypes');
+const pieIconsSvg = require('../filters/pieIconsSvg');
+const pieDesignTokenColours = require('../filters/pieDesignTokenColours');
 
 /**
  * A List HTML component – takes an array of list items and turns them into a marked-up list
- * @param {string} type - Type of list: ordered, pill
- * @param {string[]} items - An array of list items
+ * @param {string} type - Type of list: ordered, pill, icon
+ * @param {string[] | {iconName: 'string', iconFill?: 'string', content: 'string'}[]} items - An array of list items
```
jamieomaguire, line 9: please could we add some snapshot tests for this shortcode?

jamieomaguire, line 9: could we ask users for a colour token name rather than a fill value?

jamieomaguire: check line 1 of `header.njk` for how we already do something similar

jamieomaguire: sorry - I think you've actually done this - ignore if so :D

apps/pie-microsite/src/content/pages/designers/how-we-support-you.md
```diff
@@ -0,0 +1,130 @@
...
+
+As a result of that, we’re often asked how PIE designers are usually involved when working together with business area designers. To help describe it, we’ve put together a list of the things we do and don’t as part of our role:
+
+### What we do:
+
+{% list {
+    type: listTypes.icon,
+    items: [
+        {
+          iconName: 'CheckCircleFilledSmall',
```
jamieomaguire, line 81: do we anticipate needing a different type of icon per list item? if not we could reduce some duplication here and just specify what icon to use for the entire list

apps/pie-microsite/src/content/pages/designers/how-we-support-you.md
```diff
@@ -0,0 +1,130 @@
...
+          iconFill: 'support-positive',
```
jamieomaguire, line 82: same for fill

jamieomaguire:
Screenshot 2023-02-03 at 13 01 26

@raoufswe this is a question to relay back to Jordi please - the flow chart image I've pasted above is extremely difficult to read. I think we need to get something more legible. I'm trying to read this on a 16 inch macbook pro :)

apps/pie-microsite/src/assets/styles/components/_list.scss
```diff
@@ -51,3 +51,17 @@
...
+        display: flex;
+        gap: f.spacing();
+        margin-bottom: f.$spacing-02;
+
+        & > svg {
+            flex-shrink: 0;
+            margin-top: 2px;
+        }
+    }
+
```
jamieomaguire, line 66: NIT: can we remove this line break?

jamieomaguire (dismissed):

jamieomaguire (approved):

jordifornellsJET:
> Screenshot 2023-02-03 at 13 01 26
> 
> @raoufswe this is a question to relay back to Jordi please - the flow chart image I've pasted above is extremely difficult to read. I think we need to get something more legible. I'm trying to read this on a 16 inch macbook pro :)

This image is for purely for illustrative purposes and just an overview. The paragraph above clearly states "You can check the process closely in our Process file" which is a way for us to redirect them to a place where they can check the process closely and more easily. We want designers to check this Figma file as having another format (like an explanation in paragraphs or lists) will make it less visual and thus harder to understand (there's too many back and forth parts to this process to write it in a lineal way).

PR #171, opened 2023-02-08
pie-button add variant property and add prop validation

jamieomaguire:
Adds a new `variant` prop to `pie-button` and refactors all props to validate using Enums and use console errors when invalid values are provided.

Updated Storybook to reflect changes and use `select` controls for both props.

Use `Fozzie` for button styling


xander-marjoram (approved):

packages/components/pie-button/src/decorators.ts
```diff
@@ -0,0 +1,34 @@
+/**
+ * A decorator for specifying a list of valid values for a property.
+ * If this property's setter is called with an invalid value, an error is logged and the default value will be assigned instead.
+ * @param validValues - The array of valid values
+ * @param defaultValue - The value to fall back on
+ * @returns
+ */
+export const validValues = (validValues: any[], defaultValue: any) => {
+    return function (target: any, propertyKey: string) {
```
jamieomaguire, line 9: I think we can annotate this function as ` : void`

apps/pie-storybook/stories/pie-button.stories.ts
```diff
@@ -1,12 +1,25 @@
 import { html } from 'lit';
 import type { Meta, StoryObj } from '@storybook/web-components';
-import { PieButton } from '@justeattakeaway/pie-button';
+import { PieButton, BUTTON_TYPE, BUTTON_VARIANT } from '@justeattakeaway/pie-button';
```
FayeCarter, line 3: Should it be `BUTTON_TYPES` and `BUTTON_VARIANTS`?

packages/components/pie-button/src/button.scss
```diff
@@ -0,0 +1,3 @@
+@use '@justeat/fozzie/src/scss/fozzie' as f;
```
ashleynolan, line 1: Probably worth catching up about this separately – I personally would avoid pulling in fozzie into the web components.

It'd be quicker in the short term, but it'll mean a lot of rework down the line if we go down this route. The future will be that we write a new CSS/SCSS utility library (pie-scss/pie-css) that looks across Snacks and Fozzie and takes the best of both worlds. If we just start using fozzie now, we'll get tied into its functions and mixins and it'll mean we have to rework that down the line. We also tie ourselves into patterns like using SCSS variables instead of the PIE custom properties.

Totally fine with us pulling styles out of fozzie if we think it's better than re-doing work from scratch, but we should look to make them work aligned with the CSS Variables. In terms of things like functions we use atm, I'd just hard-code px values for now and then we can add helper functions for things like rem conversion down the line (probably as we start the modal work)

ashleynolan: Also, I'm fine with this being added for this PR (to get the variant JS added first) and then us changing the styling as a separate PR potentially, as I think the JS stuff all looks good and can be merged.

.changeset/eighty-ties-judge.md
```diff
@@ -0,0 +1,5 @@
+---
+"@justeattakeaway/pie-button": minor
+---
+
+Create decorator for validating property values
```
ashleynolan, line 5:
```suggestion
[Added] Create decorator for validating property values
```

packages/components/pie-button/src/decorators.ts
```diff
@@ -0,0 +1,34 @@
+/**
+ * A decorator for specifying a list of valid values for a property.
+ * If this property's setter is called with an invalid value, an error is logged and the default value will be assigned instead.
+ * @param validValues - The array of valid values
+ * @param defaultValue - The value to fall back on
+ * @returns
+ */
+export const validValues = (validValues: any[], defaultValue: any) => {
```
ashleynolan, line 8: Not sure if it's a bit confusing having the decorator function and param use the same name here? I'd probably call the decorator `validateValues`

apps/pie-storybook/stories/pie-button.stories.ts
jamieomaguire: It's actually quite common to go with singular naming for Enums. Here's an example of some [guidance from Microsoft](https://learn.microsoft.com/en-us/previous-versions/dotnet/netframework-1.1/4x252001(v=vs.71)?redirectedfrom=MSDN)

.changeset/eighty-ties-judge.md
xander-marjoram: Did we agree on this convention?

packages/components/pie-button/src/index.ts
```diff
@@ -1,17 +1,48 @@
...
-import styles from './styles.scss?inline';
+import styles from './button.scss?inline';
+import { BUTTON_VARIANT, BUTTON_TYPE } from './defs';
+import { validValues } from './decorators';
+
+// Valid values available to consumers
+export { BUTTON_VARIANT, BUTTON_TYPE };
+
+const validButtonVariants = Object.values(BUTTON_VARIANT);
+const validButtonTypes = Object.values(BUTTON_TYPE);
```
fernandofranca, line 12: Is there some limitation or use case that prevents us from passing the enum, so we don't need to evaluate `Object.values(...)` for each and every component prop? Those consts don't have any other role than validation.

xander-marjoram: I was discussing with @jamieomaguire yesterday about the possiblity of exporting arrays as well as the enums, so we wouldn't have to do this.

packages/components/pie-button/src/decorators.ts
fernandofranca, line 8: Curious to look at some reference material supporting this method for Lit prop validation, if you may have some at hand 🤔

packages/components/pie-button/src/button.scss
jamieomaguire: Yeah that makes sense, honestly I'm really not keen on the idea of defining the styles in the web components themselves but I appreciate how subjective this is. 

It feels to me like writing new styles for the sake of the proof of concept may not be as useful as figuring out a more scalable styling solution - maybe like you said, lift and shift existing button styles (either fozzie or snacks) into this for now? 

It would be great if this proof of concept could help inform the future direction of that potential styling library as I think they will influence each other quite a lot.

Happy to catch up offline about this :)

packages/components/pie-button/src/index.ts
jamieomaguire: We could probably move that into the decorator and just pass in the enum?

packages/components/pie-button/src/decorators.ts
jamieomaguire: Sure, [Lit suggests](https://lit.dev/docs/components/properties/#accessors-custom) adding custom getters/setters for stuff like validation and fallback value handling. An example of this in the wild can be seen in [Spectrum](https://github.com/adobe/spectrum-web-components/blob/007a762c0257af4cd74e3781d877a8a8343acfd0/packages/button/src/Button.ts#L57). Our opinion is this is messy and leads to quite bloated classes. Personally I think the decorator is far more elegant easier to understand but love to hear other people's thoughts!

packages/components/pie-button/src/button.scss
ashleynolan: Yeh, that's what I think should happen tbh – if we think we need to break down the styling into clearer tickets at this stage, I'm happy to discuss too.

The way I imagine it working (for now), would be to have a set of CSS/SCSS files alongside the component. Then we can start to write helpers for use across components once we start on our second component.

Maybe it's worth setting up a hangout to discuss styling more generally now we know a bit more about how Lit handles styles and we can work out next steps in terms of turning that into tickets.

packages/components/pie-button/src/decorators.ts
xander-marjoram: Hey, the Lit docs don't really mention property validation, and we had a look around at some of the other design systems and how they're using web components, but all of those seemed to have a lot of duplication.

But we thought that validation was required for a few reasons:

1. If someone enters an incorrect value (e.g., `size="xlarge"` which is not supported), we need to know what to do in that situation (fall back to a default value), and also to let the engineer know that they are using a bad value (via an error in the console).
2. To fulfil the first requirement, we need to be able to say which values are valid or not.

In the button component there are at least 3 properties (size, style and icon position) that require validation, and there will be plenty of examples in other components too. Most other people who are doing this seem to be writing a `get` and `set` for each property which is not very DRY and is also bad for readability in the component. Having this decorator not only keeps the logic reusable (we'll extract it out of the button at some point) but also lightweight in the component file.

I'll be doing a demo of this to the team soon, but if you have any more questions in the mean time then feel free to message me or I can show you one-on-one!

fernandofranca: The decorator solution looks super nice. Good job!

packages/components/pie-button/src/decorators.ts
```diff
@@ -0,0 +1,34 @@
...
+        Object.defineProperty(target, propertyKey, {
+            get () : any {
+                return target[privatePropertyKey];
+            },
+            set (value: any) : void {
+                const oldValue = target[privatePropertyKey];
+
+                if (!validValues.includes(value)) {
+                    console.error(
+                        `Invalid type value provided: "${value}".`,
```
fernandofranca, line 21: I'd suggest to add more context to the error message, like the component name.

Currently we have a single component, but if we imagine a scenario where the user updates several components to the newest version, it would be nice to know which component is raising the error message without effort.

xander-marjoram: Nice idea, updated.

packages/components/pie-button/src/index.ts
```diff
@@ -1,30 +1,33 @@
...
-const validButtonTypes = Object.values(BUTTON_TYPE);
-
 @customElement('pie-button')
 export class PieButton extends LitElement {
     /**
      * The Button type to use
      */
     @property()
-    @validValues(validButtonTypes, BUTTON_TYPE.SUBMIT)
+    @validPropertyValues(VALID_BUTTON_TYPES, BUTTON_TYPE.SUBMIT)
```
fernandofranca, line 23: Is there any other use case for `VALID_BUTTON_TYPES` and `VALID_BUTTON_VARIANTS` that would justify not doing 
`@validPropertyValues(Object.values(BUTTON_TYPE), BUTTON_TYPE.SUBMIT)`?

xander-marjoram: Probably not, I wanted to keep this line as clean as possible but I'll update it

ashleynolan (approved):

xander-marjoram (approved):

PR #178, opened 2023-02-14
Add size property to pie-button

xander-marjoram:
... and bring styles into the component


packages/components/pie-button/src/button.scss
```diff
@@ -1,3 +1,60 @@
-@use '@justeat/fozzie/src/scss/fozzie' as f;
+@import '@justeat/pie-design-tokens/dist/jet.scss';
```
jamieomaguire, line 1: I don't know if we want to use `@import` here... I suspect you can still use `@use` but worth checking. AFAIK `@import` is deprecated

packages/components/pie-button/src/button.scss
```diff
@@ -1,3 +1,60 @@
...
 
-@include f.buttons();
+.o-btn {
+    border-radius: $radius-rounded-e;
```
jamieomaguire, line 4: do we want to namespace the variables like we have else where i.e. `v.$radius-rounded-e` or similar?

packages/components/pie-button/src/index.ts
```diff
@@ -4,38 +4,47 @@ import { customElement, property } from 'lit/decorators.js';
...
 
 // Valid values available to consumers
-export { BUTTON_TYPE, BUTTON_VARIANT };
+export { BUTTON_SIZE, BUTTON_TYPE, BUTTON_VARIANT };
 
 @customElement('pie-button')
 export class PieButton extends LitElement {
     /**
-     * The Button type to use
+     * The size of the button. Available sizes are xsmall, small, medium (default) and large.
```
jamieomaguire, line 15: NIT: This comment feels like it's more appropriate for the enum itself. Otherwise if the enum changes we have two places to update

packages/components/pie-button/src/index.ts
```diff
@@ -4,38 +4,47 @@ import { customElement, property } from 'lit/decorators.js';
...
+            'o-btn': true,
+            [`o-btn--${size}`]: size,
             [`o-btn--${variant}`]: variant
-        });
+        };
 
         return html`
             <button
-                class="${classes}"
+                class=${classMap(classes)}
```
jamieomaguire, line 47: nice, does this solve the whitespace issue we saw?

jamieomaguire:
Are the stories on the Docs page meant to look this barebones?
Screenshot 2023-02-14 at 16 23 08

packages/components/pie-button/src/index.ts
xander-marjoram: Unfortunately not (although that's what I was hoping for), but either way it matches the docs more closely.

xander-marjoram:
@jamieomaguire I did notice the lack of styling on the default page; for some reason it's defaulting to `ghost` and I'm not sure where that's coming from. I've tried to fix it through default values on the args but for some reason those aren't being recognised. As soon as you change one of the other props (e.g., set `type` to `submit`) the styles correct themselves.

I've mentioned it on this issue here: https://github.com/storybookjs/storybook/issues/21016#issuecomment-1429938150

packages/components/pie-button/src/defs.ts
```diff
@@ -1,49 +1,29 @@
...
-    GHOST = 'ghost'
+export enum BUTTON_SIZE {
+    XSMALL = 'xsmall',
+    SMALL = 'small',
+    MEDIUM = 'medium',
+    LARGE = 'large'
 }
 
 /**
  * Button style variants
```
jamieomaguire, line 12:
```suggestion
 * Button type variants
```

apps/pie-storybook/.storybook/main.ts
```diff
@@ -14,7 +14,7 @@ const config: StorybookConfig = {
         options: {}
     },
     docs: {
-        autodocs: true
+        autodocs: false
```
xander-marjoram, line 17: The auto-generated docs page wasn't being particularly useful, and the example components weren't displaying correctly, so I've removed this for now.

jamieomaguire (approved):

fernandofranca (approved):

packages/components/pie-button/src/button.scss
```diff
@@ -1,3 +1,60 @@
-@use '@justeat/fozzie/src/scss/fozzie' as f;
+@use '@justeat/pie-design-tokens/dist/jet.scss' as dt;
 
-@include f.buttons();
+.o-btn {
```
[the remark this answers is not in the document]
jamieomaguire: @fernandofranca I believe Xander has ported the old styles over from Fozzie. This was part of our naming convention which can be [seen here](https://vue.pie.design/?path=/story/documentation-guides-fozzie-css-naming--page) - we should probably agree on a naming convention that works for everyone going forward :)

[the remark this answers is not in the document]
xander-marjoram: Not directly ported over, but I did stick with the naming convention we had in fozzie. Also happy to change this in future if we decide on a new convention.

PR #186, opened 2023-02-15
chore: watcher mechanism for typescript

LTurns:
`yarn watch` script to recompile js and ts code into the `dist` folder whenever source code is changed and saved.


.changeset/healthy-games-relate.md
```diff
@@ -0,0 +1,5 @@
+---
+"@justeattakeaway/pie-button": patch
+---
+
+[Added] watcher mechanism for running component locally
```
jamieomaguire, line 5:
```suggestion
[Added] - watcher mechanism for running component locally
```

.changeset/itchy-dodos-mate.md
```diff
@@ -0,0 +1,5 @@
+---
+"@justeattakeaway/pie-button": patch
+---
+
+[Changed] Updated README with new script
```
jamieomaguire, line 5:
```suggestion
[Changed] - Updated README with new script
```

.changeset/sweet-meals-report.md
```diff
@@ -0,0 +1,5 @@
+---
+"pie-storybook": patch
+---
+
+[Changed] Updated README to include new script
```
jamieomaguire, line 5:
```suggestion
[Changed] - Updated README to include new script
```

packages/components/pie-button/package.json
```diff
@@ -7,8 +7,13 @@
...
     "dev": "run -T vite",
+    "watch": "concurrently \"run -T vite build --watch\" \"tsup --watch\"",
     "test": "echo \"Error: no test specified\" && exit 1"
   },
   "author": "JustEatTakeaway - Design System Web Team",
-  "license": "Apache-2.0"
+  "license": "Apache-2.0",
+  "dependencies": {
+    "concurrently": "^7.6.0",
+    "tsup": "^6.6.3"
```
jamieomaguire, line 17: We probably want to remove the `^` and pin the versions. I imagine we'd want both of these dependencies at the root of the monorepo too as we'd need them for other components?

packages/components/pie-button/tsup.config.ts
```diff
@@ -0,0 +1,16 @@
+import type { Options } from 'tsup';
```
jamieomaguire, line 1: It might be worth pulling out into a shared config but we could do this when we add the next component

fernandofranca:
Will this solution handle multiple components/packages and interdependencies?

Let's say we have a fictitious `pie-card` and it depends on `pie-icon-button` which depends on `pie-button`. Will it require starting a watcher for each component?

Spinning multiple watchers can become costly (in memory consumption) and implicitly requires knowing the relationships between components.

I believe we might be able to solve this with a single watcher that handles the dependency orchestration.

packages/components/pie-button/tsup.config.ts
ashleynolan: I would def do this when we come to creating the next component – we can do this alongside the more general TS config

packages/components/pie-button/README.md
```diff
@@ -19,3 +19,8 @@ Local dev server using Vite (with hot module reloading)
 ```
 yarn dev
 ```
+
+Local watcher using tsup and Vite (watches for changes and recompiles `dist`)
```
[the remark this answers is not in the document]
LTurns: Arguably we don't need the local `dev` anymore since running `watch` alongside the `dev` script inside `pie-storybook` does the same thing, whilst also recompiling. But could be good to keep anyway as a backup 

`watch` on its own doesn't open up a browser, it just recompiles the dist so can technically be used alongside any `dev` script if needed

PR #208, opened 2023-02-20
feat(pie-button): add button state styles

raoufswe:
[Added] - state styling to button (focus/hover/active/disabled)


packages/components/pie-button/src/button.scss
```diff
@@ -2,37 +2,47 @@
...
 }
 
 // TODO - Use latest design tokens for font size and line height
-//        once fozzie has been updated to handle them.
+// once fozzie has been updated to handle them.
 
+/* NOTE: button padding-y is calculated by excluding border-width (1px) */
 .o-btn--xsmall {
-    padding: 6px dt.$spacing-b;
+    padding: 5px dt.$spacing-b;
```
xander-marjoram, line 24: This doesn't seem ideal. Is there another way around this? Could we use outline instead of border, perhaps?

raoufswe: Hey @xander-marjoram, the PR isn't ready yet and I'm still playing around with different approaches :) thanks for the heads up tho.

raoufswe: Done ✨

raoufswe: so the main issue I was trying to solve is that the `outline` variant doesn't have the same height as other variants. this is because we're adding a border, which by default is placed outside of the element content, resulting in an extra 2px. 
to overcome this issue in Snacks and Fozzie, we subtract 1px from top/bottom padding values, however, I agree with you, it feels a bit hacky. 

I think one way to have the border included in the height is to use the `box-shadow` property to simulate inner borders. [commit](https://github.com/justeattakeaway/pie/pull/208/commits/2eda74f13fc0b625ebc825918639b7cd05c29599)

Alternatively, we can use `outline` but I'm afraid we might introduce layout edge cases as they are drawn outside of the element, and some old browsers don't support rounded outlines. 👀

packages/components/pie-button/src/button.scss
```diff
@@ -38,23 +46,68 @@
...
+    &:hover {
+        background-color: darken(dt.$color-container-default, dt.$color-hover-01);
+    }
+
+    &:active {
+        background-color: darken(dt.$color-container-default, dt.$color-active-01);
+    }
 }
+
+.is-disabled {
```
jamieomaguire, line 102: Pretty sure the answer is no, but can this class affect elements outside of the web component in the DOM?

jamieomaguire: is there any reason not to use this as a modifier class i.e. `o-btn--is-disabled`?

packages/components/pie-button/src/button.scss
jamieomaguire: For what it's worth I don't think we're supporting browsers old enough to have issues with `outline` :)

packages/components/pie-button/src/button.scss
raoufswe: I think styles are scoped using shadow DOM in lit, but I'm happy to update the class name to `o-btn--is-disabled` for consistency with modes.

kevinrodrigues (approved):

raoufswe:
close in favor of https://github.com/justeattakeaway/pie/pull/234

PR #211, opened 2023-02-21
pie-docs: append -dark- for dark theme token scss variable names

jamieomaguire:
append -dark- for dark theme token scss variable names


LTurns (approved):

siggerzz (approved):

apps/pie-docs/src/_11ty/shortcodes/tokensTable/index.js
```diff
@@ -25,7 +25,13 @@ const {
  * @param {string} tokenType - the type of token i.e. color, spacing, radius
  * @returns {string} the SCSS variable name
  */
-const createScssTokenName = (tokenKey, tokenType) => `$${tokenType}-${tokenKey}`;
+const createScssTokenName = (tokenKey, tokenType, path) => {
+    // TODO: This is a little hacky and we should revisit it as part of a wider refactor
+    // of how token information is generated for the docs site
+    const isDarkToken = path.includes('dark');
```
ashleynolan, line 31: There are quite a few token names that have `dark` in their name, but are in the default theme (like `background-dark` or `container-dark`).

Would this catch those names falsely as well potentially and add this to their token name?

(I would check, but it looks like the PR preview has broken)

jamieomaguire: The path string that this checks for the word dark is a path to the token group (i.e. `color.alias.default`, `elevation.global`, `elevation.alias.dark`). So the token name won't have any bearing on this code :) 

The storybook tests should help you see the changes if the preview isn't working: https://percy.io/4bc223d1/pie-microsite/builds/25321804?utm_campaign=4bc223d1&utm_content=pie-microsite&utm_source=github_status_private

ashleynolan: Ok, cool. Just wanted to check. If that's the case, then all should be good 👍

ashleynolan (approved):

PR #225, opened 2023-02-23
feat(pie-pie-eslint-config): setup shared eslint config package

fernandofranca:
[Added] - Setup `pie-eslint-config` package
[Added] - `pie-eslint-config` base and strict rules
[Added] - `pie-eslint-config` tests
[Changed] - Set the PIE repo to use the new `pie-eslint-config` strict rules
[Removed] - Removed `eslint-config-fozzie` dependency
[Changed] - Updated `pie-docs`, `pie-icons`, and `pie-icons-vue` ESLint config
[Changed] - Lint `pie-docs`, `pie-icons`, `pie-icons-vue`, and `.eslintrc.js` source code
[Changed] - Add missing dependency on `pie-icons-vue`
[Changed] - Updated `pie-stylelint-config` `package.json` `author` field

## How to review

First, sorry about the number of changes!

Most relevant additions are in the `packages/tools/pie-eslint-config` path.

Besides that, changes on other workspaces are only to update `.eslintrc.js` or for linting code with the strict rules.


.changeset/spicy-teachers-dance.md
```diff
@@ -0,0 +1,13 @@
+---
+"@justeattakeaway/pie-eslint-config": patch
+---
+
+[Added] - Setup `pie-eslint-config` package
+[Added] - `pie-eslint-config` base and strict rules
+[Added] - `pie-eslint-config` tests
+[Changed] - Set the PIE repo to use the new `pie-eslint-config` strict rules
+[Removed] - Removed `eslint-config-fozzie` dependency
+[Changed] - Updated `pie-docs`, `pie-icons`, and `pie-icons-vue` ESLint config
```
siggerzz, line 10: Would is be possible to generate new changesets for package-specific changes so those packages get versioned correctly?

Happy to give if a hand if you need it!

siggerzz:
It might actually be worth defining this new package at the root `package.json`, rather than having it explicitly defined in each package. This way we not only ensure every package in the repo is going to use the same version, but also reduces duplication of shared deps across packages, something that can be really hard to manage as the monorepo scales. 👍🏼

.changeset/spicy-teachers-dance.md
fernandofranca: I believe it is, however, I was wondering why, since we have this in the `README`:

> Changesets are only required for Major, Minor, and Patch changes that have an effect on consumers. Changes that don't affect consumers do not need to be versioned. Examples include linting, testing or CI updates.

raoufswe (approved):

.changeset/spicy-teachers-dance.md
siggerzz: Generally I think most look okay, but some things shouldn't be added to this changeset file, as they don't relate directly to the `@justeattakeaway/pie-eslint-config` directly.

```
[Changed] - Lint `pie-docs`, `pie-icons`, `pie-icons-vue`, and `.eslintrc.js` source code
[Changed] - Add missing dependency on `pie-icons-vue`
[Changed] - Updated `pie-docs`, `pie-icons`, and `pie-icons-vue` ESLint config
```

If you still want to capture these changes, I'd recommend creating changesets for the affected packages. But as you mentioned, as they're internal changes, don't feel obliged. I personally just don't think they should be added to this changeset, as they'll get added to the release notes when this package is published 👍🏼

siggerzz:
> > It might actually be worth defining this new package at the root `package.json`, rather than having it explicitly defined in each package. This way we not only ensure every package in the repo is going to use the same version, but also reduces duplication of shared deps across packages, something that can be really hard to manage as the monorepo scales. 👍🏼
> 
> @siggerzz You mean adding it to the `dependencies`? I'll try that, but I suspect ESLint itself will complain about not having `pie-eslint-config` listed as a dependency of the other packages

Yeah you're right. I tried this locally, but honestly, it's just a nitpick, and isn't worth the added time given we're already referencing `"@justeattakeaway/pie-eslint-config": "workspace:*"`.

Happy for you to leave as is! 🎉

siggerzz (approved):

PR #232, opened 2023-02-24
chore(pie-monorepo): Only run visual tests if pie-docs has changed + pie-monorepo changeset support

siggerzz:
**pie-monorepo**
[Added] - Support for `pie-monorepo` changeset entires
[Added] - Functionality to only run visual tests against changed packages for PRs

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,30 +12,31 @@
   },
   "workspaces": [
     "apps/**/*",
-    "packages/**/*"
+    "packages/**/*",
+    "./"
   ],
   "scripts": {
-    "build": "turbo run build",
+    "build": "turbo run build --filter=!pie-monorepo",
     "changeset": "changeset",
     "changeset:publish": "changeset publish",
     "changeset:version": "changeset version",
-    "clean": "turbo run clean",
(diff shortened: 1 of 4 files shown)


JoshuaNg2332 (dismissed):

package.json
```diff
@@ -12,30 +12,31 @@
   },
   "workspaces": [
     "apps/**/*",
-    "packages/**/*"
+    "packages/**/*",
+    "./"
   ],
   "scripts": {
-    "build": "turbo run build",
+    "build": "turbo run build --filter=!pie-monorepo",
```
jamieomaguire, line 19: what does this do? :)

ashleynolan: I'm guessing it runs it for all packages except the root monorepo package?

ashleynolan (dismissed):

package.json
siggerzz: Correct. If we don't provide this, turborepo goes into a recursive loop 😭

JoshuaNg2332 (approved):

jamieomaguire (approved):

PR #233, opened 2023-02-24
chore(pie-button): watch script for auto-compiling code changes

LTurns:
running `yarn dev` in storybook and `yarn watch` from the root means any CSS and JS/TS changes to a component will compile into it's `dist` folder and be visible in storybook. Dependencies also compile - tested by adding a `hello-world` npm package to `pie-button`.

@fernandofranca tested on a mac and me on windows - all compiles 👍  turbo allows us to watch all components in parallel, so in future we'll be able to make changes to any component and they'll be reflected in storybook

Vite automatically transpiles typescript under the hood so there is actually no need for a separate watcher/transpiler


fernandofranca:
Quite happy to see an elegant solution! 👏 👏 👏 

I have one question: Is it possible to make Vite watch/build all components at once instead of only individual components?

Thinking in terms of DX, I think it would be nice to enable devs to experiment and update any component without having to start the corresponding watcher for each component in advance. The library tends to grow, and having to keep in mind the watcher instances can become an inconvenience, IMO.

JoshuaNg2332 (approved):

JoshuaNg2332 (approved):

JoshuaNg2332 (approved):

fernandofranca:
- checked the latest branch (`629664`)
- ran yarn
- started storybook (`yarn dev`)
- ran `turbo run watch`

Looks like it entered in loop or recursion, I can demonstrate in a call, or attach the video here.

image

turbo.json
```diff
@@ -7,6 +7,10 @@
         "^build"
       ]
     },
+    "watch": {
+      "cache": true,
```
fernandofranca, line 11: I'm not entirely sure if in this context wouldn't be better to set `cache: false`. I tested both options in my environment and didn't saw a big change in behavior.

PR #235, opened 2023-02-27
docs(pie-docs): Populate contributing page for engineers

xander-marjoram:


apps/pie-docs/src/engineers/contributing.md
```diff
@@ -8,4 +8,35 @@ title: Contributing
...
+- **#guild-frontend**
+- **#guild-accessibility**
+
+We also run a [PIE engineering clinic](https://calendly.com/jetpie/engineering-clinics) every other Tuesday where you can book a slot to speak to one of our engineers.
+
+Please note that the booking link only goes live the day before the clinic.
+
+### How can I make a feature request?
+
+Please use one of the workflows in the **#help-designsystem** channel to make a feature request, report a defect or request help for anything else.
```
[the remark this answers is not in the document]
jamieomaguire: too British @xander-marjoram :D

jamieomaguire (approved):

apps/pie-docs/src/assets/styles/components/_contentPage.scss
```diff
@@ -130,19 +130,14 @@
...
-
         > p + h2 {
             margin-top: f.spacing(h);
         }
 
-        ul + h3 {
-            margin-top: f.spacing(f);
-        }
-
+        ul + h3,
```
ashleynolan, line 137: Not sure if there's a better way of doing this (as I'm guessing it's due to us using markdown), but selectors like this do tend to scare me a little bit as their scope is a bit vague.

ashleynolan (approved):

apps/pie-docs/src/assets/styles/components/_contentPage.scss
xander-marjoram: I do see the potential for scary scope, but these aren't too bad IMO, it's just for `h3`s and `h4`s that directly follow some other element types (as I'm sure you can tell 😄)

jamieomaguire: it is scary, I'd probably try to limit it by only applying to immediate children using `>`. It _could_ be that there is a way to prepend elements created by the markdown renderer with a class which would solve this

apps/pie-docs/src/engineers/contributing.md
```diff
@@ -8,4 +8,38 @@ title: Contributing
...
+
+### Can I build a new component?
+
+Yes. If the component has been approved by designers then you are more than welcome to contribute.
+
+Please speak to the Design System team about your requirements and we will help by guiding you.
+
+### Can I make a custom implementation for a component that already exists?
+
+No. We suggest that you raise a feature request describing your needs and we will discuss how they can be met by using (or extending) an existing component.
```
jamieomaguire, line 45: this could be a fun opportunity to use the warning or error notifications :)

xander-marjoram: Could be fun... At the notification semantically is an `aside`, i.e., not directly related to the main content, whereas the answer is directly relevant, so it might not work as-is

apps/pie-docs/src/assets/styles/components/_contentPage.scss
xander-marjoram: I've not added this particular rule, only moved it, but I'll see if it all still behaves with the added `>`

jamieomaguire (approved):

PR #250, opened 2023-02-28
(pie-docs): Added typography pill list.

kevinrodrigues:
[Added] - Typography pill list.

### Screenshot

Screenshot 2023-02-28 at 14 38 44

Screenshot 2023-02-28 at 14 38 30


apps/pie-docs/src/_11ty/shortcodes/tokensTable/index.js
```diff
@@ -81,11 +81,26 @@ const buildTokenExampleElement = (token, tokenType, tokenMetadata, path = {}) =>
...
+        globalToken.forEach(tokenName => {
+            tokenMarkup += `<span class="c-tokensTable-token c-tokensTable-token--light c-tokensTable-token--list">${tokenName}</span>`;
+        });
+
+        tokenDescription = 'Global tokens used:';
+    } else {
+        tokenMarkup += `<span class="c-tokensTable-token c-tokensTable-token--light">${globalToken}</span>`;
+    }
+
+    return deindentHTML(`
```
jamieomaguire, line 98: I know its used like this elsewhere but I _think_ you can just indent however you like and not wrap it in this function as the entire shortcode is wrapped in one. Might be worth trying and see?

jamieomaguire (approved):

fernandofranca (approved):

apps/pie-docs/src/_11ty/shortcodes/tokensTable/index.js
kevinrodrigues: Works. Thanks. Maybe we should look at the other areas too that don't need this so we can clean it up a bit more?

fernandofranca (approved):

PR #269, opened 2023-03-02
feat(pie-docs): add footer

raoufswe:
[Added] Footer to the entire site

- [ ] If it is a core feature, I have added thorough tests.

### Reviewer 1

### Reviewer 2
- [X] If it is a `PIE Docs` change, I have reviewed the PR preview
- [X] If there are visual test updates, I have reviewed them


apps/pie-docs/src/_includes/pie-logo.njk
```diff
@@ -1 +1,26 @@
...
\ No newline at end of file
+<svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 60 24">
```
xander-marjoram, line 1: I know it was the same before your PR, but why do we need both this file and the SVG below?

raoufswe: good question, I would assume it's to render the logo as an SVG instead of img, or to avoid any extra assignments to make the import work on all pages like what I did with the jet logo [here](https://github.com/justeattakeaway/pie/pull/269/files#diff-4d23f6020146d70080f2f2d7f9344219b5c80df94d09bfaaf4f43c7a33b3d70cR6-R7), but I might be wrong, so cc @jamieomaguire :D

apps/pie-docs/src/assets/styles/components/_footer.scss
```diff
@@ -1,22 +1,89 @@
...
+.c-footer-info {
+    display: flex;
+    flex-flow: row wrap;
+    align-items: center;
+    justify-content: space-between;
+    gap: f.spacing(a);
+    padding-top: f.spacing(b);
+    border-top: 1px solid f.$color-divider-default;
+
+    .c-footer-info-ellipsis {
```
xander-marjoram, line 81: An ellipsis is specifically three dots: `...`
Maybe separator might be a better name?

apps/pie-docs/src/_data/metadata.js
```diff
@@ -3,4 +3,6 @@ module.exports = {
     url: 'https://www.pie.design/',
     language: 'en',
     description: 'PIE is the global design system for Just Eat Takeaway.com, powering all our products.',
+    lastUpdated: new Intl.DateTimeFormat('en-GB').format(),
```
xander-marjoram, line 6: I'm guessing this isn't unique per page?

jamieomaguire: we opted for a general site last updated approach for now @xander-marjoram 👍

apps/pie-docs/src/_includes/pie-logo.njk
jamieomaguire: previously we used `{% include "pie-logo.njk" %}` in both the header and footer. we can use a single approach now I think. would an inline svg be present immediately on page load vs loading an image in from the server? that could be a case for inlining on both the header and footer rather than an `` tag

jamieomaguire:
@raoufswe I actually think the logo in the header looks too small now, was that an intentional resize?

apps/pie-docs/src/_includes/pie-logo.njk
jamieomaguire: Sorry, I misunderstood the above. It looks like we can delete the `svg` file as I can't see any references in the project to it

raoufswe:
> 



> @raoufswe I actually think the logo in the header looks too small now, was that an intentional resize?

yes, was raised by @jordifornellsJET.

apps/pie-docs/src/_includes/footer.njk
```diff
@@ -1,8 +1,43 @@
...
-    </div>
-</footer>
+            <picture>
+                <source media="(max-width: 600px)" srcset={{ jetLogoSrcNarrow }}>
+                    <img src={{ jetLogoSrc }} alt="Just Eat Takeaway- Logo"></picture>
+                    <span class='c-footer-logo-divider' aria-hidden></span>
+                    {% include "pie-logo.njk" %}
+                </div>
+                <main class='c-footer-main'>
+                    {% for section in footerNavItems %}
```
jamieomaguire, line 19: @raoufswe this looks great, I think we could probably add some navigation tests here like we do for the sidebar. Might be worth chatting with @siggerzz about?

jamieomaguire (approved):

FayeCarter (approved):

apps/pie-docs/src/_includes/footer.njk
raoufswe: will reach out to @siggerzz about this 💯

PR #271, opened 2023-03-02
✨ feat(pie-docs): create card component

FayeCarter:
- [ ] If it is a core feature, I have added thorough tests.
- [X] If it is a `PIE Docs` change, I have reviewed the PR preview
- [X] If there are visual test updates, I have reviewed them properly before approving

### Reviewer 1

### Reviewer 2


apps/pie-docs/src/assets/styles/components/_card.scss
```diff
@@ -0,0 +1,67 @@
+@use '../fozzie-config' as f;
+
+$card-width: 328px;
```
raoufswe, line 3: do you think we should rename this to: 
```suggestion
$card-max-width: 328px;
```

apps/pie-docs/src/assets/styles/components/_card.scss
```diff
@@ -0,0 +1,67 @@
...
+$card-image-height: 180px;
+$card-label-height: 72px;
+
+.c-card-wrapper {
+    container-type: inline-size;
```
raoufswe, line 8: nice one ⚡ 
https://www.youtube.com/watch?v=ZSaAHb5dRwQ

apps/pie-docs/src/_11ty/shortcodes/card.js
```diff
@@ -0,0 +1,46 @@
+const pieIconsSvg = require('../filters/pieIconsSvg');
+
+/**
+ * A Card HTML component – takes an array of list items and turns them into a list of cards
+ * @param {string[]} items - An array of card items
+ * @returns {string}
+ */
```
raoufswe, line 7: the naming is quite misleading here. the name (Card) indicates an atomic component that should only render a card regardless of its context or quantity, but instead, it accepts items and renders multiple cards while its name indicates singularity.

I think ideally we should have a shortcut called card, which only builds a card and is not aware of its surroundings (don't add margins, no layout-specific styles, etc ) so we can drop it anywhere standalone, and then create another shortcut called CardList / CardsGrid / CardsWrapper which adds the layout styles [here](https://github.com/justeattakeaway/pie/pull/271/files#diff-469e20b58f0ff869d00a0e07549f804ed6704128b6816fd78b9d233f9180df79R37-R43).

this way we separate the logic of the card and its container and allow us to reuse it with different layouts like the one we have on the home page, wdyt?

jamieomaguire: I think that would be ideal @raoufswe but unfortunately I don't think you can reference a shortcode from another shortcode. It's a known limitation in Eleventy and why, when 2.0. comes out, it would be cool if we can leverage `webc` instead of shortcodes, or just use our own web component library in the future.

I think given the time constraints, I'd opt for whatever is the simplest approach here to get us to the deadline and then Faye can plan any refactors we'd like to achieve in the next initiative, which she will be leading. 

I appreciate this comes off as kicking the can down the road, but we should bear in mind this is not a component library and I don't think we need to be as stringent for the docs site components as we would for Snacks or Fozzie.

Happy to discuss further though, and it's not just my call :)

apps/pie-docs/src/assets/styles/components/_card.scss
jamieomaguire: I absolutely love that we're finding a legit use case for container queries!

Looking at [caniuse](https://caniuse.com/css-container-queries) and comparing with our [browser support guidelines](https://www.pie.design/engineers/guidelines/browser-support/) we should just triple check we're happy with the level of support, and consider whether or not we need to figure out a fallback. 

@ashleynolan would be great to hear your thoughts on this too :)

apps/pie-docs/src/assets/styles/components/_card.scss
```diff
@@ -0,0 +1,67 @@
...
+    margin: f.$spacing-f 0;
+    display: flex;
+    flex-wrap: wrap;
+    align-items: flex-start;
+    gap: f.$spacing-e;
+    list-style: none;
+    padding: 0;
+}
+
+@container (max-width: 680px) {
```
jamieomaguire, line 21: perhaps we could flip this to be mobile first, so center justified by default and then at a min of 680, we switch to the 'desktop' styling?

apps/pie-docs/src/assets/styles/components/_card.scss
ashleynolan: I checked this yesterday – I'm happy with it being used on the docs site. I don't think we should use them in any production customer components/sites yet, mainly because they only just got simplemented in by the Samsung Internet browser last week, and that's a big browser in some of our markets.

FayeCarter: 🎉

raoufswe (dismissed):

apps/pie-docs/src/_11ty/shortcodes/card.js
```diff
@@ -0,0 +1,46 @@
...
+
+    if (items.length > 1) {
+        return `<div class="c-card-wrapper">
+            <ul class="c-card-list">
+                ${Object.values(items).map((card) => `<li class="c-card">${buildCard(card)}</li>`).join('')}
+            </ul>
+        </div>`;
+    }
+
+    return `<div class="c-card">${buildCard(items[0])}`;
```
jamieomaguire, line 45: using something like `article` is quite common for cards. They're a standalone, independent chunk of content :)

apps/pie-docs/src/_11ty/shortcodes/card.js
```diff
@@ -0,0 +1,51 @@
...
+        const target = shouldOpenInNewTab ? 'target="_blank"' : '';
+
+        const labelClasses = [
+            'c-card-labelContainer',
+            src && 'c-card-labelContainer--hasImage'
+        ].filter(Boolean).join(' ');
+
+        const labelId = `link-${label.replaceAll(' ', '-').toLowerCase()}`;
+
+        return `<article class="c-card" aria-labelledby=${labelId}>
```
jamieomaguire, line 31: the `aria` attribute here causes duplication on audible screenreaders when tabbing. I think there's enough context from the link alone that we can remove this

apps/pie-docs/src/assets/styles/components/_card.scss
```diff
@@ -0,0 +1,88 @@
...
+    margin: f.$spacing-f 0;
+    display: flex;
+    flex-wrap: wrap;
+    justify-content: center;
+    gap: f.$spacing-e;
+    list-style: none;
+    padding: 0;
+}
+
+@container (min-width: 680px) {
```
jamieomaguire, line 21: out of interest can this be added into the block above much like media queries in sass?

apps/pie-docs/src/assets/styles/components/_card.scss
```diff
@@ -0,0 +1,88 @@
...
+    .c-card-list {
+        justify-content: start;
+    }
+}
+
+.c-card {
+    position: relative;
+    border: 1px solid f.$color-border-default;
+    border-radius: f.$radius-rounded-c;
+    width: 100%;
```
jamieomaguire, line 31: I expect the width 100% can be removed?

jamieomaguire: just be sure to test they don't squash down too small at any screen width

apps/pie-docs/src/assets/styles/components/_card.scss
FayeCarter: it does 🥳

apps/pie-docs/src/_11ty/shortcodes/card.js
```diff
@@ -0,0 +1,49 @@
...
+        const labelClasses = [
+            'c-card-labelContainer',
+            src && 'c-card-labelContainer--hasImage'
+        ].filter(Boolean).join(' ');
+
+        return `<article class="c-card">
+            ${src ? `<img class="c-card-image" src="${src}" role="presentation" alt="">` : ''}
+            <div class="${labelClasses}">
+                <p class="c-card-label">
+                    <a href=${href} ${target}>${label}</a>
```
jamieomaguire, line 33: I think you want double quotes about the href value

apps/pie-docs/src/_11ty/shortcodes/card.js
```diff
@@ -0,0 +1,49 @@
...
+        </article>`;
+    };
+
+    if (items.length > 1) {
+        return `<div class="c-card-wrapper">
+            <ul class="c-card-list">
+                ${Object.values(items).map((card) => `<li>${buildCard(card)}</li>`).join('')}
+            </ul>
+        </div>`;
+    }
```
jamieomaguire, line 46: for all of these functions you should be able to safely tidy up indentation

apps/pie-docs/src/assets/styles/components/_card.scss
```diff
@@ -0,0 +1,83 @@
...
+    border: 1px solid f.$color-border-default;
+    border-radius: f.$radius-rounded-c;
+    max-width: $card-max-width;
+    background-color: f.$color-container-default;
+
+    &:hover {
+        background-color: darken(f.$color-container-default, f.$color-hover-01);
+    }
+
+    // make card clickable
```
jamieomaguire, line 37:
```suggestion
    // makes the entire card clickable
```

apps/pie-docs/src/assets/styles/components/_card.scss
```diff
@@ -0,0 +1,83 @@
...
+        bottom: 0;
+        left: 0;
+    }
+}
+
+.c-card-image {
+    border-radius: f.$radius-rounded-c f.$radius-rounded-c 0 0;
+    height: $card-image-height;
+    display: block;
+    width: 100%;
```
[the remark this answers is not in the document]
FayeCarter: image
This is what I get if I remove the width 😬

apps/pie-docs/src/assets/styles/components/_contentPage.scss
```diff
@@ -170,7 +170,7 @@
         margin-bottom: f.spacing(i);
     }
 }
-.c-contentPage-content a {
+.c-content-container > p > a {
```
jamieomaguire, line 173: are all generated links on the pages nested within p tags?

apps/pie-docs/src/_11ty/shortcodes/card.js
```diff
@@ -0,0 +1,49 @@
...
+        name: 'link-external',
+        attrs: {
+            'aria-hidden': 'true',
+            height: 21,
+            width: 21,
+        },
+    });
+
+    const buildCard = ({
+        label, href, src, shouldOpenInNewTab = false,
```
jamieomaguire, line 20: label is possibly a little misleading. its the link text, right?

apps/pie-docs/src/assets/styles/components/_card.scss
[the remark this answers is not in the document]
jamieomaguire: huh, fair enough!

jamieomaguire:
Screenshot 2023-03-03 at 16 22 06
link colours look like they may have changed

apps/pie-docs/src/assets/styles/components/_contentPage.scss
jamieomaguire: I think I'd be really careful with this. what was the reason for the change?

FayeCarter: `c-contentPage-content a`overwrites the text in the cards so that it's a blue link.
Even if I put a class directly on the link text and set the text to black it stays blue

image

apps/pie-docs/src/assets/styles/components/_contentPage.scss
```diff
@@ -170,7 +170,11 @@
         margin-bottom: f.spacing(i);
     }
 }
-.c-contentPage-content a {
+
+.c-content-container > p a,
+.c-content-container > span a,
+.c-content-container > ul a,
+.c-notification-message a {
     color: f.$color-content-link-distinct;
```
raoufswe, line 178: I understand we need to do that to avoid specificity issues but I'm afraid this will get out of sync easily.

have you considered moving/nesting `.c-card-label` styles in `.c-card-labelContainer`? I think this will have a higher specificity and will apply the correct color for anchor tags without changing other files.

apps/pie-docs/src/__tests__/_11ty/shortcodes/card.test.js
```diff
@@ -0,0 +1,25 @@
+const card = require('../../../_11ty/shortcodes/card');
+
+const item = {
+    linkText: 'label',
+    href: 'link.com',
+};
```
LTurns, line 6: is it worth us adding a test for the target attribute? I know default for `shouldOpenInNewTab` is false but wonder if it's worth covering

apps/pie-docs/src/assets/styles/components/_card.scss
```diff
@@ -0,0 +1,83 @@
...
+.c-card-list {
+    margin: f.$spacing-f 0;
+    display: flex;
+    flex-wrap: wrap;
+    justify-content: center;
+    gap: f.$spacing-e;
+    list-style: none;
+    padding: 0;
+
+    @container (min-width: 680px) {
```
LTurns, line 20: should the `680px` be declared at the top with the other variables? 😄

apps/pie-docs/src/assets/styles/components/_contentPage.scss
FayeCarter: thank you!

LTurns (dismissed):

raoufswe (approved):

raoufswe (approved):

LTurns (approved):

PR #273, opened 2023-03-02
test(pie-docs): fix cookie banner test

siggerzz:


siggerzz:
Introducing this change as part of #268

PR #278, opened 2023-03-03
test(pie-docs): Fix mobile navigation

siggerzz:
**pie-docs**
- Fix mobile navigation
- Add new mobile navigation system / visual tests
- Hide 'Last Updated' in footer for Percy tests
- Wait for GHA deployment before running tests


jamieomaguire (approved):

apps/pie-docs/src/assets/js/nav-toggle-listener.js
```diff
@@ -0,0 +1,14 @@
+const navEl = document.querySelector('#navToggle');
```
ashleynolan, line 1: Is this JS just to add JS for the tests to hook onto?

Depending on how the selector engine in wdio works, there's probably a way of just checking the `c-navToggle-input:checked` as that should be either selected or not to show if the menu is open or closed. Would save having to add a JS file for the tests that way.

siggerzz: Good idea, didn't think of this!

ashleynolan (dismissed):

.changeset/strong-weeks-push.md
```diff
@@ -0,0 +1,5 @@
+---
+"pie-docs": patch
+---
+
+[Fixed] - Issue where mobile navigation toggle wouldn't close.
```
raoufswe, line 5: let's log the other changes as well

siggerzz: Given that all the other changes are internal to the repo and all test specific, I didn't add these as they don't warrant a version bump, and as a result, shouldn't be added to the changeset, as it has no impact on users 👍🏼

raoufswe (approved):

.changeset/strong-weeks-push.md
ashleynolan: Shouldn't they go in the root changelog though? (not sure if we have changesets there yet, but they should be logged in there)

siggerzz: Ah yeah, root is a good place :)

ashleynolan (dismissed):

apps/pie-docs/src/assets/styles/base/_page.scss
```diff
@@ -18,3 +18,10 @@ body {
         scroll-behavior: auto !important;
     }
 }
+
+/* CSS rules to be applied when running visual regression test */
+@media only percy {
+    .hide-in-percy {
+        visibility: hidden;
+    }
+}
```
xander-marjoram, line 27: I wonder if we could do something like this?
It _should_ switch the date for `XX/XX/XXXX`, so it's still present in the screenshot. That way, if anything else (like font-size, position, etc) changes or it gets deleted then the screenshot should still show a change.

```suggestion
@media only percy {
    .updated-date {
        visibility: hidden;
    
        &:before {
            visibility: visible;
            content: "XX/XX/XXXX";
        }
    }
}
```

siggerzz: Ah yeah, good shout!

apps/pie-docs/src/_includes/footer.njk
```diff
@@ -36,7 +36,7 @@
                     <div>
                         <a href="https://www.just-eat.co.uk/info/privacy-policy">Privacy policy</a>
                         <span class='c-footer-info-separator' aria-hidden></span>
-                        <span>Updated: {{ metadata.lastUpdated }}</span>
+                        <span class='hide-in-percy'>Updated: {{ metadata.lastUpdated }}</span>
```
xander-marjoram, line 39: You'll need to make this change as well to go with my other suggestion
```suggestion
                        <span>Updated: <span class="updated-date">{{ metadata.lastUpdated }}</span></span>
```

ashleynolan (approved):

PR #312, opened 2023-03-14
test(pie-docs): updated visual tests to support playwright

JoshuaNg2332:
- [x] If it is a core feature, I have added thorough tests.

### Reviewer 1

### Reviewer 2


.github/workflows/build.yml
```diff
@@ -268,6 +268,12 @@ jobs:
         if: contains(github.ref_name, 'main')
         run: |
           yarn test:visual:ci
+      - uses: actions/upload-artifact@v3
+        if: always()
+        with:
+          name: visual-report
+          path: visual-report/
+          retention-days: 3
```
xander-marjoram, line 276: Does this mean that Friday builds will only last until Monday? I think we could up this to a week unless there are any other issues

JoshuaNg2332: Fair point. Will update to 7.

playwright-visual.config.ts
```diff
@@ -0,0 +1,77 @@
+import type { PlaywrightTestConfig } from '@sand4rt/experimental-ct-web';
```
xander-marjoram, line 1: This looks like a pretty new and fairly niche package, do we know how reliable it's going to be?

siggerzz: There's currently an open PR in playwright to absorb this package - https://github.com/microsoft/playwright/issues/20076

This packages functionality / API is identical to the React / Vue component testing libraries offered by Playwright as seen here:
https://github.com/microsoft/playwright/tree/main/packages/playwright-ct-vue
https://github.com/microsoft/playwright/tree/main/packages/playwright-ct-react

I'm pretty confident this will get adopted eventually, it's just due to Lits infancy. Worst case, if this functionality is never natively adopted by Playwright themselves, we can just test in Storybook like we do in Fozzie Components.

playwright-visual.config.ts
```diff
@@ -0,0 +1,77 @@
...
+     */
+    timeout: 5000
+  },
+  /* Run tests in files in parallel */
+  fullyParallel: true,
+  /* Fail the build on CI if you accidentally left test.only in the source code. */
+  forbidOnly: !!process.env.CI,
+  /* Retry on CI only */
+  retries: process.env.CI ? 2 : 0,
+  /* Opt out of parallel tests on CI. */
```
xander-marjoram, line 34: This comment doesn't seem to align with the code, was there an if/ternary here before?
Also, looks like it's gonna try and use every CPU, is that likely to be an issue?
And... if that's not an issue, could we just specify `'100%'` instead?

playwright-visual.config.ts
```diff
@@ -0,0 +1,77 @@
...
+  workers: os.cpus().length,
```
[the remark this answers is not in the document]
siggerzz: Doesn't seem to be an issue on CI, however you raise a good point when testing locally. We'll modify this to use half of the available CPU cores to avoid local machines grinding to a halt!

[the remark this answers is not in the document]
xander-marjoram: Nice, the docs have an example of that actually, doing something like this:
```javascript
workers: process.env.CI ? 2 : undefined,
```
`undefined` locally meaning fall back to the default

playwright-visual.config.ts
siggerzz: We're going to temporarily remove this package, and introduce it as part of -https://github.com/justeattakeaway/pie/pull/303

apps/pie-docs/test/playwright/fixtures.js
```diff
@@ -6,13 +6,13 @@ const AxeBuilder = require('@axe-core/playwright').default;
 // This new "test" can be used in multiple test files, and each of them will get
 // a consistently configured AxeBuilder instance.
 exports.test = base.test.extend({
-    makeAxeBuilder: async ({ page }, use) => {
+    makeAxeBuilder: [async ({ page }, use) => {
```
xander-marjoram, line 9: What's with the array?

playwright-browsers.config.ts
```diff
@@ -31,10 +30,10 @@ const config: PlaywrightTestConfig = {
   forbidOnly: !!process.env.CI,
   /* Retry on CI only */
   retries: process.env.CI ? 2 : 0,
-  /* Opt out of parallel tests on CI. */
-  workers: os.cpus().length,
+  /* All CPUs on CI / half of availible CPUs when testing locally. */
+  workers: process.env.CI ? os.cpus().length : os.cpus().length / 2,
```
xander-marjoram, line 34: I think this syntax should work and it is more readable
```suggestion
  workers: process.env.CI ? '100%' : '50%',
```

playwright-browsers.config.ts
```diff
@@ -31,10 +30,10 @@ const config: PlaywrightTestConfig = {
   forbidOnly: !!process.env.CI,
   /* Retry on CI only */
   retries: process.env.CI ? 2 : 0,
-  /* Opt out of parallel tests on CI. */
-  workers: os.cpus().length,
+  /* All CPUs on CI / half of availible CPUs when testing locally. */
```
xander-marjoram, line 33:
```suggestion
  /* All CPUs on CI / half of availiable CPUs when testing locally. */
```

apps/pie-docs/test/playwright/fixtures.js
siggerzz: Previously the `timeout` value would've applied to all test fixtures defined. This array means that the timeout only applies to the `makeAxeBuilder` function 👍🏼

playwright-browsers.config.ts
siggerzz: Are these valid values?

xander-marjoram: Looks like it from the docs: https://playwright.dev/docs/test-configuration#testing-options

apps/pie-docs/test/playwright/fixtures.js
xander-marjoram: Do you have a link to the docs where it shows it'll work when passed an array like this? The ones I've managed to find don't really show it

siggerzz: Here ya go! - https://playwright.dev/docs/test-fixtures#worker-scoped-fixtures

```
  account: [async ({ browser }, use, workerInfo) => {
    // Unique username.
    const username = 'user' + workerInfo.workerIndex;
    const password = 'verysecure';

    // Create the account with Playwright.
    const page = await browser.newPage();
    await page.goto('/signup');
    await page.getByLabel('User Name').fill(username);
    await page.getByLabel('Password').fill(password);
    await page.getByText('Sign up').click();
    // Make sure everything is ok.
    await expect(page.locator('#result')).toHaveText('Success');
    // Do not forget to cleanup.
    await page.close();

    // Use the account value.
    await use({ username, password });
  }, { scope: 'worker' }],
```

playwright-visual.config.ts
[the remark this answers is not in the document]
siggerzz: Fixed!

xander-marjoram (approved):

apps/pie-docs/test/visual/navigation.spec.js
```diff
@@ -1,46 +1,44 @@
...
+test.describe('PIE - Page Visual Tests', () => {
+    test.beforeEach(async ({ page, context, baseURL }) => {
+        await page.goto(baseURL);
+        await disableCookieBanner(page, context, true);
     });
 
     expectedRoutesJson.forEach((route) => {
-        it(`Should respond take a screenshot of the requested route: - ${route}`, async () => {
-            const url = `${browser.options.baseUrl}/${route}`;
+        test(`Should respond take a screenshot of the requested route: - ${route}`, async ({ page }) => {
```
LTurns, line 13: super nit but I don't think you need the word `respond` in this test?

LTurns (approved):

xander-marjoram (approved):

LTurns (approved):

apps/pie-docs/test/visual/navigation.spec.js
JoshuaNg2332: @xander-marjoram you missed one. Get good.

xander-marjoram: I'm so proud.

playwright-visual.config.ts
sand4rt: Hey guys, i'm the maintainer of the lib and i contributed to Playwright component testing for the other frameworks as well. Glad to see you want to use it! It should work straight forward, but let me know if you have any questions.

PR #346, opened 2023-03-21
feat(wc-next13): add button to Next.js app

fernandofranca:
### wc-next13

[Added] - `pie-button` to app
[Added] - Counter for testing click event handlers and toggle buttons for displaying different variants
[Added] - Fonts markup
[Added] - Volta settings
[Added] - ESLint settings
[Added] - npm script for building a static app

### pie-button

[Updated] - Register approach for button web component
[Added] - "types" field in package.json for better DX and IDE TypesScript support
[Updated] - Provide Vite settings for generating TypeScript declaration files during build

### defineCustomElement as replacement of customElement decorator

I added this decorator to bypass an issue I was facing with Next.js, and it'd likely affect the experience of other developers.

- when a change was introduced in the Next.js app, HMR automatically updated the browser content
- reloading the browser tab after introducing any change leads to a crash

The crash happens because Next.js keeps the previous app state of its development server, and the `pie-button` tries to re-register the web component. The API for custom elements does not allow such thing and it will trigger an error.

> `Failed to execute 'define' on 'CustomElementRegistry': the name "pie-button" has already been used with this registry`.

The new decorator uses the `customElement` decorator function from lit as its base and verifies if the component was already defined before trying to re-register it. 

I also tested the button with the new decorator with the Vue 3 and React app.


FayeCarter (approved):

.changeset/six-boats-yell.md
```diff
@@ -0,0 +1,10 @@
+---
+"wc-next13": patch
```
jamieomaguire, line 2: I don't think integrating PIE Button would be a patch. I think it would be minor.

apps/examples/wc-next13/pages/api/hello.ts
```diff
@@ -1,13 +1,10 @@
...
+    name: string
 }
 
-export default function handler(
-  req: NextApiRequest,
-  res: NextApiResponse<Data>
-) {
-  res.status(200).json({ name: 'John Doe' })
+export default function handler (req: NextApiRequest, res: NextApiResponse<Data>) {
+    res.status(200).json({ name: 'John Doe' });
```
jamieomaguire, line 9: is this needed for the purpose of a WC demo?

jamieomaguire:
Awesome stuff @fernandofranca! I've just added a few questions 😃

FayeCarter (approved):

apps/examples/wc-next13/pages/index.tsx
```diff
@@ -1,21 +1,76 @@
-import Head from 'next/head'
-import Image from 'next/image'
-import { Inter } from 'next/font/google'
+import React, { useState, useRef } from 'react';
+import Head from 'next/head';
+import { createComponent } from '@lit-labs/react';
+import { PieButton, BUTTON_SIZE, BUTTON_VARIANT } from '@justeattakeaway/pie-button';
 
-const inter = Inter({ subsets: ['latin'] })
+const Button = createComponent({
```
jamieomaguire, line 6: Would every React consumer have to do this the same way?

If so, could we export our components with the React wrapper already applied for React consumers?

jamieomaguire (approved):

PR #348, opened 2023-03-22
fix(pie-icons): filename fixes to a few icons

ashleynolan:

[Fixed] - Capitalisation of some SVG assets in github now changed to lowercase
[Fixed] - Sandwhich renamed to Sandwich (typo)


JoshuaNg2332 (approved):

FayeCarter (approved):

.changeset/beige-pumpkins-brush.md
```diff
@@ -0,0 +1,7 @@
+---
+"@justeattakeaway/pie-icons-react": minor
+"@justeattakeaway/pie-icons-vue": minor
+"@justeattakeaway/pie-icons": minor
+---
+
+[Fixed] - Sandwhich renamed to Sandwich (typo)
```
jamieomaguire, line 7: I wonder if we could run a spell-check extension against the icon list in the docs site and find any more typos (or something smarter)

.changeset/beige-pumpkins-brush.md
```diff
@@ -0,0 +1,7 @@
+---
+"@justeattakeaway/pie-icons-react": minor
+"@justeattakeaway/pie-icons-vue": minor
+"@justeattakeaway/pie-icons": minor
```
jamieomaguire, line 4: is renaming an icon a breaking change for existing consumers?

jamieomaguire (approved):

.changeset/beige-pumpkins-brush.md
ashleynolan: Usually it would be, but this package isn't used anywhere yet, so I've taken a bit of a liberty with it and just minor bumped it

.changeset/beige-pumpkins-brush.md
ashleynolan: Yeh, I don't know. The easiest way may be to spellcheck the index files created by the react/vue packages potentially, as they output all the names in code (although would have to do some normalisation of name prefixes to do that)

JoshuaNg2332 (approved):

PR #364, opened 2023-03-28
feat(pie-button): add slot to pie button

jamieomaguire:
[Added] - Slot functionality to pie-button component and associated SB and example app integrations

### DIFF
--- a/apps/examples/wc-next13/pages/index.tsx
+++ b/apps/examples/wc-next13/pages/index.tsx
@@ -3,7 +3,7 @@ import Head from 'next/head';
 import { createComponent } from '@lit-labs/react';
 import { PieButton, BUTTON_SIZE, BUTTON_VARIANT } from '@justeattakeaway/pie-button';
 
-const Button = createComponent({
+const PieBtn = createComponent({
     tagName: 'pie-button',
     elementClass: PieButton,
     react: React,
@@ -38,20 +38,20 @@ export default function Home () {
                 <div>
                     <h3>Counter</h3>
                     <div className="flex-wrapper">
-                        <Button onClick={decrement}>decrement</Button>
+                        <PieBtn onClick={decrement}>decrement</PieBtn>
                         <div className="padding">
                             Counter: { count }
                         </div>
-                        <Button onClick={increment}>decrement</Button>
+                        <PieBtn onClick={increment}>decrement</PieBtn>
                     </div>
                 </div>
                 <h2>Test props</h2>
                 <div>
                     <h3>Click the button to switch the variant</h3>
                     <div className="flex-wrapper">
-                        <Button variant={variantName} onClick={switchVariant}>
+                        <PieBtn variant={variantName} onClick={switchVariant}>
                             Switch variant
-                        </Button>
+                        </PieBtn>
                         <div className='padding'>Variant: <b>{variantName}</b></div>
                     </div>
                 </div>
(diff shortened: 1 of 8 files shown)


apps/examples/wc-next13/pages/index.tsx
```diff
@@ -3,7 +3,7 @@ import Head from 'next/head';
 import { createComponent } from '@lit-labs/react';
 import { PieButton, BUTTON_SIZE, BUTTON_VARIANT } from '@justeattakeaway/pie-button';
 
-const Button = createComponent({
+const PieBtn = createComponent({
```
kevinrodrigues, line 6: Would `ButtonComponent` be a little better thinking about it now? As it's a result of the create component method and might read better?

jamieomaguire: I personally feel like the word `component` is redundant. So it comes down to `Button` or `PieButton`, of which I think `button` is too generic/ambiguous :)

kevinrodrigues (dismissed):

packages/components/pie-button/src/index.ts
```diff
@@ -32,7 +32,7 @@ export class PieButton extends LitElement {
     @validPropertyValues(Object.values(BUTTON_VARIANT), BUTTON_VARIANT.PRIMARY)
     variant : BUTTON_VARIANT = BUTTON_VARIANT.PRIMARY;
 
-    @property()
+    @property({type: Boolean, reflect: true})
```
fernandofranca, line 35: I re-read the docs and feel that it's not 100% clear what `reflect` does and why it is necessary here. Have you found any issues without it?

fernandofranca (approved):

packages/components/pie-button/src/index.ts
jamieomaguire: Hey, yes I've found the documentation unclear as well. Adding the `reflect` property to this appears to get it correctly working reactively in Storybook when we use the new templating function added. I have looked at some other design systems built with Lit and they appear to also use this approach for `boolean` component properties. [Example](https://github.com/axa-ch-webhub-cloud/pattern-library/blob/aeceed8b925c156570980eea0b65a3941bfc08b7/src/components/10-atoms/button/index.js#L36)

fernandofranca: Something I find curious from that example is how only 3 out of 10 props use `reflect` and a certain lack of clear relationship. It's not a type thing, since one is string, and the others are boolean. Also doesn't look like it relates to native attributes.

kevinrodrigues: I'm just wondering if we should add this value now, i feel as if we need it but only so storybook works correctly which seems wrong. The reason we needed it was so that storybook will pick up property changes (a bit like a watcher) via the controls panel but if we've added it purely for storybook it feels a bit wrong as the component passed with the correct props will render correctly in a consuming app and for the button it is never really a dynamically changing component if it is then we should probably add the reflect attribute to only those properties that are required.

ashleynolan: Just to ask – is `reflect` a Lit only property? Could be something we ask the Lit team if it's something we think is a little unclear (I can reach out to them on Discord and see if we can setup that private channel)

ashleynolan (approved):

fernandofranca (approved):

PR #387, opened 2023-04-11
feat(pie-monorepo): add pie-button to wc-next10 example app 

LTurns:

[Added] - pie-button to example app

note: SSR will be covered in a later ticket, this is just to get the component rendering :)


apps/examples/wc-next10/components/PieButton.js
```diff
@@ -0,0 +1,76 @@
...
+import { PieButton, BUTTON_SIZE, BUTTON_VARIANT } from '@justeattakeaway/pie-button';
+
+const PieBtn = createComponent({
+    tagName: 'pie-button',
+    elementClass: PieButton,
+    react: React,
+    events: { onCustomEvent: 'CustomEvent' },
+});
+
+export default function PButton () {
```
jamieomaguire, line 13: We could do with breaking down this file. This is supposed to be a react component wrapping around our web component (as I understand it). However, we appear to be returning the entire application shell (the main index.html content).

apps/examples/wc-next10/next.config.js
```diff
@@ -1,6 +1,9 @@
...
+const transpileModules = [
+    '@justeattakeaway/pie-button',
+    '@lit-labs/react',
+    'lit'
+];
 
-module.exports = nextConfig
+const withTM = require('next-transpile-modules')(transpileModules);
+
+module.exports = withTM;
```
ashleynolan, line 9: If you wanted to include some Next config alongside this transpilation, what would you need to do?

ashleynolan: It looks like you could do this potentially (from a v quick look):

`withTM({ … }`

Would be worth a check with a simple Next 10 option though and make sure you can pass through config in that way (and if so, just adding that to the README)

LTurns: You can use lodash/flow to return transpilation along with the next config -
 ```
return flow([withTM])(
      nextConfig,
    );
```

File: apps/examples/wc-next10/components/PieButton.js [hunk 3]
```diff
@@ -0,0 +1,68 @@
...
+    const switchVariant = () => {
+        variantIndex.current += 1;
+        const variantsKeys = Object.keys(BUTTON_VARIANT);
+        const variantKey = variantsKeys[variantIndex.current % variantsKeys.length];
+        const variant = BUTTON_VARIANT[variantKey];
+
+        setVariantName(variant);
+    };
+
+    return (
```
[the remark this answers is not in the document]
jamieomaguire: Screenshot 2023-04-13 at 10 33 37

It means this looks like it's just adding a button but it's actually adding headings and all sorts :)

kevinrodrigues (approved):

apps/examples/wc-next10/ButtonLogic.js
```diff
@@ -0,0 +1,68 @@
...
+    events: { onCustomEvent: 'CustomEvent' },
+});
+
+export default function PButton () {
+    const variantIndex = useRef(0)
+    const [count, setCount] = useState(0);
+    const [variantName, setVariantName] = useState(BUTTON_VARIANT.PRIMARY);
+
+    const increment = () => setCount(count + 1);
+    const decrement = () => setCount(count - 1);
```
raoufswe, line 18: in React, it's quite common to prefix callbacks with `on` to indicate a function will be used as a callback handler for a click or key up events.

```suggestion
    const onIncrement = () => setCount(count + 1);
    const onDecrement = () => setCount(count - 1);
```

apps/examples/wc-next10/pages/index.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import Head from 'next/head';
+import dynamic from 'next/dynamic'
+
+const PButton = dynamic(() => import('../ButtonLogic'), { ssr: false });
```
raoufswe, line 5: is there a reason why we are setting ssr to false here?

apps/examples/wc-next10/pages/index.jsx
```diff
@@ -0,0 +1,20 @@
...
+
+export default function Home() {
+    return (
+        <>
+            <Head>
+                <title>Create Next App</title>
+                <meta name="viewport" content="width=device-width, initial-scale=1" />
+                <link rel="icon" href="/favicon.ico" />
+            </Head>
```
raoufswe, line 14: I believe that in nextjs, you can specify the head globally for all pages in _document.jsx as this will only attach the title meta to the index page "/" only. i suggest we move it to `_document.jsx` just in case we add a new page in the fiture.

apps/examples/wc-next10/pages/index.jsx
LTurns: yep, if I set this to true I get a `window is not defined` error I think because it tries to render the component before a window exists. I've looked into this and it was between this option or using the UseEffect hook? In the ticket it said we'd look into ssr later down the line so I thought I'd go with this option to get it rendering for now. Happy to jump on a call if you have any other suggestions though 😄

raoufswe: I see you already mentioned the reason on top, but that means the button will not render on the server. is this expected for now?

raoufswe: I see, I think your solution is the cleanest. but my main concern is that we aren't SSR-ing the component which might be a bottleneck for many of our apps. happy to continue the conversation down the line when we need to look into this but from quick research, it seems the lit team has just released a wrapper for nextjs that we might want to consider. 

https://github.com/lit/lit/tree/main/packages/labs/nextjs

jamieomaguire: @raoufswe I believe handling SSR is a separate piece of work coming up very soon :)

apps/examples/wc-next10/pages/index.jsx
```diff
@@ -0,0 +1,20 @@
...
+            <main>
+                <PButton />
```
jamieomaguire, line 16: Here's what I'd suggest to clean this up (I'm sorry it's gone on over a few comments):

Rename `ButtonLogic.js` to `Counter.js`. Rename the exported component from `PButton` to `Counter`. 

I am happy to hop on a call and explain my reasoning a little more if you'd like! But in a nut shell, even though this is an example app I think we should try to keep the architecture clean and reasonably close to what we'd expect from an actual NextJS application.

_Caveat: I am not super experienced with Next/React so perhaps @raoufswe could weigh in and validate/disagree with my suggestion? 😂 _

Then in here just use `<Counter />`

raoufswe: I agree, I also think renaming it to `Counter` encapsulates what the component actually does, not, what element it uses.

jamieomaguire (dismissed):

jamieomaguire (approved):

JoshuaNg2332 (approved):

PR #402, opened 2023-04-20
feat(pie-monorepo): [Added] react 17 empty project

kevinrodrigues:
[Added] wc react 17 app


apps/examples/wc-react17/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "react-scripts": "5.0.1"
+  },
+  "scripts": {
+    "dev": "react-scripts start",
+    "build": "react-scripts build"
+  },
+  "eslintConfig": {
+    "extends": [
+      "react-app"
+    ]
```
jamieomaguire, line 17: I think the other example apps are using the JET eslint package

jamieomaguire (approved):

apps/examples/wc-react17/package.json
kevinrodrigues: Good spot!

FayeCarter (approved):

fernandofranca (approved):

PR #426, opened 2023-04-28
feat(pie-wrappers-react): added react wrapper for lit components

LTurns:
**"pie-monorepo": minor**

[Removed] - react `createComponent` wrappers from react/next apps
[Added] - `custom-elements.json` and `custom-elements-manifest.config.js` for converting custom-elements into JSON
[Added] -  `cem-plugin-module-file-extensions` and `@custom-elements-manifest/analyzer` to package.json

**"@justeattakeaway/pie-wrapper-react": minor**

[Added] - react wrapper package for wrapping lit components

**"@justeattakeaway/pie-button": minor**

[Changed] - build script to include generating react wrapper

The aim is for react users to import the lit component without having to first wrap it around the `createComponent` function. The other aim was to automate this process on our side to reduce the amount of effort and time needed to apply this to other future components

ticket referenced a very useful example used by adobe: https://github.com/adobe/spectrum-web-components/pull/2745/files
tested by adding an additional `pie-label` component with different events and testing with PieModal & PieButton 👍


custom-elements-manifest.config.mjs
```diff
@@ -0,0 +1,14 @@
+import { moduleFileExtensionsPlugin } from 'cem-plugin-module-file-extensions';
```
jamieomaguire, line 1: Should this live in the project root? It seems like it should live in a dedicated folder closer to the components, in my opinion.

custom-elements.json
```diff
@@ -0,0 +1,316 @@
+{
+  "schemaVersion": "1.0.0",
```
jamieomaguire, line 2: Why is this file here, but also listed in .gitignore? :)

package.json
```diff
@@ -98,8 +99,10 @@
     "vue-template-compiler": "2.7.14"
   },
   "dependencies": {
+    "@custom-elements-manifest/analyzer": "^0.8.0",
```
jamieomaguire, line 102: do we want to pin these versions?

packages/components/pie-button/package.json
```diff
@@ -1,13 +1,15 @@
...
   "version": "0.10.2",
   "description": "PIE design system button built using web components",
   "type": "module",
   "main": "dist/index.js",
   "module": "dist/index.js",
-  "types": "dist/types/index.d.ts",
   "scripts": {
-    "build": "run -T vite build",
+    "build": "yarn create-react-wrapper && run -T vite build && yarn remove-wrapper",
+    "create-react-wrapper": "cd ../../tools/pie-wrappers-react && yarn build:wrapper",
```
jamieomaguire, line 11: is there a cleaner way to do this that doesn't require relative paths? just wondering if we'd notice this fail if a folder structure changed

packages/components/pie-button/package.json
```diff
@@ -1,13 +1,15 @@
...
+    "remove-wrapper": "cd ../../tools/pie-wrappers-react && yarn remove:wrapper",
```
jamieomaguire, line 12: presumably, we'll be doing this for every web component. Is there any way to prevent adding these commands to every package.json file we create in future?

packages/components/pie-button/package.json
```diff
@@ -20,5 +22,10 @@
   "license": "Apache-2.0",
   "volta": {
     "extends": "../../../package.json"
+  },
+  "dependencies": {
+    "@lit-labs/react": "1.1.1",
+    "react": "18.2.0",
+    "react-dom": "18.2.0"
```
jamieomaguire, line 29: should these be dependencies, dev dependencies or peer dependencies?

packages/components/pie-button/src/index.ts
```diff
@@ -37,9 +36,8 @@ export class PieButton extends LitElement {
         };
 
         const raiseWCEvent = () => {
-            const event = new CustomEvent('CustomEvent', { detail: 'WC event dispatched' })
             console.info('WC event dispatched')
-            this.dispatchEvent(event)
+            this.dispatchEvent(new CustomEvent('onCustomEvent', { detail: 'WC event dispatched' }))
```
jamieomaguire, line 40: just my personal preference but I think defining the event before dispatching it is a bit more readable

packages/components/pie-button/src/index.ts
```diff
@@ -56,6 +54,8 @@ export class PieButton extends LitElement {
     static styles = unsafeCSS(styles);
 }
 
+customElements.define('pie-button', PieButton);
```
jamieomaguire, line 57: where is `customElements` defined? is that a browser property?

packages/components/pie-button/src/index.ts
jamieomaguire: I don't think the event should be called `onCustomEvent`. That sounds more like a handler function that would respond to `CustomEvent`. What's the need for the change?

packages/components/pie-button/test/component/pie-button.spec.ts
```diff
@@ -41,7 +41,7 @@ test('should emit an event when clicked', async ({ mount }) => {
         default: 'Click me!'
       },
       on: {
-        CustomEvent: (data: string) => messages.push(data),
+        onCustomEvent: (data: string) => messages.push(data),
```
jamieomaguire, line 44: Maybe there's something I'm missing but again, the name change causes this to read in a bit of a confusing way in my opinion. `on: onCustomEvent` seems a little strange (I could be completely missing something here)

packages/tools/pie-wrappers-react/scripts/react-remove-wrapper.js
```diff
@@ -0,0 +1,21 @@
+import { readFile, writeFile } from 'fs';
+import { components } from './react-gen-wrapper.js';
+
+// removes react wrapper from index.ts after dist has built
+function removeReactWrapper (component) {
```
jamieomaguire, line 5: I think we need a suite of tests for both of these files

custom-elements-manifest.config.mjs
LTurns: yeah I wasn't sure actually - I'll try and add it to the react-wrapper instead which lives inside `tools`. I wasn't sure if we'd want to use it for anything else, but it does seem a bit random in the root since it's web-component specific

custom-elements.json
LTurns: no idea 😂 just googled this and it's likely because I pushed up the file before adding it to gitignore, so I just need to fix that in my local git history

packages/components/pie-button/src/index.ts
LTurns: yep this is a browser property 😄 https://lit.dev/docs/components/defining/

packages/components/pie-button/src/index.ts
LTurns: I agree with the `onCustomEvent` naming, it should be separate to the handler - I've reverted that back and updated the wrapper to match 👍

In terms of the dispatch, unfortunately defining the event before dispatching means the event isn't read by the `custom-elements-manifest` so isn't added to the `custom-elements.json`

packages/components/pie-button/test/component/pie-button.spec.ts
LTurns: fixed by reverting this change 👍

packages/components/pie-button/package.json
LTurns: I've actually just tested with workspaces and yalc and everything works without these dependencies 😄  Since these dependencies actually exist in the package.json of the react example apps anyway we probably don't need them here too - have removed 👍

packages/components/pie-button/package.json
LTurns: see comment below 👍

packages/tools/pie-wrappers-react/scripts/react-gen-wrapper.js
```diff
@@ -0,0 +1,77 @@
+import fs from 'fs-extra';
+import { appendFile } from 'fs'
+
+// fetches custom-elements.json file from the root
+const loadJSON = (path) => JSON.parse(fs.readFileSync(new URL(path, import.meta.url)));
```
fernandofranca, line 5: First time I see the usage of the URL constructor for reading local files. 

Are there any advantages over `path.join()` or `path.resolve()`?

In the PIE codebase I see the usage of `path.resolve` as well as `path.join`. I'd suggest sticking to a single approach for the whole repo, if possible.

I prefer `path.resolve` as it returns an absolute path. It makes debugging easier and doesn't have any downsides.

packages/tools/pie-wrappers-react/scripts/react-gen-wrapper.js
```diff
@@ -0,0 +1,77 @@
...
+    }
+})
+
+/**
+ * function to fetch events from the component
+ *
+ * @param {*} component object from within components array
+ * @param {*} event empty array to be populated with events
+ */
+function getEvents(component, events) {
```
fernandofranca, line 30: If there's no other reason for passing `events` as an argument, I'd embed it's initialization inside the `getEvents` function, and return its value at the function end.

```
function getEvents(component) {
  const events = [];
  ...
  return events;
}

components.forEach(component => {
  const events = getEvents(component)
  ...
}
```

packages/tools/pie-wrappers-react/scripts/react-gen-wrapper.js
```diff
@@ -0,0 +1,77 @@
...
+                    pre +
+                    `${'on' + event.name}: '${event.name}' as EventName<${event.type}>, ${
+                        event.description ? `// ${event.description}` : ''
+                    }`,
+                ''
+            )}
+        }
+    });`
+
+    appendFile(`../../components/${component.tagName}/src/index.ts`, componentSrc, (err) => {
```
fernandofranca, line 74: Since we're defining the `componentSrc` from scratch is there any reason for appending more content instead of simply using `writeFileSync`?

packages/tools/pie-wrappers-react/scripts/react-remove-wrapper.js
```diff
@@ -0,0 +1,21 @@
...
+    readFile(`../../components/${component.tagName}/src/index.ts`, { encoding: 'utf8', flag: 'r' }, (err, data) => {
```
fernandofranca, line 6: This can be simplified by dropping the callback, since we can use `readfileSync`.

apps/examples/wc-next10/ButtonSizes.js
```diff
@@ -21,7 +13,7 @@ export default function ButtonSizes () {
                                 const size= BUTTON_SIZE[key];
 
                                 return (
-                                    <PieBtn size={size} key={key}>{key.toLowerCase()}</PieBtn>
+                                    <PButton size={size} key={key}>{key.toLowerCase()}</PButton>
```
ashleynolan, line 16: More a question than anything else – is this the naming convention we've agreed to go with when importing (or showing the import) of components? Not sure if others have any thoughts on this, but would be good to document this somewhere if we have come to some sort of consensus.

/cc @xander-marjoram @fernandofranca ?

packages/tools/pie-wrappers-react/README.md
```diff
@@ -0,0 +1,31 @@
+# pie-wrappers-react
+
+## Usage
+
+This package is for generating react wrappers during the build process of a lit component. Previously, react developers would need to install the component, then manually wrap it within a `createComponent` function before it could be used. This is because React 18 and previous versions don't handle web components and custom elements out of the box correctly, due to how React treats custom props and events.
```
fernandofranca, line 5: Minor nit: React and Lit spelling with Title case.

packages/tools/pie-wrappers-react/scripts/react-gen-wrapper.js
fernandofranca: nvm, just figured it!

packages/tools/pie-wrappers-react/scripts/react-remove-wrapper.js
```diff
@@ -0,0 +1,21 @@
+import { readFile, writeFile } from 'fs';
+import { components } from './react-gen-wrapper.js';
```
fernandofranca, line 2: ESLint is complaining about the file extension, it can be removed.

packages/tools/pie-wrappers-react/scripts/react-gen-wrapper.js
```diff
@@ -0,0 +1,77 @@
...
+            );
+        }
+    }
+
+// create wrapper src code and add to index.ts file
+components.forEach(component => {
+    const events = []
+    getEvents(component, events)
+
+    const componentSrc = `import * as React from 'react';
```
fernandofranca, line 51: Can you add a couple of line breaks before the `import`?

The temporary code looks a bit squeezed.

```
...
declare global {
    interface HTMLElementTagNameMap {
        'pie-button': PieButton;
    }
}import * as React from 'react';
...
```

packages/tools/pie-wrappers-react/scripts/react-remove-wrapper.js
```diff
@@ -0,0 +1,21 @@
...
+            writeFile(`../../components/${component.tagName}/src/index.ts`, fileData, () => {
+                if (err) throw (err);
+            });
+        }
+    });
+}
+
+if (components.length > 0) {
+    components.forEach((component) => removeReactWrapper(component));
+}
```
fernandofranca, line 21: Wouldn't be part of the clean-up process to remove the generated `custom-elements.json`?

LTurns: we could remove it - I wasn't sure whether it was necessary since the old one is removed whenever the new one gets generated, so I've just added it to `gitignore` instead. Do you think it's better completely removed? happy to change 👍

apps/examples/wc-next10/ButtonSizes.js
LTurns: @fernandofranca @ashleynolan I've tried using the exports but unfortunately it doesn't work due to how the config is set up. The `createComponent` function is generated in the index.ts under the component instead of in a separate file - I tried generating it in a separate file called `react.ts` but I couldn't seem to get this to work and compile. Placing it under the component means it works easily, but we have to export it as a different name. I've changed it to `Pie(component)React` so it's clear that this is specific for react components. 

Happy to try and get the exports working if anyone wants to pair? 🙏  If not, maybe we can keep it as the React name for now and investigate in a separate ticket 😄

packages/tools/pie-wrappers-react/scripts/react-remove-wrapper.js
LTurns: ^ update on this, I tried to remove the generated custom-elements.json but it doesn't actually remove and replace the file, just repopulates the content inside - my mistake! For this reason though, I think it needs to simply be put inside .gitignore and remain constant 😄

LTurns: turns out I need to push up the custom-elements.json for the CI build to pass since the components repopulate it in the build process 🤦 so I'm removed it for gitignore too!

apps/examples/wc-next10/ButtonSizes.js
```diff
@@ -1,13 +1,5 @@
...
-import { createComponent } from '@lit-labs/react';
-import { PieButton, BUTTON_SIZE } from '@justeattakeaway/pie-button';
-
-const PieBtn = createComponent({
-    tagName: 'pie-button',
-    elementClass: PieButton,
-    react: React,
-    events: { onCustomEvent: 'CustomEvent' },
-});
+import { PieButtonReact, BUTTON_SIZE } from '@justeattakeaway/pie-button';
```
raoufswe, line 2: @justeattakeaway/ui-reviewers-design-system, not sure if we've discussed this yet did we consider the following format for the imports?

`import { PieButton } from '@justeattakeaway/react/pie-button';`

I believe the above format is more semantic to me and will allow us to divide and compose wrappers in a more structured way. for example, we can have a wrapper for:
`import { PieButton } from '@justeattakeaway/vue/pie-button';`
`import { PieButton } from '@justeattakeaway/angular/pie-button';`

instead of postfixing the framework name 👀 

Alternatively, we can ship another package that wraps our components and automates the generation like the current approach of [Adobe](https://opensource.adobe.com/spectrum-web-components/using-swc-react/).

LTurns: I do like the first approach, won't that require a lot of restructuring? I like the idea of something like this -

```
import { PieButton } from '@justeattakeaway/pie-button/react';
import { PieButton } from '@justeattakeaway/pie-button';
```

But I tried this and struggled to get it to work and compile - if you have time to pair, happy to give it another go! We sort of discussed this here through using exports - https://github.com/justeattakeaway/pie/pull/426#discussion_r1188627708

I personally don't think the use of another package is optimal just because it's another separate thing to update and publish. I'm not sure it's necessary if we can do it within the component - but maybe I'm just oversimplifying it 😄

ashleynolan: I think the format @LTurns mentioned would be my preferred import syntax (and that fits directly with how NPM lets you extend packages). It might be good to try and pair someone (likely with React experience I'd guess) to see if it's possible to get that syntax working or if it's just not possible.

I think if we can do it in one component that feels a bit cleaner (to me anyway) – but it could be that we wrap all of our components into a React bundle potentially that includes the wrappers.

Interested to hear others thoughts on this though too – /cc @fernandofranca @xander-marjoram @jamieomaguire

packages/components/pie-button/src/index.ts
raoufswe: Question: why we had to change this?

raoufswe:
Awesome stuff here 🔥

could you please organise a knowledge sharing session about this? I think it's very useful to get all of us onboarded about the logic 🧡

apps/examples/wc-next10/ButtonSizes.js
LTurns: managed to get this working with the import thanks to @siggerzz for the extra help! It currently works as:

```
import { PieButton } from '@justeattakeaway/pie-button/dist/react';
import { PieButton } from '@justeattakeaway/pie-button';`
```
We reference it through the dist, but much better than changing the name 😄

packages/tools/pie-wrapper-react/scripts/react-gen-wrapper.js
```diff
@@ -0,0 +1,87 @@
...
+        return events;
+    }
+
+    // create wrapper src code and add to index.ts file
+    components.forEach(component => {
+        const events = getEvents(component.class)
+
+        componentSrc = `// @ts-nocheck 
+    import * as React from 'react';
+    import { ${component.class.name} as ${component.class.name + 'React'}} from './index';
```
raoufswe, line 50: do we still need to add + 'React' after the last update as well?

packages/tools/pie-wrapper-react/scripts/react-remove-wrapper.js
```diff
@@ -0,0 +1,33 @@
+import { unlink, existsSync } from 'fs';
+
+// removes react wrapper from index.ts after dist has built
+export default function removeReactWrapper (customElementsObject) {
+    let components = []
+    let removedWrapper = ''
```
raoufswe, line 6: `removedWrapper` hasn't been reassigned after it was declared? is this intended?

packages/tools/pie-wrapper-react/__tests__/mocks/mock-component.ts
```diff
@@ -0,0 +1,15 @@
+// @ts-nocheck 
+    import * as React from 'react';
```
fernandofranca, line 2: This file indentation needs some adjustment

packages/tools/pie-wrapper-react/__tests__/mocks/mock-component.ts
```diff
@@ -0,0 +1,15 @@
...
+    import { MockComponent as MockComponentReact} from './index';
```
fernandofranca, line 3: I can't find the `./index` file or the `MockComponent`

packages/tools/pie-wrapper-react/__tests__/wrapper.test.js
```diff
@@ -0,0 +1,23 @@
...
+      const wrapper = genReactWrapper(loadJSON(`/mocks/mock-custom-elements.json`))
+
+      expect(wrapper).toMatchSnapshot();
+    });
+
+    it('should remove generated wrapper from mock component', () => {
+
+      const wrapper = removeReactWrapper(loadJSON(`/mocks/mock-custom-elements.json`))
+
+      expect(wrapper).toMatchSnapshot();
```
fernandofranca, line 19: Snapshot tests are easy to add, at the same time they can be deceiving.

If you look at the snapshot file there isn't much there: `__tests__/__snapshots__/wrapper.test.js.snap`

Another reason I advise against using snapshot tests instead of regular unit tests is because they don't bring the same advantage: one can use the unit test to understand how an API works without having to read each line of its implementation.

packages/tools/pie-wrapper-react/react-gen-wrapper-config.js
```diff
@@ -0,0 +1,16 @@
...
+import path from 'path';
+import fs from 'fs-extra';
+
+const componentPath = path.resolve(process.cwd(), './custom-elements.json')
+
+// fetches custom-elements.json file
+const loadJSON = (path) => JSON.parse(fs.readFileSync(path));
+
+export default {
+    plugins: [
```
fernandofranca, line 11: The same comment goes for `react-remove-wrapper-config`: why is it being exported as an object? Are these files being used anywhere else?

packages/tools/pie-wrapper-react/scripts/react-gen-wrapper.js
```diff
@@ -0,0 +1,87 @@
+import { createWriteStream } from 'fs'
+let componentSrc;
+
+export default function genReactWrapper(customElementsObject){
```
fernandofranca, line 4: It'd be nice to have the JSDoc description for this function

packages/tools/pie-wrapper-react/react-remove-wrapper-config.js
```diff
@@ -0,0 +1,16 @@
...
+import path from 'path';
+
+const componentPath = path.resolve(process.cwd(), './custom-elements.json')
+
+// fetches custom-elements.json file
+const loadJSON = (path) => JSON.parse(fs.readFileSync(path));
+
+export default {
+    plugins: [
+        removeReactWrapper(
```
fernandofranca, line 12: Since these files can actually perform changes, I wouldn't name them `*-config`. Usually config related files only provide settings, not much else.

fernandofranca (commented):
I'd suggest some changes in how scripts are named:

## pie-wrapper-react

### npm scripts

`build:wrapper -> wrapper-react:add`

`remove:wrapper -> wrapper-react:remove`

### script files

`react-gen-wrapper.js -> add-react-wrapper`

`react-remove-wrapper.js -> remove-react-wrapper`

apps/examples/wc-react18/src/App.js
```diff
@@ -1,33 +1,22 @@
...
-const PieBtn = createComponent({
-    tagName: 'pie-button',
-    elementClass: PieButton,
-    react: React,
-    events: { onCustomEvent: 'CustomEvent' },
-});
+import './styles.css';
 
 function App () {
     const [counter, setCounter] = useState(0);
```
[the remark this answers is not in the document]
LTurns: I believe there was a PR recently to remove customEvent from `pie-button`, since it could actually just be handled directly inside the consuming apps - https://github.com/justeattakeaway/pie/blob/main/packages/components/pie-button/README.md?plain=1#L68

I tested it by adding the event locally to `onClick` like: `<pie-button onclick="e => console.log(e)">Click me!</pie-button>` and it all works 👍

packages/tools/pie-wrapper-react/__tests__/mocks/mock-component.ts
LTurns: the test actually just creates the template that a component would get - and references it's index. Since the test isn't a component, it doesn't actually directly link to anywhere 😄 I've actually removed this file though - it's auto generated through the test but isn't needed and probably ends up being more confusing than helpful

packages/tools/pie-wrapper-react/__tests__/wrapper.test.js
LTurns: So I've added some unit tests as well as the snapshot tests - I think both could be useful but agree more was probably needed to test the logic here 👍

packages/tools/pie-wrapper-react/react-remove-wrapper-config.js
LTurns: ah thanks! I've actually been able to refactor the code to remove this and the `add-wrapper-config` entirely 👍 the package.json script now heads directly to the react-wrapper scripts themselves. I originally got the config approach from the adobe example but I don't think it's needed in our case 😄

packages/tools/pie-wrapper-react/react-gen-wrapper-config.js
LTurns: ended up removing this file 👍 see comment below

packages/tools/pie-wrapper-react/scripts/add-react-wrapper.js
```diff
@@ -0,0 +1,113 @@
...
+    const customElements = Object.entries(customElementsObject);
+
+    // sort through customElements array and put all components into a separate array
+    customElements.forEach(([key, value]) => {
+        let componentObject;
+        if (key === 'modules') {
+            value.forEach((k) => {
+                if (k.path.includes(folderName)) {
+                    k.declarations.forEach((decl) => {
+                        componentObject = decl.customElement === true ? components.push({ class: decl, path: k.path.replace('index.js', 'react.ts') }) : '';
```
fernandofranca, line 31: Looks like `componentObject` is not needed and this could become a simple if statement.

`if(decl.customElement === true) components.push({ class: decl, path: k.path.replace('index.js', 'react.ts') });`

packages/tools/pie-wrapper-react/scripts/add-react-wrapper.js
```diff
@@ -0,0 +1,113 @@
...
+    function getEvents (component) {
+        const events = [];
+        if (component?.events) {
+            events.push(component.events
+                    .filter((event) => !!event.name)
+                    .map((event) => ({
+                        name: event.name,
+                        type: event.type?.text || 'Event',
+                        description: event.description,
+                    })));
```
fernandofranca, line 55: I believe we can make it more readable by unnesting:

```
    if (component?.events) {
        const events = component.events
            .filter((event) => !!event.name)
            .map((event) => ({
                name: event.name,
                type: event.type?.text || 'Event',
                description: event.description,
            }));
            
        events.push(events);
    }
```

packages/tools/pie-wrapper-react/scripts/remove-react-wrapper.js
```diff
@@ -0,0 +1,49 @@
...
+    const components = [];
+    const customElements = Object.entries(customElementsObject);
+
+    // sort through customElements array and put all components into a separate array
+    customElements.forEach(([key, value]) => {
+        let componentObject;
+        if (key === 'modules') {
+            value.forEach((k) => {
+                k.declarations.forEach((decl) => {
+                    componentObject = decl.customElement === true ? components.push({ class: decl, path: k.path.replace('index.js', 'react.ts') }) : '';
```
fernandofranca, line 28: As in `add-react-wrapper.js`, it can be an `if` statement and `componentObject` can be removed.

packages/components/pie-button/src/index.ts
LTurns: It needs to be changed for the @custom-elements-manifest/analyzer package which doesn't recognise the shorthand unfortunately - https://custom-elements-manifest.open-wc.org/analyzer/getting-started/ but both mean the same thing 👍

packages/tools/pie-wrapper-react/scripts/add-react-wrapper.js
LTurns: Ah I'm not sure I can do this actually because I need to return the events array at the bottom, so the const needs to be declared outside the if statement, otherwise will receive undefined? let me know if there's another way round that though 👍

fernandofranca (approved):

FayeCarter:
Hey, will you need to get rid of the bits in the readme for the other react example apps now?
https://github.com/justeattakeaway/pie/blob/main/apps/examples/wc-react18/README.md#testing-web-components-in-react-18

LTurns:
> Hey, will you need to get rid of the bits in the readme for the other react example apps now? https://github.com/justeattakeaway/pie/blob/main/apps/examples/wc-react18/README.md#testing-web-components-in-react-18

ah yep, thanks @FayeCarter - will get rid of those 👍

FayeCarter (approved):

fernandofranca (approved):

packages/tools/pie-wrapper-react/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "testPathIgnorePatterns": [
+      "__tests__/mocks/"
+    ]
+  },
+  "devDependencies": {
+    "@custom-elements-manifest/analyzer": "0.8.0",
+    "@lit-labs/react": "1.1.1",
+    "cem-plugin-module-file-extensions": "0.0.5"
+  },
+  "dependencies": {
```
siggerzz, line 31: All of these should be devDeps I think.

Can you try removing `jest` from here, and updating the root to 29.5.0 - https://github.com/justeattakeaway/pie/blob/main/package.json#L83

Ideally we want a single `jest` version for the entire monorepo.

@babel/preset-env can also be removed, as we define it at the root - https://github.com/justeattakeaway/pie/blob/main/package.json#L55

siggerzz (dismissed):
Minor comment around deps!

apps/examples/wc-next10/README.md
```diff
@@ -39,7 +39,7 @@ module.exports = withTM(nextConfig);
 
 In order to use the Pie Button components in the Next 10 application, it needs to be dynamically imported. Without this, you are likely to receive the `window is not defined` error.
 ```
-  const PButton = dynamic(() => import('../components/PieButton'), { ssr: false });
+  const PButton = dynamic(() => import('../components/PieButton/dist/react'), { ssr: false });
```
ashleynolan, line 42: Should this be…

```suggestion
  const PieButton = dynamic(() => import('../components/PieButton/dist/react'), { ssr: false });
```

packages/tools/pie-wrapper-react/__tests__/wrapper.test.js
```diff
@@ -0,0 +1,61 @@
+import fs from 'fs-extra';
+import path from 'path';
+import { addReactWrapper } from '../scripts/add-react-wrapper';
+import removeReactWrapper from '../scripts/remove-react-wrapper';
```
ashleynolan, line 4: Is it possible to make these imports consistent? Looks like you're using a default export for `removeReactWrapper` and a named export for `addReactWrapper`. I don't have a preference either way, but I'd probably make sure that they are importable in the same way for consistency.

.changeset/calm-beers-compare.md
```diff
@@ -0,0 +1,8 @@
+---
+"@justeattakeaway/pie-wrapper-react": minor
+---
+
+[Added] - react wrappers package for wrapping lit components
+[Added] - scripts for adding and removing createComponent wrapper
+[Added] - `custom-elements.json` and `custom-elements-manifest.config.js` for converting custom-elements into JSON
```
ashleynolan, line 7:
```suggestion
[Added] - React wrappers package for wrapping lit components
[Added] - Scripts for adding and removing createComponent wrapper
[Added] - `custom-elements.json` and `custom-elements-manifest.config.js` for converting custom-elements into JSON
```

.changeset/loud-olives-relate.md
```diff
@@ -0,0 +1,5 @@
+---
+"pie-monorepo": minor
+---
+
+[Removed] - react `createComponent` wrappers from react/next apps
```
ashleynolan, line 5:
```suggestion
[Removed] - React `createComponent` wrappers from react/next apps
```

.changeset/moody-apricots-behave.md
```diff
@@ -0,0 +1,5 @@
+---
+"@justeattakeaway/pie-button": minor
+---
+
+[Changed] - build script to include generating and removing react wrapper
```
ashleynolan, line 5:
```suggestion
[Changed] - Build script to include generating and removing react wrapper
```

apps/examples/wc-react18/README.md
```diff
@@ -11,32 +11,8 @@ Open [http://localhost:3000](http://localhost:3000) to view it in your browser.
 Builds the app for production to the `build` folder.\
 
 
-## Testing Web Components in React 18
+#### Using Web Components in React
 
-React 18 and previous versions don't handle web components and custom elements out of the box correctly in all cases due to how React treats custom props and events (More details can be found [here](https://lit.dev/docs/frameworks/react/)). While the React team fixed some of the issues with this [experimental release](https://github.com/justeattakeaway/pie/compare/0.0.0-experimental-56a3c18e5-20230314?expand=1), we shouldn't rely on it as many of our apps are using older instances of React and a few [issues](https://custom-elements-everywhere.com/libraries/react/results/results.html) haven't been addressed yet.
+React 18 and previous versions don't handle web components and custom elements out of the box correctly in all cases due to how React treats custom props and events (More details can be found [here](https://lit.dev/docs/frameworks/react/)). We have created the fix for this by generating a `react` file inside the component's `dist` folder. Therefore, in order to import the web component into your React application, please use:
```
ashleynolan, line 16:
```suggestion
React 18 and previous versions don't handle web components and custom elements out of the box correctly in all cases due to how React treats custom props and events (More details can be found [here](https://lit.dev/docs/frameworks/react/)). Our solution for this is to wrap the web component to include the `[@lit-labs/react](https://lit.dev/docs/frameworks/react/)` package, generating a `react` file inside the component's `dist` folder.

Therefore, in order to import the web component into your React application, please use:
```

packages/components/pie-button/package.json
```diff
@@ -5,9 +5,8 @@
   "type": "module",
   "main": "dist/index.js",
   "module": "dist/index.js",
-  "types": "dist/types/index.d.ts",
   "scripts": {
-    "build": "run -T vite build",
+    "build": "yarn wrapper-react:remove && yarn wrapper-react:add pie-button && run -T vite build && yarn wrapper-react:remove",
```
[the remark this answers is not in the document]
LTurns: I've updated the scripts to `yarn build:wrapper pie-button && run -T vite build && yarn remove:wrapper` - hopefully makes it less confusing for now 👍

ashleynolan (approved):

FayeCarter (dismissed):

packages/tools/pie-wrapper-react/scripts/add-react-wrapper.js
```diff
@@ -0,0 +1,113 @@
...
+export function addReactWrapper (customElementsObject, folderName = process.argv[2]) {
+    const components = [];
+    const customElements = Object.entries(customElementsObject);
+
+    // sort through customElements array and put all components into a separate array
+    customElements.forEach(([key, value]) => {
+        let componentObject;
+        if (key === 'modules') {
+            value.forEach((k) => {
+                if (k.path.includes(folderName)) {
```
webbertakken, line 29: should this be using `.endsWith(folderName)` (removing the filename first)? Or will there never be conflicts when using `includes`?

webbertakken (dismissed):
Great work! Keep it up :)

packages/components/pie-button/package.json
```diff
@@ -5,9 +5,8 @@
   "type": "module",
   "main": "dist/index.js",
   "module": "dist/index.js",
-  "types": "dist/types/index.d.ts",
   "scripts": {
-    "build": "run -T vite build",
+    "build": "yarn build:wrapper pie-button && run -T vite build",
```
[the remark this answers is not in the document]
LTurns: not temporary but could be revisited - I added the `pie-button` as a parameter so only the pie-button wrapper gets built when inside that directory or when built from the root. Otherwise, yeah it does become generic and generates the wrappers for all components - maybe in future we could just do this but it might be confusing running yarn build in `pie-button` and having a wrapper created in `pie-modal` too 😄

packages/tools/pie-wrapper-react/scripts/add-react-wrapper.js
LTurns: I don't think there should ever be conflicts based on how these folders are structured, but it's probably more logical to change to `endsWith` anyway! will change 👍

webbertakken (approved):

webbertakken (dismissed):

packages/components/pie-button/package.json
```diff
@@ -5,6 +5,7 @@
   "type": "module",
   "main": "dist/index.js",
   "module": "dist/index.js",
+  "types": "dist/types/index.d.ts",
```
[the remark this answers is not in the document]
LTurns: agree with this but gonna add in a separate PR 👍

ashleynolan (approved):

jamieomaguire (approved):

PR #447, opened 2023-05-11
feat(pie-docs): add global token usage warning to all global pages

jamieomaguire:
[Added] - a reusable warning about global token usage to all global tokens pages
[Added] - a reusable warning about global token usage to all global tokens pages

### DIFF
--- a/.changeset/eight-plants-relate.md
+++ b/.changeset/eight-plants-relate.md
@@ -1,0 +1,5 @@
+---
+"pie-docs": minor
+---
+
+[Added] - a reusable warning about global token usage to all global tokens pages
--- a/.changeset/lovely-rats-crash.md
+++ b/.changeset/lovely-rats-crash.md
@@ -1,0 +1,5 @@
+---
+"pie-docs": minor
+---
+
+[Added] - missing horizontal dividers to spacing token pages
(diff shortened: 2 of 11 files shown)


jordifornellsJET:
Hi @jamieomaguire, here are my initial comments:

- I wouldn't capitalise 'Alias Tokens' (let's use 'alias tokens')
- Would it be worth linking to the alias tab for the word 'alias tokens'? Or is it a bit of overkill?

Other than that, it looks great 👍

jamieomaguire:
> Hi @jamieomaguire, here are my initial comments:
> 
> * I wouldn't capitalise 'Alias Tokens' (let's use 'alias tokens')
> * Would it be worth linking to the alias tab for the word 'alias tokens'? Or is it a bit of overkill?
> 
> Other than that, it looks great 👍

Happy to update the spacing, but can you also please do so in Figma? I just copied the text out of the design :) 

Personally, I think the linking is more effort than it's worth but up to you.

jordifornellsJET:
> 

Will update it in Figma 👍, and let's leave this as it is and not include a link. Thanks!
✅ **Design review: Done**

FayeCarter (dismissed):

jamieomaguire:
> > 
> 
> Will update it in Figma 👍, and also can we change the Global Tokens to also not be capitalised? It will read: **Don’t use global tokens directly**. Please make sure you’re using alias tokens in your products.
> 
> Let's leave this as it is and not include a link. Thanks!
> 
> ✅ **Design review: Done**

Thanks @jordifornellsJET, as I mentioned offline I actually think your suggestion would be worth doing as it will improve the mobile UX (prevents having to scroll back up to go to alias). I've just pushed an implementation of this. Please re-review if you can ❤️

jordifornellsJET:
> > > 
> > 
> > 
> > Will update it in Figma 👍, and also can we change the Global Tokens to also not be capitalised? It will read: **Don’t use global tokens directly**. Please make sure you’re using alias tokens in your products.
> > Let's leave this as it is and not include a link. Thanks!
> > ✅ **Design review: Done**
> 
> Thanks @jordifornellsJET, as I mentioned offline I actually think your suggestion would be worth doing as it will improve the mobile UX (prevents having to scroll back up to go to alias). I've just pushed an implementation of this. Please re-review if you can ❤️

Amazing! This looks great, if we could just change the capitalisation on Global Tokens to global tokens, that would be the last bit! Sorry about that, thanks! :)

jordifornellsJET (approved):

FayeCarter (approved):

raoufswe (approved):
image

PR #472, opened 2023-05-18
test(pie-icons-vue): [Added] - Jest test support and basic demo tests

xander-marjoram:
The ticket wasn't around adding a large number of tests, but rather setting up the package so that we can successfully run Jest tests.

I've got it working with a standard unit test and also one that mounts an icon component using `@vue/test-utils`.

### Reviewer 1

### Reviewer 2

### DIFF
--- a/packages/tools/pie-icons-vue/helpers/test/index.test.js
+++ b/packages/tools/pie-icons-vue/helpers/test/index.test.js
@@ -1,0 +1,21 @@
+const { removeHyphenBeforeDigits } = require('../index');
+
+describe('removeHyphenBeforeDigits', () => {
+    it.each([
+        ['component-123', 'component123'],
+        ['test-', 'test-'],
+        ['test123', 'test123'],
+        ['-123', '123'],
+        ['7', '7'],
+        ['', ''],
+        ['component-name', 'component-name'],
+        ['component-123-extra', 'component123-extra'],
+        ['component-1-2-3', 'component1-2-3']
+    ])('should remove a hyphen before digits - case = %s', (testString, expected) => {
+        // Act
+        const result = removeHyphenBeforeDigits(testString);
+
+        // Assert
+        expect(result).toBe(expected);
+    });
+});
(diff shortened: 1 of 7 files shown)


ashleynolan (approved):

raoufswe (dismissed):

fernandofranca (changes requested):
The unit test doesn't prove we can now mount and test Vue components. If it's the case, can you add an example?

Please give the ticket (DSW-672) another read or let's pair on this one if it's unclear.

>Test suite files should be able to mount pie-icons-vue components and check its properties

What is the plan for testing Vue icons? To use the CJS build or the ones from `generated/components`?

packages/tools/pie-icons-vue/package.json
```diff
@@ -39,7 +39,8 @@
     "lint:scripts": "run -T eslint .",
     "lint:scripts:fix": "yarn lint:scripts --fix",
     "prebuild:example": "yarn build:icons",
-    "prepublishOnly": "yarn lint:scripts && yarn build"
+    "prepublishOnly": "yarn lint:scripts && yarn build",
+    "test": "run -T jest"
```
webbertakken, line 43: Side note: Any reason you're not using the much more modern Vitest (which has the same programmatic API as Jest)

fernandofranca: I think it's a good idea since it will allow us to import components from ES modules out of the box.

webbertakken: Exactly :) And it's generally more modern and shows nicer output. Easier to integrate with  coverage tools as well (Core team uses C8 which is default from the docs I think).

fernandofranca (dismissed):
Component tests! Yay!!!

packages/tools/pie-icons-vue/helpers/test/index.test.js
```diff
@@ -0,0 +1,21 @@
+import { removeHyphenBeforeDigits } from '../index';
+
+describe('removeHyphenBeforeDigits', () => {
+    test.each([
```
kevinrodrigues, line 4: More of a suggestion, It would be nice to split this out so you have just the tests that remove all cases for hyphens and then another test that shows where it won't apply the method i.e `test123` will always be `test123` That way the second test description can read `should NOT remove a hyphen before digits/cases`. At the moment when the test description is read out the test looks like it `should` apply the logic to the cases where it shouldn't remove it if that makes sense.

kevinrodrigues (dismissed):

kevinrodrigues (approved):

ashleynolan (approved):

PR #478, opened 2023-05-19
feat(pie-webc-core): add core package with RTL support

jamieomaguire:
Adds a new private package, `@justeattakeaway/pie-web-components-core` which will house any reusable base classes, mixins, utilities and such that we need to share between web components.

Adds a basic RTL support mixin. It provides two behaviours:

1. A reactive, explicitly set `dir` attribute which behaves the same as the native [dir](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir) attribute.

2. A static, non-reactive property that detects the nearest `dir` set using `window. getComputedStyle`. This is probably what will be used for 90% of use cases as most products that have the ability to switch between LTR and RTL languages involve a page refresh. We can revisit this and look to making this fallback reactive in the future if there is demand but for now it seems overkill

The mixin can be used either alone or along side other mixins to compose custom base classes:

```
class SomeComponent extends RTLMixin(LitElement) {}
class SomeComponent extends AnotherMixin(RTLMixin(LitElement)) {}
```

In the future we may want to select a number of 'base' mixins to combine and create a web component base class which extends the desired mixins (perhaps RTL, something style related, etc...).

### Why do we need the RTL Mixin?
Whilst we can likely support 90% of RTL designs using CSS only, there will be scenarios where we need to change what markup we render based on the writing direction. For example, icons cannot simply be mirrored. Some RTL icons are entirely different images from their LTR equivalents. In these cases, we need a way to detect the writing direction in Javascript.

This mixin gives us an easy to use flag to hook into when creating our render logic. For example:

```
render() {
        // isRTL is defined on the base class created by the RTLMixin
        const {
            size, type, variant, disabled, isFullWidth, isRTL,
        } = this;

        return html`
            <button
                class=${classMap(classes)}
                type=${type}
                ?disabled=${disabled}
                ?isFullWidth=${isFullWidth}>
                ${isRTL ? html`<RTLIcon />` : html`<LTRIcon />`} // <-- use the isRTL flag to change markup rendered
                <slot></slot>
            </button>`;
}
```

### DIFF
--- a/.changeset/ten-lamps-drive.md
+++ b/.changeset/ten-lamps-drive.md
@@ -1,6 +1,0 @@
----
-"@justeattakeaway/pie-button": minor
-"pie-storybook": minor
----
-
-[Changed] - use RTL Mixin from the new core library


.changeset/smart-buses-knock.md
```diff
@@ -0,0 +1,5 @@
+---
+"@justeattakeaway/pie-web-components-core": patch
```
siggerzz, line 2: Should this be `minor`?

packages/components/pie-button/src/index.ts
```diff
@@ -10,7 +11,7 @@ import { BUTTON_SIZE, BUTTON_TYPE, BUTTON_VARIANT } from './defs';
 export { BUTTON_SIZE, BUTTON_TYPE, BUTTON_VARIANT };
 
 @customElement('pie-button')
-export class PieButton extends LitElement {
+export class PieButton extends RtlMixin(LitElement) {
```
ashleynolan, line 14: More a general q – should we expect that all components will wrap this, or only when we actually want to use it?

Mainly the q is – should we remove the mixin from pie-button as we don't need it (now that you've shown how it's used)

packages/components/pie-web-components-core/README.md
```diff
@@ -0,0 +1,7 @@
+<p align="center">
+  <img align="center" src="../../../readme_image.png" height="200" alt="">
+</p>
+
+# pie-web-components-core
```
ashleynolan, line 5: To keep the naming with our docs site (and because I'm lazy) – would this be potentially better named `pie-webc-core`

jamieomaguire: sounds good!

ashleynolan (dismissed):

.changeset/smart-buses-knock.md
raoufswe: https://www.youtube.com/watch?v=hsornYQ7Uvs

packages/components/pie-webc-core/src/mixins/rtl/rtlMixin.ts
```diff
@@ -0,0 +1,41 @@
...
+            get isRTL () : boolean {
+                if (this.dir === 'ltr') {
+                    return false;
+                }
+
+                if (this.dir === 'rtl' || window?.getComputedStyle(this)?.direction === 'rtl') {
+                    return true;
+                }
+
+                return false;
```
raoufswe, line 36: I believe `auto` is also a valid value for the [dir](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir) attribute. there are some cases where a component might mix between ltr and rtl and based on this approach, we would assume it `isRTL` is false. 👀

packages/components/pie-modal/src/index.ts
```diff
@@ -1,8 +1,9 @@
 import { LitElement, html } from 'lit'; // eslint-disable-line import/no-extraneous-dependencies
 import { customElement } from 'lit/decorators.js'; // eslint-disable-line import/no-extraneous-dependencies
+import { RtlMixin } from '@justeattakeaway/pie-webc-core';
 
 @customElement('pie-modal')
-export class PieModal extends LitElement {
+export class PieModal extends RtlMixin(LitElement) {
```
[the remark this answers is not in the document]
jamieomaguire: @raoufswe from the description above:

"The mixin can be used either alone or along side other mixins to compose custom base classes:

class SomeComponent extends RTLMixin(LitElement) {}
class SomeComponent extends AnotherMixin(RTLMixin(LitElement)) {}
In the future we may want to select a number of 'base' mixins to combine and create a web component base class which extends the desired mixins (perhaps RTL, something style related, etc...)."

:)

packages/components/pie-webc-core/src/mixins/rtl/rtlMixin.ts
jamieomaguire: yes, we should consider it in the future but I'm pretty happy with not worrying about `auto` for now :)

raoufswe (approved):

PR #484, opened 2023-05-24
test(pie-docs): add system test for navigation on desktop

FayeCarter:

[Added] - Percy test for sticky nav on desktop
[Added] - `isElementVisibleInViewport` test helper function


apps/pie-docs/test/visual/navigation.spec.js
```diff
@@ -35,7 +35,7 @@ test.describe('PIE - Site Nav Menu', () => {
         await navToggleLabel.click();
         await navMenu.isVisible();
 
-        const mobileWidths = [PERCY_BREAKPOINTS.MOBILE, PERCY_BREAKPOINTS.TABLET];
+        const mobileWidths = [PERCY_BREAKPOINTS.MOBILE, PERCY_BREAKPOINTS.TABLET, PERCY_BREAKPOINTS.DESKTOP];
```
kevinrodrigues, line 38: Probably should rename this const as it contains multiple widths which aren't specific to mobile now.

apps/pie-docs/test/visual/navigation.spec.js
```diff
@@ -26,7 +26,7 @@ test.describe('PIE - Site Nav Menu', () => {
         await disableCookieBanner(page, context, true);
     });
 
-    test('Should open and close the mobile navigation menu - @mobile', async ({ page }) => {
+    test('Should open and close the navigation menu', async ({ page }) => {
```
kevinrodrigues, line 29: Sorry, I read this test again so wanted to check, doesn't the mobile nav only open on small screens (mobile and tablet) anyway? So adding desktop probably doesn't really test anything visually unless i am missing something here?

apps/pie-docs/test/system/navigation.spec.js
```diff
@@ -45,3 +46,28 @@ test.describe('PIE - site nav menu - @mobile', () => {
         await expect(navMenu).not.toBeVisible();
     });
 });
+
+test.describe('PIE - site nav menu - @desktop', () => {
+    test.beforeEach(async ({ page, context }) => {
+        await disableCookieBanner(page, context);
+    });
+
+    test('Should still be visible when the page has scrolled down', async ({ page }) => {
```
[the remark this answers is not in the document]
FayeCarter: Hey, thanks, I'll split these up.

On what the tests cover, I chatted with Josh and we realised that the tests the ticket ask for don't actually cover any new functionality.

The Percy tests already cover that the navigation is visible as the they show the navigation on each page in desktop mode already. We also already have the mobile tests that check that it's present on mobile on click.

For desktop, there isn't anything to click that would open or close the navigation. The only functionality that needs checking is that the navigation remains at the top of the page when the user scrolls.

raoufswe (approved):

apps/pie-docs/test/helpers/functions.js
```diff
@@ -0,0 +1,8 @@
+const isElementVisibleInViewport = (element) => element.evaluate((el) => {
+    const rect = el.getBoundingClientRect();
+    return rect.top >= 0 && rect.bottom <= window.innerHeight;
+});
+
+export {
+    isElementVisibleInViewport,
+};
```
jamieomaguire, line 8: pretty sure Playwright provides this kind of utility. Have you seen anything similar on their docs?

jamieomaguire: such as these: https://playwright.dev/docs/api/class-elementhandle#element-handle-is-visible

FayeCarter: I did yeah, so that checks if it's in the DOM, not if it's in the viewport. So the test will fail because it thinks that the header is still on the page.

apps/pie-docs/test/system/navigation.spec.js
```diff
@@ -45,3 +46,33 @@ test.describe('PIE - site nav menu - @mobile', () => {
...
+
+        // Arrange
+        siteNavigation = await page.getByTestId('site_nav_section_1');
+        contentHeader = await page.getByTestId('content_header');
+    });
+
+    test('Should be visible at the top of the page', async () => {
+        // Assert - Navigation and top of the page are visible
+        await expect.soft(await isElementVisibleInViewport(siteNavigation)).toBeTruthy();
+        await expect.soft(await isElementVisibleInViewport(contentHeader)).toBeTruthy();
```
JoshuaNg2332, line 66: Each test should end with a normal except. So you can remove the `.soft` from this line so that the test ends.

apps/pie-docs/test/system/navigation.spec.js
```diff
@@ -45,3 +46,33 @@ test.describe('PIE - site nav menu - @mobile', () => {
...
+    });
+
+    test('Should still be visible when the page has scrolled down', async ({ page }) => {
+        // Act - Scroll to the bottom of the page
+        await page.keyboard.press('End');
+        await page.waitForTimeout(1000);
+
+        // Assert - Navigation is still visible, but top of the page is not
+        await expect(await isElementVisibleInViewport(siteNavigation)).toBeTruthy();
```
JoshuaNg2332, line 75: You can swap the positions of line 75 and 76 so that the siteNav expect is last and also is a hard expect. If we have a hard expect before a `.soft` expect, the test technically ends on this line without ever testing if the `contentHeader` is not visible.

apps/pie-docs/test/system/navigation.spec.js
```diff
@@ -45,3 +46,33 @@ test.describe('PIE - site nav menu - @mobile', () => {
...
+        await page.goto('all-about-pie/what-is-pie/');
+
+        // Arrange
+        siteNavigation = await page.getByTestId('site_nav_section_1');
+        contentHeader = await page.getByTestId('content_header');
+    });
+
+    test('Should be visible at the top of the page', async () => {
+        // Assert - Navigation and top of the page are visible
+        await expect(await isElementVisibleInViewport(siteNavigation)).toBeTruthy();
```
JoshuaNg2332, line 65: make this line a `.soft` or it will end the test at this point as opposed to the line below.

FayeCarter: Thanks!

JoshuaNg2332 (approved):

fernandofranca (approved):

ashleynolan (approved):

PR #487, opened 2023-05-25
feat(pie-monorepo): extract and share web components configuration files

fernandofranca:
### pie-monorepo

[Added] - pie-components-config package

### @justeattakeaway/pie-components-config

[Added] - common components configurations

### @justeattakeaway/pie-icon-button, pie-button & pie-modal

[Updated] - components to use the shared configurations

## Context

The following build, development and testing related settings files were unified and shared between `pie-icon-button`, `pie-button`, and `pie-modal`:

- `vite.config.js`
- `tsconfig.json`
- `playwright-lit-config.js`
- `playwright-lit-visual-config.js`

These configuration files are now located at `configs/pie-components-config`

From now on we should ideally concentrate components configurations (build, testing, etc...) in a single place to avoid synchronization issues and further maintenance.

I also took the liberty to move the `cookies.js` file under `configs/constants`

### Reviewer 1

### Reviewer 2


configs/pie-components-config/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "@justeattakeaway/pie-components-config",
+  "version": "0.1.0",
+  "private": true,
```
jamieomaguire, line 4: I realise the `pie-webc-core` package is also private. Is there any reason we might want to make them both public?

packages/components/pie-button/tsconfig.json
```diff
@@ -1,30 +1,8 @@
...
-      "noImplicitAny": true,
-      "noImplicitThis": true,
-      "moduleResolution": "node",
-      "allowSyntheticDefaultImports": true,
-      "experimentalDecorators": true,
-      "forceConsistentCasingInFileNames": true
+    "baseUrl": ".",
+    "rootDir": ".",
   },
   "include": ["src/**/*.ts","./declaration.d.ts", "test/**/*.ts"],
```
jamieomaguire, line 7: Why don't we want the includes in the shared config?

packages/components/pie-button/tsconfig.json
```diff
@@ -1,30 +1,8 @@
...
-      "noUnusedParameters": true,
-      "noImplicitReturns": true,
-      "noFallthroughCasesInSwitch": true,
-      "noImplicitAny": true,
-      "noImplicitThis": true,
-      "moduleResolution": "node",
-      "allowSyntheticDefaultImports": true,
-      "experimentalDecorators": true,
-      "forceConsistentCasingInFileNames": true
+    "baseUrl": ".",
```
jamieomaguire, line 4: Does TS make us declare these again? I see we already declare in the shared file

jamieomaguire (dismissed):

packages/components/pie-button/tsconfig.json
fernandofranca: The initial attempt was to bring everything into the shared tsconfig, though during testing I found some settings weren't working properly because TS infers some paths as being relative to the config folder not the actual one... 🤷‍♂️

packages/components/pie-button/tsconfig.json
fernandofranca: Yes. Otherwise, it won't point to the component folder.

configs/pie-components-config/package.json
fernandofranca: Good question. 

I'd say, at least for now, `pie-components-config` is more like internal, dev-oriented content being shared as a workspace, than a package consumers might have interest in.

Do you have any opinions on that? I might be missing something...

configs/pie-components-config/placeholder.ts
```diff
@@ -0,0 +1 @@
+// This file exists solely to suppress the "No inputs were found in configuration file." TS warning.
```
raoufswe, line 1: nit: do you think we can have another workaround that doesn't require creating an empty file? 

I assume we will eventually have a ts file somewhere in the directory to get rid of the issue, but maybe we can change the file extension of playwright files to ts? I don't think it will break anything but worth trying :D

raoufswe (approved):

configs/pie-components-config/placeholder.ts
fernandofranca: I was experimenting with your idea and came to the conclusion that it might not be as convenient as thought.

If we rename the playwright files to ts, the package index file also should be ts. With that, we lose the possibility of consuming `pie-components-config` files from vanilla js workspaces.

jamieomaguire (approved):

raoufswe (approved):

configs/pie-components-config/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "@justeattakeaway/pie-components-config",
+  "version": "0.1.0",
```
ashleynolan, line 3: I think this will need to be `v0.0.0`, otherwise changesets will try and bump the package to `v0.2.0` when you merge the PR

configs/pie-components-config/placeholder.ts
fernandofranca: The issue happens in the consuming side, if it's not a TS project it will reject the `index` file.

`TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".ts"`

package.json
```diff
@@ -13,6 +13,7 @@
   "workspaces": {
     "packages": [
       "apps/**/*",
+      "configs/**/*",
```
ashleynolan, line 16: Would it make sense to have this package at `/packages/configs` (similar to how we organise `packages/tools`)?

configs/pie-components-config/package.json
fernandofranca: Indeed!

package.json
fernandofranca: With the exception of `pie-icons-configs`, everything in `packages/*` is useful for customers and I believe we should have a way of segregating customer oriented packages that are published from internal/tooling/dev oriented code that doesn't need to be published.

ashleynolan (approved):

jamieomaguire (approved):

raoufswe (approved):

PR #491, opened 2023-05-25
feat(pie-icon-button): add variant and disabled props and styles

jamieomaguire:
- Add the variant styles for `pie-icon-button`
- Hardcode an inline SVG temporarily to demonstrate the styles
- Move prop validation logic into shared core package (NOTE: we will will likely move from Enums to string literal union types in the future but for now this keeps the code base smaller)


packages/components/pie-icon-button/src/iconButton.scss
```diff
@@ -0,0 +1,122 @@
...
+    background-color: var(--btn-bg-color);
+    cursor: pointer;
+    user-select: none;
+
+    display: flex;
+    align-items: center;
+    justify-content: center;
+
+    &:focus-visible {
+        // TODO - Need to update to use the latest focus styling
```
sebastienrocks, line 46: Forgotten TODO? Or is this for later?

jamieomaguire: I think we need to update the design tokens first which we could do separately (needs to be done for the pie-button package too)

packages/components/pie-icon-button/src/index.ts
```diff
@@ -1,20 +1,45 @@
-import { LitElement, html } from 'lit'; // eslint-disable-line import/no-extraneous-dependencies
-import { customElement } from 'lit/decorators.js'; // eslint-disable-line import/no-extraneous-dependencies
+import { LitElement, html, unsafeCSS } from 'lit'; // eslint-disable-line import/no-extraneous-dependencies
```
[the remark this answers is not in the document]
jamieomaguire: General test strategy is still being agreed upon however we will be adding visual Percy tests near the end of this epic

sebastienrocks (dismissed):

packages/components/pie-icon-button/src/index.ts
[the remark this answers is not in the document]
ashleynolan: Not sure what our flow should be for visual regression tests, but if they aren't part of this PR, can you create a ticket to add the visual regression tests for these for `icon-button`? (so we can get that pulled in separately)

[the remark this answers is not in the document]
jamieomaguire: I can raise visual tests in a separate PR. I didn't believe they were a part of this ticket but happy to add them separately 👍

apps/pie-storybook/stories/pie-icon-button.stories.ts
```diff
@@ -19,22 +27,49 @@ export default {
...
 
 interface IconButtonProps {
-    slot: TemplateResult;
+    variant: ICON_BUTTON_VARIANT;
+    disabled: boolean;
 }
 
 const Template = ({
-    slot,
+    variant, disabled,
```
kevinrodrigues, line 35: Minor, but could you line break this?

packages/components/pie-icon-button/src/iconButton.scss
```diff
@@ -0,0 +1,122 @@
...
+        outline: 2px solid var(--btn-focus);
+    }
+
+    svg {
+        height: 24.5px;
```
kevinrodrigues, line 51: Are these values intended?

packages/components/pie-icon-button/src/index.ts
```diff
@@ -1,20 +1,45 @@
...
+const componentSelector = 'pie-icon-button';
+
+@customElement(componentSelector)
 export class PieIconButton extends LitElement {
-    // eslint-disable-next-line class-methods-use-this
+    @property()
+    @validPropertyValues(componentSelector, Object.values(ICON_BUTTON_VARIANT), ICON_BUTTON_VARIANT.PRIMARY)
+        variant : ICON_BUTTON_VARIANT = ICON_BUTTON_VARIANT.PRIMARY;
+
+    @property({ type: Boolean, reflect: true })
```
kevinrodrigues, line 18: Did we ever ask the LIT team about the reflect property and sb out of interest?

jamieomaguire: Didn't ask them, but my understanding is that it means if the prop changes internally then the attribute in the DOM will update to reflect that value. It's probably not needed here.

jamieomaguire: "You can configure a property so that whenever it changes, its value is reflected to its [corresponding attribute](https://lit.dev/docs/components/properties/#observed-attributes)."

https://lit.dev/docs/components/properties/#reflected-attributes

jamieomaguire: removed as we don't need it :)

jamieomaguire:
I've agreed with Ash to add visual tests in a follow up PR straight after

sebastienrocks (approved):

kevinrodrigues (approved):

packages/components/pie-icon-button/src/defs.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Icon Button style variants
+ */
+export enum ICON_BUTTON_VARIANT {
```
ashleynolan, line 4: Should we be using enums here? (after the comments from Skip)

packages/components/pie-icon-button/src/iconButton.scss
ashleynolan: We likely need to flag this with @RLumleyJET, as that size doesn't match our own definitions. Not sure if it's intended (I imagine it should just be `24px`?)

PR #492, opened 2023-05-25
fix(pie-monorepo): correct turborepo dependency tree for test tasks

siggerzz:

**pie-monorepo**
[Fixed] - Issue where `yarn test` cache wasn't invalidated where `__snapshots__` directory was updated.
[Removed] - `@justeattakeaway/pie-icons` dependency

**pie-docs**
[Added] - `@justeattakeaway/pie-icons` as an explicit dependency to ensure dependant tasks are correct for pie-docs

### Reviewer 1

### Reviewer 2


package.json
```diff
@@ -101,7 +101,6 @@
   },
   "dependencies": {
     "@justeat/fozzie": "10.11.1",
-    "@justeattakeaway/pie-icons": "3.2.0",
```
siggerzz, line 104: Now that pie-icons is defined in the pie-docs/package.json, we don't need to define this dependency anymore.

This was needed because previously, Turborepo didn't know pie-icons was part of the build chain for pie-docs, as it was a pie-monorespo dependency

turbo.json
```diff
@@ -33,8 +33,7 @@
         "PIE_URL_PREFIX"
       ],
       "dependsOn": [
-        "^build",
-        "@justeattakeaway/pie-icons#build"
```
siggerzz, line 37: Now that pie-icons is defined in the pie-docs/package.json, we don't need to define this dependency anymore.

This was needed because previously, Turborepo didn't know pie-icons was part of the build chain for pie-docs, as it was a pie-monorespo dependency

turbo.json
```diff
@@ -43,8 +42,7 @@
     "pie-docs#build:dev": {
       "cache": true,
       "dependsOn": [
-        "^build",
-        "@justeattakeaway/pie-icons#build"
```
siggerzz, line 47: ditto

turbo.json
```diff
@@ -77,9 +75,13 @@
     "test": {
       "cache": true,
       "dependsOn": [
-        "build"
+        "build",
+        "^test"
+      ],
+      "outputs": [
```
siggerzz, line 81: This new config ensures that the following is done before running unit tests:

- The component `build` task is executed
- Unit tests pass for any dependencies

It also invalidates the turborepo cache if any changes have been made to the snapshot directories.

apps/pie-docs/package.json
```diff
@@ -32,6 +32,7 @@
   "devDependencies": {
     "@11ty/eleventy": "1.0.2",
     "@11ty/eleventy-navigation": "0.3.5",
+    "@justeattakeaway/pie-icons": "3.2.0",
```
webbertakken, line 35: This isn't a dev dependency though, is it?

webbertakken (dismissed):
Nice work

apps/pie-docs/package.json
fernandofranca: Is that a case for not using directly the workspace?

fernandofranca (dismissed):
Yay!

apps/pie-docs/package.json
siggerzz: @fernandofranca Just replicating the version number that was being used previously.

Having it hardcoded also has the advantage of not being forced to refactor consumers right away when a change is made. Something we need to agree within the team I think on how to handle deps like this.

@webbertakken Ah yeah, good spot! Force of habit. Will update :D

jamieomaguire (dismissed):

jamieomaguire:
Nice one!

jamieomaguire (approved):

packages/components/pie-button/package.json
```diff
@@ -10,6 +10,7 @@
     "build": "run -T vite build",
     "watch": "run -T vite build --watch",
     "test": "echo \"Error: no test specified\" && exit 0",
+    "test:ci": "yarn test",
```
webbertakken, line 13: This should not be needed. CI systems have an environment variable called `CI` set to `true`. Quality tools work with that variable natively. This includes Vitest. 

Usually in a modern repo I'd expect two jobs:
```
    "test": "vitest",
    "coverage": "vitest --coverage",
```

CI would then simply `run: yarn coverage`

webbertakken (approved):

fernandofranca (approved):

jamieomaguire (approved):

packages/components/pie-button/package.json
jamieomaguire: this is fair point @webbertakken but @siggerzz please raise a separate PR if you wish to address this :)

PR #510, opened 2023-06-05
fix(pie-icons-react): rename css icon classes to match vue lib

dandel10n:

[Fixed] Renames css icon classes to match size classes and vue library


packages/tools/pie-icons-react/scripts/react-icons-gen.js
```diff
@@ -66,7 +66,7 @@ async function build() {
         const { pathPrefix } = icons[iconKey];
         const capitalisedPathPrefix = (pathPrefix !== undefined ? (pathPrefix).substring(1, 2).toUpperCase() + (pathPrefix).substring(2) : '');
         const componentName = `Icon${capitalisedPathPrefix + pascalCase(handleComponentName(iconKey))}`;
-        const iconClasses = `pie-icon pie-icon--${iconKey.toLowerCase()}`;
+        const iconClasses = `c-pieIcon c-pieIcon--${iconKey.toLowerCase()}`;
```
fernandofranca, line 69: With the class name consistency, I'd extract `c-pieIcon` class name / prefix to the `pie-icons-configs` package and update both `pie-icons-vue` + `pie-icons-react`.

fernandofranca (approved):

raoufswe (approved):

packages/tools/pie-icons-react/scripts/react-icons-gen.js
dandel10n: I thought about this but it will take more time to implement while this change is quick and effective :) I can create a separate ticket for refactoring this part.

dandel10n: A follow-up: css icon size classes were removed in [this PR](https://github.com/justeattakeaway/pie/pull/532) in favour of applying width and height to the regular icons similar to large icons behaviour. getSvgProps() was moved to configs.js 🎊

PR #530, opened 2023-06-09
feat(pie-modal): Modal heading props

raoufswe:
- adds heading props to modal
- modifies the react-wrapper script to take care of cases where the custom element is defined via a variable

### Reviewer 1

### Reviewer 2

### DIFF
--- a/packages/components/pie-modal/README.md
+++ b/packages/components/pie-modal/README.md
@@ -52,6 +52,8 @@ import { PieModal } from '@justeattakeaway/pie-button/dist/react';
 | Property | Type      | Default | Description                                     |
 |----------|-----------|---------|-------------------------------------------------|
 | isOpen   | `Boolean` | `false` | Controls if the modal element is open or closed |
+| heading*   | `String` | - | Sets the heading of the modal |
+| headingLevel   | `String` | `h2` | Allows you to set the heading tag (from `h1` to `h6`) |
 
 
 ## Testing
@@ -77,4 +79,3 @@ Under scripts `test:visual` replace the environment variable with the below:
 
 ```
 PERCY_TOKEN_PIE_MODAL=abcde
-
--- a/packages/components/pie-modal/src/defs.ts
+++ b/packages/components/pie-modal/src/defs.ts
@@ -1,0 +1,11 @@
+/**
+ * Modal heading levels/tags
+ */
+export enum HEADING_LEVELS {
+  H1 = 'h1',
+  H2 = 'h2',
+  H3 = 'h3',
+  H4 = 'h4',
+  H5 = 'h5',
+  H6 = 'h6'
+}
--- a/packages/components/pie-modal/src/index.ts
+++ b/packages/components/pie-modal/src/index.ts
@@ -17,7 +17,7 @@ export class PieModal extends RtlMixin(LitElement) {
 
     @property({ type: String })
     @requiredProperty(componentSelector)
-        heading: string | undefined;
+        heading!: string;
 
     @property()
     @validPropertyValues(componentSelector, Object.values(HEADING_LEVELS), HEADING_LEVELS.H2)
--- a/packages/components/pie-modal/src/index.ts
+++ b/packages/components/pie-modal/src/index.ts
@@ -1,18 +1,38 @@
-import { LitElement, html, unsafeCSS } from 'lit'; // eslint-disable-line import/no-extraneous-dependencies
-import { property } from 'lit/decorators.js';
-import { RtlMixin } from '@justeattakeaway/pie-webc-core';
+import { LitElement, unsafeCSS } from 'lit'; // eslint-disable-line import/no-extraneous-dependencies
+import { html, unsafeStatic } from 'lit/static-html.js';
+import { property } from 'lit/decorators.js'; // eslint-disable-line import/no-extraneous-dependencies
+import { RtlMixin, validPropertyValues, requiredProperty } from '@justeattakeaway/pie-webc-core';
 
 import styles from './modal.scss?inline';
+import { HEADING_LEVELS } from './defs';
+
+// Valid values available to consumers
+export { HEADING_LEVELS };
+
+const componentSelector = 'pie-modal';
 
 export class PieModal extends RtlMixin(LitElement) {
     @property({ type: Boolean })
         isOpen = false;
 
-    // eslint-disable-next-line class-methods-use-this
+    @property({ type: String })
+    @requiredProperty(componentSelector)
+        heading: string | undefined;
+
+    @property()
+    @validPropertyValues(componentSelector, Object.values(HEADING_LEVELS), HEADING_LEVELS.H2)
+        headingLevel : HEADING_LEVELS = HEADING_LEVELS.H2;
+
     render () {
+        const {
+            heading,
+        } = this;
+
+        const headingTag = unsafeStatic(this.headingLevel);
+
         return html`
             <dialog class="c-modal" ?open="${this.isOpen}">
-                <h3 class="c-modal-heading">Modal header</h3>
+                <${headingTag} class="c-modal-heading">${heading}</${headingTag}>
                 <div class="c-modal-contentWrapper">
                     <slot></slot>
                 </div>
@@ -24,7 +44,7 @@ export class PieModal extends RtlMixin(LitElement) {
     static styles = unsafeCSS(styles);
 }
 
-customElements.define('pie-modal', PieModal);
+customElements.define(componentSelector, PieModal);
 
 declare global {
     interface HTMLElementTagNameMap {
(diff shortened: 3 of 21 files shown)


packages/components/pie-modal/README.md
```diff
@@ -52,6 +52,8 @@ import { PieModal } from '@justeattakeaway/pie-button/dist/react';
 | Property | Type      | Default | Description                                     |
 |----------|-----------|---------|-------------------------------------------------|
 | isOpen   | `Boolean` | `false` | Controls if the modal element is open or closed |
+| heading*   | `String` | - | Sets the heading of the modal |
```
kevinrodrigues, line 55: I think we should add in another column to state which props are required. Then you can mark the props you added as required `yes`

raoufswe: I thought about the same thing but then decided to use `*` to indicate a prop is required as there will be few required props compared to optional and the table might be a bit hard to digest. Skip's [storybook](https://react.pie.design/?path=/story/components-modal--playground) has a similar table structure. 

as a side note, ideally, we should generate the metadata through jsDoc or the typescript definition as this approach will get out of sync easily but something to keep in mind

kevinrodrigues (dismissed):

packages/components/pie-modal/README.md
jamieomaguire: Fully agree - we should automate this!

packages/components/pie-modal/src/index.ts
```diff
@@ -1,18 +1,38 @@
...
+const componentSelector = 'pie-modal';
 
 export class PieModal extends RtlMixin(LitElement) {
     @property({ type: Boolean })
         isOpen = false;
 
-    // eslint-disable-next-line class-methods-use-this
+    @property({ type: String })
+    @requiredProperty(componentSelector)
+        heading: string | undefined;
```
jamieomaguire, line 20: if it's required then I don't think it can be `undefined`, right?

packages/components/pie-modal/src/index.ts
```diff
@@ -1,18 +1,38 @@
...
+        const {
+            heading,
+        } = this;
+
+        const headingTag = unsafeStatic(this.headingLevel);
+
         return html`
             <dialog class="c-modal" ?open="${this.isOpen}">
-                <h3 class="c-modal-heading">Modal header</h3>
+                <${headingTag} class="c-modal-heading">${heading}</${headingTag}>
```
jamieomaguire, line 35: could you potentially do:

```
<h${headingLevel}>${heading}</h${headingLevel>
```

This way you wouldn't need to call `unsafeStatic`?

jamieomaguire: in this example, heading level would be a string number like '1', '2' etc

packages/components/pie-modal/test/visual/pie-modal.spec.ts
```diff
@@ -2,21 +2,21 @@ import { test } from '@sand4rt/experimental-ct-web';
 import percySnapshot from '@percy/playwright';
 import { PieModal } from '@/index';
 
-const propIsOpenValues = [true, false];
+const propIsOpenValues = [
+    { heading: 'Modal Header / isOpen = true', isOpen: true },
```
jamieomaguire, line 6: whilst this is perfectly reasonable, if you need to label the tests I would suggest using the test component wrapper and providing a string label.

Entirely up to you but it might be nicer for consistency. Example can be found in the pie-button visual tests

packages/components/pie-modal/src/index.ts
raoufswe: I agree, but I get this error if I don't specify if the component potentially won't be assigned, do you have any idea?
Screenshot 2023-06-09 at 16 43 03

jamieomaguire:
> Screenshot 2023-06-09 at 15 47 00
> Storybook doesn't appear to be working @raoufswe

@raoufswe it actually seems like a refresh fixes this but it broke for me when I clicked on each component and had to refresh. Does this occur for you too?

jamieomaguire:
> @jamieomaguire, yes, it's the issue I shared with you yesterday. the core package is imported several times and it declares the test custom element every time it's imported, thus, the error.

yes that makes sense - can we add a temporary registry check then pull that test code into a separate package next week as discussed?

packages/components/pie-modal/src/index.ts
jamieomaguire: I don't have an immediate suggestion but maybe we can look into this a bit more as I think it could make the codebase messy if we set a precedent here

packages/components/pie-modal/src/index.ts
raoufswe: Unfortunately, I believe you still need to use `unsafeStatic` as it's still considered a static expression. I find [the lit doc ](https://lit.dev/docs/templates/expressions/#static-expressions)about the topic quite useful. but in a nutshell, we have a validation layer before using `unsafeStatic ` and only valid heading tags will be passed through with the help of `@validPropertyValues` 

Screenshot 2023-06-09 at 16 50 57

jamieomaguire: ah, that makes sense. I'm still unsure if I like the idea of a vanilla consumer having to write:

`<pie-modal heading-level="h2" />`

but it's probably fine!

raoufswe: It's very common to pass the html tag as a prop as far as I know, tho I would argue the benefits such flexibility will provide to our users, but I assume it was in the requirements for a reason. 

I'm happy to make the prop numerical if it makes more sense and the rest of the team 🚀💯

packages/components/pie-modal/src/defs.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Modal heading levels/tags
+ */
+export enum HEADING_LEVELS {
```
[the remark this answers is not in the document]
raoufswe: I will check before merging the PR if the union literal work is ready or not, if not, I reckon we shouldn't block the PR because of that and assume a clean up will needed 👍

packages/components/pie-modal/src/index.ts
raoufswe: I did a bit of digging and I believe we can use something called _definite assignment assertion operator_ or `!` in short to achieve a semi-required behavior in classes and ts. 

assuming the title is a required field and should always be string, setting `heading!: string` will tell the compiler that the heading property will be instantiated with a value before it's used, even though it might not be apparent for you now, but every time we get or set that property, we want it to be treated it as a type of string. 👓

packages/components/pie-modal/test/visual/pie-modal.spec.ts
raoufswe: actually, I might've just been a bit expressive here and reverted that.  The tests are to make sure the component is open or not and I thought we should add the state to the heading, but the heading won't render anyway because the modal is closed and not in the DOM 😵‍💫 😆

packages/components/pie-modal/src/index.ts
```diff
@@ -1,18 +1,40 @@
...
     @property({ type: Boolean })
         isOpen = false;
 
-    // eslint-disable-next-line class-methods-use-this
+    @property({ type: String })
+    @requiredProperty(componentSelector)
+        heading!: string;
+
+    @property()
+    @validPropertyValues(componentSelector, Object.values(HEADING_LEVELS), HEADING_LEVELS.H2)
```
ashleynolan, line 23: I'm guessing we have some generic tests around the `validPropertyValues` decorator, but is it worth adding specific tests for modal where we pass in the heading types and explicitly test these usecases?

Just to cover if we ever hit an edge case that isn't covered by the generic function tests.

ashleynolan (commented):
Looks good – mostly tiny NIT things from me. Otherwise, looks good to merge

packages/components/pie-modal/src/index.ts
```diff
@@ -1,18 +1,40 @@
...
+const componentSelector = 'pie-modal';
 
 export class PieModal extends RtlMixin(LitElement) {
     @property({ type: Boolean })
         isOpen = false;
 
-    // eslint-disable-next-line class-methods-use-this
+    @property({ type: String })
+    @requiredProperty(componentSelector)
+        heading!: string;
```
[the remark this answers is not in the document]
raoufswe: i shared about that [here](https://github.com/justeattakeaway/pie/pull/530#discussion_r1224738121)

fernandofranca (dismissed):
Good stuff!!! 🚀

packages/components/pie-modal/src/index.ts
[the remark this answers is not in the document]
ashleynolan: Thanks – makes sense now 👍

ashleynolan (approved):

fernandofranca (approved):

PR #536, opened 2023-06-14
fix(pie-monorepo): fix majority of TS/ESlint issues

jamieomaguire:
---
"pie-storybook": minor
---

[Changed] - use our eslint config, fix various linting issues and use render function syntax for story declarations

---
"@justeattakeaway/pie-icon-button": minor
---

[Changed] - use our eslint config, remove eslint comments

---
---

[Changed] - use our eslint config and fix various linting issues

---
---

[Changed] - use our eslint config, remove eslint comments

---
"@justeattakeaway/eslint-config-pie": minor
---

[Added] - disabled 'import/no-extraneous-dependencies', allow omitting of .ts and .js file extensions with 'import/extensions' and update the 'import/resolver' to recognise .ts files

---
"@justeattakeaway/pie-icons-vue": minor
---

[Removed] - eslint comments

---
---

[Changed] - use our eslint config, add SCSS type declarations, remove eslint comments

---
"@justeattakeaway/pie-components-config": minor
---

[Changed] - add our eslint config, remove eslint comments and .js file extensions from imports


siggerzz:
Just a reminder that changesets should only be added for consumer facing changes - https://github.com/justeattakeaway/pie/blob/main/README.md#versioning--publishing-packages

I feel like we'll be unnecessarily be creating unneeded versions / npm publishes in it's current state. In fairness, this is something we probably need to make more visible, as I've only been pointing this out in PR's on an ad-hoc basis, so apologies if this rule has been broken a lot already.

jamieomaguire:
> Just a reminder that changesets should only be added for consumer facing changes - https://github.com/justeattakeaway/pie/blob/main/README.md#versioning--publishing-packages
> 
> I feel like we'll be unnecessarily be creating unneeded versions / npm publishes in it's current state. In fairness, this is something we probably need to make more visible, as I've only been pointing this out in PR's on an ad-hoc basis, so apologies if this rule has been broken a lot already.

Sure thing - how do you define consumer facing @siggerzz? 

Perhaps the changeset cli shouldn't include the packages we don't need changesets for in that case. Easier to not add incorrect files that way.

siggerzz:
@jamieomaguire I guess anything that has an effect on users of the **published** package and not 'internal' monorepo specific changes like the ones mentioned in the readme.

For example a change to `dependencies` / `peerDependencies` would warrant a changeset, as that has a potential impact on a consumer of the package.

However, a change to `devDependencies` would not need a changeset, as those are only used internally.

In the case of this PR, these linting changes have an effect internally within the monorepo, but an end-user of the package does not need to know / care about such changes.

jamieomaguire:
@siggerzz yeah makes sense. I've removed most of the changes but the changes to our listing package will affect consumers of it so I've left that in 👍

siggerzz (dismissed):

ashleynolan (dismissed):

ashleynolan (approved):

ashleynolan (approved):

xander-marjoram (approved):

PR #549, opened 2023-06-19
feat(pie-modal): Add close button

kevinrodrigues:
[Added] - close button to modal


packages/components/pie-modal/test/component/pie-modal.spec.ts
```diff
@@ -25,3 +25,28 @@ Object.values(HEADING_LEVELS).forEach((headingLevel) => test(`should render the
...
+
+        await component.update({
+            on: {
+                click: (event: string) => messages.push(event)
+            }
+        });
+
+        await component.locator('.c-modal-closeBtn').click();
+
+        expect(messages).toHaveLength(1);
```
kevinrodrigues, line 49: I'm not too happy with this part of the test. Have tried to pull out the event name in the expects which I think should be doable but need to read up a bit more to work out the best way to do that. For now it picks up the event length.

xander-marjoram: I'm assuming you've tried something like this?
```suggestion
        expect(messages).toHaveLength(1);
        const [message] = messages;
        expect(message.type).toBe(ON_MODAL_CLOSE_EVENT);
```

packages/components/pie-modal/src/index.ts
```diff
@@ -23,6 +26,9 @@ export class PieModal extends RtlMixin(LitElement) {
     @validPropertyValues(componentSelector, headingLevels, 'h2')
         headingLevel: ModalProps['headingLevel'] = 'h2';
 
+    @query('dialog')
```
xander-marjoram, line 29: This is a nice use for query

packages/components/pie-modal/test/component/pie-modal.spec.ts
xander-marjoram: You could even throw in an `expect(message instanceof CustomError).toBe(true);`

packages/components/pie-modal/src/constants.ts
```diff
@@ -0,0 +1,8 @@
+// MODAL EVENTS
+
+/**
+ * Event name for when the modal is closed.
+ *
+ * @constant
+ */
+export const ON_MODAL_CLOSE_EVENT = 'pie-modal-close';
```
raoufswe, line 8: can't we just add this to defs.ts?

packages/components/pie-modal/src/index.ts
```diff
@@ -33,15 +39,39 @@ export class PieModal extends RtlMixin(LitElement) {
...
+    };
+
+    /**
+     * Dispatch `ON_MODAL_CLOSE_EVENT` event.
+     * To be used whenever we close the modal.
+     *
+     * @event
+     */
+    _onDialogClose = () => {
+        let event = new CustomEvent(ON_MODAL_CLOSE_EVENT);
```
raoufswe, line 71: doesn't seem to be reassigned
```suggestion
        const event = new CustomEvent(ON_MODAL_CLOSE_EVENT);
```

packages/components/pie-modal/test/component/pie-modal.spec.ts
kevinrodrigues: I did try the message thing (good shout) but didn't work. I feel there is something else going on so will reattempt this test in my next ticket.

packages/components/pie-modal/src/constants.ts
raoufswe: I'm not really sure tbh, but I would assume defs stand for definitions :D 
sorry, i didn't use the defs convention before but I would just use it to not have too many files unless its against the convention

packages/components/pie-modal/src/index.ts
```diff
@@ -33,15 +39,39 @@ export class PieModal extends RtlMixin(LitElement) {
         const headingTag = unsafeStatic(headingLevel);
 
         return html`
-            <dialog class="c-modal" ?open="${isOpen}">
-                <${headingTag} class="c-modal-heading">${heading}</${headingTag}>
-                <div class="c-modal-contentWrapper">
-                    <slot></slot>
-                </div>
+            <dialog id="dialog" class="c-modal" ?open="${isOpen}" @click="${this._handleCloseDialog}">
```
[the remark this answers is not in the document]
kevinrodrigues: That's my bad i was testing something related to the test as i was having issues with testing the event and added it here to check what was happening. Will remove.

packages/components/pie-modal/src/constants.ts
kevinrodrigues: This is what I read so i am not sure it is the same? Happy to move it there but maybe we need some clarity on this file and the purpose it can be used with the correct intention?

`TypeScript definition files (.d.ts files) are used to provide type information for external JavaScript libraries or modules that do not have built-in TypeScript support. These definition files describe the shape and types of the existing JavaScript code, enabling TypeScript to perform static type checking and provide enhanced tooling support.`

So to me it looks like the defs file is for defining the shape of javascript so TS can allow for type checking and provide better support.

raoufswe: aha then, in this case, we aren't following that as we export things like ButtonSizes, headingLevels, etc which are kinda a constant to some extent as well.

raoufswe:
more of a question: when we close the modal, is it expected that the dialog element stays in the DOM?

xander-marjoram (approved):

raoufswe (approved):

kevinrodrigues:
> more of a question: when we close the modal, is it expected that the dialog element stays in the DOM?

Yeah the default behaviour of the dialog element toggles between display none and display block.

packages/components/pie-modal/src/index.ts
```diff
@@ -33,15 +36,39 @@ export class PieModal extends RtlMixin(LitElement) {
...
+                    <${headingTag} class="c-modal-heading">${heading}</${headingTag}>
+                    <pie-icon-button
+                        @click="${this._handleCloseDialog}"
+                        variant="ghost-tertiary"
+                        class="c-modal-closeBtn"></pie-icon-button>
+                </header>
+                <article>
+                    <div class="c-modal-contentWrapper">
+                        <slot></slot>
+                    </div>
```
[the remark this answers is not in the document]
kevinrodrigues: Could i look at this in my next PR please? I do think if i move it tho i would have to then maybe target the div directly in the sass so may not add any benefit but will take a look anyway.

packages/components/pie-modal/src/constants.ts
ashleynolan: Doesn't need to be done as part of this PR, but would be good to get some advice on this from @thejfreitas or @davidpn11 to check that we're defining these in the right way (if there is a 'right' way 😄

packages/components/pie-modal/test/component/pie-modal.spec.ts
```diff
@@ -25,3 +25,28 @@ headingLevels.forEach((headingLevel) => test(`should render the correct heading
     // h2 is the default / fallback value
     await expect(component.locator('h2.c-modal-heading')).toContainText(props.heading);
 }));
+
+test.describe('`When the Pie Modal is closed`', () => {
+    test.skip('should dispatch event `pie-modal-close`', async ({ mount }) => {
```
kevinrodrigues, line 30: Will come back to this test in my next pr. Locally it all passes but on CI it fails due to a timeout.

packages/components/pie-modal/src/modal.scss
```diff
@@ -53,4 +53,10 @@
         padding-block: var(--modal-content-padding-block-start) var(--modal-content-padding);
         padding-inline: var(--modal-content-padding);
     }
+
+    & .c-modal-closeBtn {
+        position: absolute;
+        right: var(--dt-spacing-d);
+        top: var(--dt-spacing-d);
```
xander-marjoram, line 60:
```suggestion
        inset-inline-end: var(--dt-spacing-d);
        inset-block-start: var(--dt-spacing-d);
```

packages/tools/pie-wrapper-react/custom-elements.json
```diff
@@ -326,6 +357,15 @@
           },
           "default": "['h1', 'h2', 'h3', 'h4', 'h5', 'h6']",
           "description": "Modal heading levels/tags"
+        },
+        {
+          "kind": "variable",
+          "name": "ON_MODAL_CLOSE_EVENT",
```
xander-marjoram, line 363: I think I missed when this was all added, but it seems... suboptimal that we need to update this package any time something changes in any component.

kevinrodrigues: I believe this was auto added.

ashleynolan: For context – this gets added as part of the react-wrapper so that reactive properties get mapped as part of the react component build.

This is autogenerated by the custom elements package that runs as part of the react-wrapper build step

packages/components/pie-modal/src/modal.scss
kevinrodrigues: Could I add this as part of my next pr please :)

xander-marjoram (approved):

PR #551, opened 2023-06-20
refactor(pie-modal): updated styles to use custom props

ashleynolan:

- [Changed] - Updated styles to use custom props

### Reviewer 1

### Reviewer 2

### DIFF
--- a/packages/components/pie-modal/src/modal.scss
+++ b/packages/components/pie-modal/src/modal.scss
@@ -1,51 +1,56 @@
-$modal-size-l: 1080px;
-$modal-size-m: 600px;
-$modal-size-s: 450px;
-
-$modal-border-radius: var(--dt-radius-rounded-d);
-$modal-font: var(--dt-font-interactive-m-family);
-$modal-bg-color: var(--dt-color-container-default);
-$modal-elevation: var(--dt-elevation-04);
-
-$modal-header-font-size: calc(var(--dt-font-heading-m-size--wide) * 1px);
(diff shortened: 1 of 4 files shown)


packages/components/pie-modal/src/modal.scss
```diff
@@ -1,51 +1,56 @@
...
+    background-color: var(--modal-bg-color);
+
+    padding: 0;
+    inline-size: var(--modal-size-m);
+
+    box-shadow: var(--modal-elevation);
+
+    & .c-modal-heading {
+        // Modal header Custom Props
+        --modal-header-font-size: calc(var(--dt-font-heading-m-size--wide) * 1px);
```
sebastienrocks, line 26: I'm curious, why the `* 1px`?

sebastienrocks (approved):

packages/components/pie-modal/src/modal.scss
ashleynolan: Good question! So we store our sizes as unitless values (as we use `px` on web, iOS use `pt` – I'm not sure what Android uses).

You can't just add a 'px' string to this number, otherwise you get a type mismatch in the variable, so we have to multiply it by 1px to convert it to a pixel value.

We'll be looking at turning this into a shared function/mixin shortly when we pull together some useful shared styling for use across components.

sebastienrocks: Nice one, I didn't even know that is a thing, mind => blown

PR #554, opened 2023-06-20
feat(pie-docs): add anchors to headings

FayeCarter:
## Changes
[Updated] - `pieIconsSvg` filter to handle icon colours
[Updated] - `markdown` filter to add anchors to all markdown headings
[Added] - `headingAnchor` filter to add anchors to all HTML headings[Added] - `scroll-margin-top` to elements so they do not scroll behind the tab container

https://github.com/justeattakeaway/pie/assets/24740015/c32c587b-b2a7-4ff5-b953-e9593fb70a49


LTurns (approved):

jamieomaguire:
Screenshot 2023-06-21 at 09 23 47

@FayeCarter should these be vertically centred between lines on mobile or aligned to the centre of the first line?

jamieomaguire:
@FayeCarter on my iphone it seems like when you tap a heading and it scrolls to just beneath the heading, the sticky tabs cover the heading. Should we offset the scroll position to the height of the sticky tabs or something? It just looks a little funky :)

jordifornellsJET:
> @FayeCarter on my iphone it seems like when you tap a heading and it scrolls to just beneath the heading, the sticky tabs cover the heading. Should we offset the scroll position to the height of the sticky tabs or something? It just looks a little funky :)

That would be excellent if it could work that way, I have the same thing on my Android phone

apps/pie-docs/src/_utilities/markdown.js
```diff
@@ -1,8 +1,33 @@
...
 
 const md = new MarkdownIt({
     html: true,
-}).use(markdownItAttrs);
+}).use(markdownItAttrs)
+    .use(anchor, {
+        level: [2, 3, 4],
+        permalink: anchor.permalink.linkInsideHeader({
+            symbol: `
+                <span class="is-visuallyHidden">Jump to heading</span>
```
jamieomaguire, line 24: is the idea that a screen reader focuses the heading and then the span is read out before they press enter/select the anchor? Just wondering if "Jump to heading" is clear enough etc. Haven't seen this pattern before but it's a cool idea!

jamieomaguire:
Looks great! Just a couple of comments 👍

apps/pie-docs/src/assets/styles/components/_anchorHeading.scss
```diff
@@ -0,0 +1,47 @@
+@use 'sass:color';
+@use '../fozzie-config' as f;
+
+$anchor-size: f.spacing(e);
+$anchor-margin: f.spacing(b);
+
+.c-anchor-icon {
```
ashleynolan, line 7: Small one – the naming is a bit conflicted at then moment. The filename is for `anchorHeading` and the component is `c-anchor-icon`.

I'd change them to match – I think just `anchor` or `anchorIcon` makes sense tbh, so you could just go for `_anchorIcon.scss` and `c-anchorIcon`

xander-marjoram:
Very nice! Happy to approve once the issue @jamieomaguire highlighted has been resolved.

apps/pie-docs/src/_utilities/markdown.js
FayeCarter: I think it was mostly an experiment on what I could get to work. The screen reader seems to ignore it anyway so I will remove

apps/pie-docs/src/assets/styles/components/_anchorHeading.scss
FayeCarter: Hey @ashleynolan, agreed, and i tried this. But as I have now added the span around the heading text I'm not sure if the naming still works?

I've gone for 
file name: `anchor.scss`
icon class: `c-anchor-icon`
heading text `c-anchor-text`

Not confident I've nailed it there so very happy to change it up again. What do you think?

FayeCarter: Ignore this 😄

apps/pie-docs/src/_11ty/filters/headingAnchor.js
```diff
@@ -0,0 +1,35 @@
+const slugify = require('slugify');
```
jamieomaguire, line 1: I'm interested to know why you chose to use a filter rather than an additional shortcode?

Can you remind me how we can use this granularly on a content page for some headings and not others please?

FayeCarter: I'm assuming you mean the anchor, not slugify?

The content page uses a markdown plugin so it only adds the anchor to markdown files.

We need a filter because we need to apply the anchor to the shortcodes. These don't have anchors automatically applied cos they're js, not md. We also wouldn't want them applied to every single shortcode because there are headings in shortcodes like `card` that we do not want the anchor on.

The filter can be applied directly to the shortcodes we want anchors on, `tokenTable` and `categorisedIconList`  as you can see here https://github.com/justeattakeaway/pie/pull/554/files#diff-6edac4cb8e6a5d69221c82c758f8476016f8228c2bd492055f947c6732efcfd2R56

I hope that answers your question

jamieomaguire: Yeah it was regarding this entire file.

I was wondering if in an `md` file we'd ever want something like

```
{% anchorHeading { 'My Heading', '#some-link' %}
```

Added to the page directly?

FayeCarter: Maybe it would be useful at some point it might make sense to, but I think it makes more sense to automatically apply the anchor to all headings instead of manually adding this in instead of the headings

xander-marjoram (approved):

apps/pie-docs/src/_11ty/filters/headingAnchor.js
jamieomaguire: Ok, so just to check I've understood - with this change are we only able to add anchor headings to shortcodes and not directly to markdown pages?

jamieomaguire (approved):

jamieomaguire:
Looks good to me but its worth noting you can now focus headings with the tab key and there is no focus styling that I can see on the preview

apps/pie-docs/src/_11ty/filters/headingAnchor.js
xander-marjoram: @jamieomaguire The markdown filter has also been extended so it adds anchors there too

FayeCarter: Nope we have that automatically.

The markdown files will all automatically have the anchor added because of the markdown utility file
[this guy](https://github.com/justeattakeaway/pie/pull/554/files#diff-b14132ae0275f8dad8ab29be8a443846b462d85ed467be4bacf6d186e4ce0247R20)

The file this comment is on is the filter to filter the shortcodes that we want to have an anchor on, and add it to them 👍

FayeCarter:
Just a note as I will be away from next week, this has been design approved by Naz

PR #556, opened 2023-06-20
fix(pie-icons-configs): removes a console error if iconSize prop is not passed

dandel10n:
[Fixed]: removes a console error if `iconSize` prop is not passed
[Changed]: in case of invalid `iconSize` being passed to large icons now the size defaults to the `largeIconSizeDefault`

### DIFF
--- a/packages/tools/pie-icons-configs/__tests__/configs.test.js
+++ b/packages/tools/pie-icons-configs/__tests__/configs.test.js
@@ -112,70 +112,129 @@ describe('validateGetRegularIconSize', () => {
 });
 
 describe('getSvgProps', () => {
-    describe('when provided with a class for regular size icon', () => {
-        const iconName = 'IconAppOrder';
-
-        describe('when provided with valid parameters', () => {
-            it('returns an object that contains a class and width and height properties', () => {
-                const received = getSvgProps('pie-icon pie-icon--app-order', null, regularIconSizeDefault, iconName);
+    const regularIconSizeName = 'IconTest';
+    describe('when provided with a', () => {
+        describe('svgClasses parameter', () => {
+            it('returns an object that contains a class attr with the expected classes', () => {
+                const received = getSvgProps('icon-test', null, null, regularIconSizeName);
 
                 expect(received).toHaveProperty('class');
-                expect(received).toHaveProperty('width');
-                expect(received).toHaveProperty('height');
-            });
-
-            it('has expected width and height values', () => {
-                const received = getSvgProps('pie-icon pie-icon--app-order', null, regularIconSizeDefault, iconName);
-                const expectedRegularIconSizeValue = sizeToValueMap[regularIconSizeDefault];
-
-                expect(received.width).toEqual(expectedRegularIconSizeValue);
-                expect(received.height).toEqual(expectedRegularIconSizeValue);
-            });
-
-            describe('when provided with a staticClasses parameter', () => {
-                it('returns an object with the expected classes', () => {
-                    const received = getSvgProps('pie-icon pie-icon--app-order', 'FOO BAR', regularIconSizeDefault, iconName);
-
-                    expect(received.class).toEqual(expect.stringContaining('pie-icon'));
-                    expect(received.class).toEqual(expect.stringContaining('pie-icon--app-order'));
-                    expect(received.class).toEqual(expect.stringContaining('FOO BAR'));
-                });
+                expect(received.class).toEqual('icon-test');
             });
         });
 
-        describe('when provided with invalid parameters', () => {
-            it('returns an object with the expected classes', () => {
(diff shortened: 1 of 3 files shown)


FayeCarter (approved):

packages/tools/pie-icons-configs/__tests__/configs.test.js
```diff
@@ -144,23 +146,45 @@ describe('getSvgProps', () => {
 
         describe('when provided with invalid parameters', () => {
             it('returns an object with the expected classes', () => {
-                const received = getSvgProps('pie-icon pie-icon--app-order', 'FOO', null, iconName);
+                const received = getSvgProps('pie-icon pie-icon--app-order', null, 'n', iconName);
```
ashleynolan, line 149: Bit confused about this change. Did the test pass fine with the old params – and if so, should we be testing other outputs as well? Or is it ok that the tests pass when both `…'Foo', null` is passed in and `null, 'n'`, but the expects are the same?

ashleynolan (dismissed):
Looks good – just had one q about the test params, but only a small thing

packages/tools/pie-icons-configs/__tests__/configs.test.js
dandel10n: Test pass with both old and new params. I just changed it for consistency. 'FOO' parameter was not related to the test, as it is a static css class that can be passed, and 'n' is invalid value while null is a valid value in this case that should mean that default width and height should be applied without a console error (covered in test below). I can rewrite the descriptions to make it clearer.

packages/tools/pie-icons-configs/__tests__/configs.test.js
```diff
@@ -110,126 +112,136 @@ describe('validateGetRegularIconSize', () => {
 });
 
 describe('getSvgProps', () => {
-    describe('when provided with a class for regular size icon', () => {
-        const iconName = 'IconAppOrder';
+    describe('when provided with a', () => {
+        describe('svgClasses parameter', () => {
+            it('returns an object that contains a class attr with the expected classes', () => {
+                const received = getSvgProps('icon-test', null, null);
```
xander-marjoram, line 118: It looks like the `getSvgProps` function has 4 parameters but we're only calling it with 3? I might be missing something but I'd expect the linter to complain, at least

dandel10n: The last parameter, `componentName`, is used for console error message so it is easier to understand where the error comes from. So I didn't pass it for tests that don't need it.

packages/tools/pie-icons-configs/__tests__/configs.test.js
```diff
@@ -110,126 +112,136 @@ describe('validateGetRegularIconSize', () => {
...
         });
 
-        describe('when provided with invalid parameters', () => {
-            describe('when provided with an invalid iconSize', () => {
-                const iconName = 'IconAppOrder';
-                const errorMock = vi.fn();
-                console.error = errorMock;
+        describe('when provided with invalid iconSizeValue parameter', () => {
+            const invalidSize = 49;
+            const expectedSize = 48;
```
xander-marjoram, line 201: Interesting that if they provide an invalid value we try to fix it for them, should we not just return the default? It might make it a bit more obvious that they've provided an invalid value (besides the error).

dandel10n: Yeah, it will align the behaviour between regular and large as well. I will add the change 👍

dandel10n: Done

packages/tools/pie-icons-configs/__tests__/configs.test.js
dandel10n: I added it to all the tests for consistency

packages/tools/pie-icons-configs/__tests__/configs.test.js
```diff
@@ -110,126 +112,136 @@ describe('validateGetRegularIconSize', () => {
...
+        const invalidIconSizeValue = 'n';
 
-                    expect(received.class).toEqual(expect.stringContaining('pie-icon'));
-                    expect(received.class).toEqual(expect.stringContaining('pie-icon--app-order'));
-                    expect(received.class).toEqual(expect.stringContaining('FOO BAR'));
-                });
-            });
+        it('returns an object with valid width and height properties', () => {
+            const errorMock = vi.fn();
+            console.error = errorMock;
```
xander-marjoram, line 153: It's probably better practice if we spy on this and mock the implementation like in the other tests

xander-marjoram: And I know it's not used in every test, but perhaps it's used enough now that we can justify pulling it out into a `beforeEach` to set up the mock and an `afterEach` to restore it

raoufswe (approved):

PR #559, opened 2023-06-21
Dsw style modal background

kevinrodrigues:
---
"pie-monorepo": minor
---

[Added] - Modal backdrop functionality

### DIFF
--- a/packages/components/pie-modal/src/index.ts
+++ b/packages/components/pie-modal/src/index.ts
@@ -24,7 +24,7 @@ export class PieModal extends RtlMixin(LitElement) {
         headingLevel: ModalProps['headingLevel'] = 'h2';
 
     @query('dialog')
-        _dialog: HTMLDialogElement;
+        _dialog!: HTMLDialogElement;
 
     render () {
         const {
@@ -40,22 +40,59 @@ export class PieModal extends RtlMixin(LitElement) {
                 <header>
                     <${headingTag} class="c-modal-heading">${heading}</${headingTag}>
                     <pie-icon-button
-                        @click="${this._handleCloseDialog}"
-                        variant="ghost-tertiary"
-                        class="c-modal-closeBtn"></pie-icon-button>
+                            @click="${this._handleCloseDialog}"
+                            variant="ghost-tertiary"
+                            class="c-modal-closeBtn"></pie-icon-button>
                 </header>
-                <article>
-                    <div class="c-modal-content">
-                        <slot></slot>
-                    </div>
+                <article class="c-modal-content">
+                    <slot></slot>
                 </article>
             </dialog>
         `;
     }
 
+    updated () {
+        this._handleModalLightDismiss();
+    }
+
+    /**
+     * We require this because toggling the prop `isOpen` itself won't
+     * allow the modal to open in the correct way (with the default background),
+     * the method `showModal()` needs to be invoked.
+     *
+     * https://lit.dev/docs/components/lifecycle/#willupdate
+     *
+     * @param changedProperties
+     */
+    willUpdate (changedProperties: Map<string, any>) {
+        if (changedProperties.has('isOpen') && this._dialog) {
+            const dialog = this._dialog;
+            const isClosed = changedProperties.get('isOpen') === false;
+
+            isClosed
+                ? dialog.showModal()
+                : dialog.close();
+        }
+    };
+
     _handleCloseDialog = () => {
         this._dialog.close();
         this._onDialogClose();
+    };
+
+    /**
+     * Dismiss modal via `_handleCloseDialog()` on backdrop click
+     * and will proceed to fire an `ON_MODAL_CLOSE_EVENT` event.
+     *
+     */
+    _handleModalLightDismiss = () => {
+        this._dialog.addEventListener('click', event => {
+            const rect = this._dialog.getBoundingClientRect();
+            const isWithinDialogArea = (rect.top <= event.clientY && event.clientY <= rect.top + rect.height
+                && rect.left <= event.clientX && event.clientX <= rect.left + rect.width);
+
+            if (!isWithinDialogArea) this._handleCloseDialog();
+        });
     };
 
     /**
--- a/packages/components/pie-modal/src/modal.scss
+++ b/packages/components/pie-modal/src/modal.scss
@@ -11,6 +11,8 @@
     --modal-bg-color: var(--dt-color-container-default);
     --modal-elevation: var(--dt-elevation-04);
 
+    --modal-overlay-color: var(--dt-color-overlay);
+
     border-radius: var(--modal-border-radius);
     border: none;
     font-family: var(--modal-font);
@@ -56,7 +58,15 @@
 
     & .c-modal-closeBtn {
         position: absolute;
-        right: var(--dt-spacing-d);
-        top: var(--dt-spacing-d);
+        inset-inline-end: var(--dt-spacing-d);
+        inset-block-start: var(--dt-spacing-d);
     }
+}
+
+::backdrop {
+  --color-backdrop: rgb(0,0,0,0.5);
+}
+
+dialog::backdrop {
+  background: var(--color-backdrop);
 }
--- a/packages/components/pie-modal/test/component/pie-modal.spec.ts
+++ b/packages/components/pie-modal/test/component/pie-modal.spec.ts
@@ -1,5 +1,4 @@
 import { test, expect } from '@sand4rt/experimental-ct-web';
-import * as events from 'events';
 import { PieModal } from '@/index';
 import { headingLevels } from '@/defs';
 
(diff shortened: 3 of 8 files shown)


packages/components/pie-modal/src/modal.scss
```diff
@@ -56,7 +58,15 @@
...
         position: absolute;
-        right: var(--dt-spacing-d);
-        top: var(--dt-spacing-d);
+        inset-inline-end: var(--dt-spacing-d);
+        inset-block-start: var(--dt-spacing-d);
     }
 }
+
+::backdrop {
+  --color-backdrop: rgb(0,0,0,0.5);
```
kevinrodrigues, line 67: I've tried to use the token `--dt-color-overlay:` but doesn't work for some reason, I bumped the token package and tried but the colour doesn't come through. Thoughts?

packages/components/pie-modal/test/component/pie-modal.spec.ts
```diff
@@ -27,27 +26,49 @@ headingLevels.forEach((headingLevel) => test(`should render the correct heading
...
 
-test.describe('`When the Pie Modal is closed`', () => {
-    test.skip('should dispatch event `pie-modal-close`', async ({ mount }) => {
-        const messages: string[] = [];
-        const component = await mount(
-            PieModal,
-            {
-                props: {
-                    isOpen: true,
+test.describe('`Pie Modal is closed`', () => {
```
kevinrodrigues, line 29: I'm still not fully convinced with these tests so will probably need to spend some time reading up on playwright. Ideally I want to be able to read the name of the event in the expects and not relie on the `1`

packages/components/pie-modal/src/modal.scss
xander-marjoram: Just checking you tried it with `var(--dt-color-overlay)`?

kevinrodrigues: Yup tried that.

packages/components/pie-modal/src/modal.scss
```diff
@@ -56,7 +56,11 @@
...
         position: absolute;
-        right: var(--dt-spacing-d);
-        top: var(--dt-spacing-d);
+        inset-inline-end: var(--dt-spacing-d);
+        inset-block-start: var(--dt-spacing-d);
     }
 }
+
+dialog::backdrop {
+  background: var(--dt-color-overlay);
```
kevinrodrigues, line 65: Will revert this change as it doesn't quite work. At the moment the only way this works is by setting the rgb directly. The style also works if we move it inside of the main selector so will move it and point it to the rgb color we use if we can't get this to work correctly.

packages/components/pie-modal/src/modal.scss
ashleynolan: I've just pulled down your PR and had a play around. One thing you can do (which saves the extra definition) is:

```
// Add this to the top of the modal.scss
@use '@justeat/pie-design-tokens/dist/jet.scss' as dt;

// and then for the definition…
--color-backdrop: #{dt.$color-overlay};
```

So that'll use sass to add the variable into the backdrop context in the web component. I tried all sorts of other ways of doing it just with native CSS, but I couldn't find a way of doing it. For some reason, `::backdrop` won't recognise the CSS variable unless it's set to a string inside the web component style context. I'm guessing it's to protect against a niche use case with shadow dom 🤷

kevinrodrigues: Cheers both appreciate you looking at it 👍

packages/components/pie-modal/src/index.ts
```diff
@@ -44,20 +44,57 @@ export class PieModal extends RtlMixin(LitElement) {
...
+    /**
+     * We require this because toggling the prop `isOpen` itself won't
+     * allow the dialog to open in the correct way (with the default background),
+     * the method `showModal()` needs to be invoked.
+     *
+     * https://lit.dev/docs/components/lifecycle/#willupdate
+     *
+     * @param changedProperties
+     */
+    willUpdate (changedProperties: Map<string, any>) {
```
jamieomaguire, line 67: my understand is `willUpdate` is to prep data ahead of an update. `updated` is for performing actions during/after. is `willUpdate` definitely right here?

kevinrodrigues: The event listener that gets attached needs to be prepped before the update otherwise there is an error. You can't really use the connected or disconnected methods on modal element as you will get an error there too. This was something i was going to ask the lit team about until i read about willUpdate.

kevinrodrigues: Screenshot 2023-06-22 at 16 36 53

Using `update`

kevinrodrigues: Because you are attaching it to a document not the dom shadow element so you wont get that error.

packages/components/pie-modal/src/index.ts
```diff
@@ -44,20 +44,57 @@ export class PieModal extends RtlMixin(LitElement) {
...
         this._onDialogClose();
     };
 
+    /**
+     * Dismiss modal via `_handleCloseDialog()` on backdrop click
+     * and will proceed to fire an `ON_MODAL_CLOSE_EVENT` event.
+     *
+     */
+    _handleDialogLightDismiss = () => {
+        this._dialog.addEventListener('click', event => {
```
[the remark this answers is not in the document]
kevinrodrigues: Where would you like it added? If you are thinking the disconnected / connected calls then there will be errors there.

packages/components/pie-modal/src/modal.scss
```diff
@@ -21,6 +23,10 @@
 
     box-shadow: var(--modal-elevation);
 
+    &::backdrop {
+        background: #{dt.$color-overlay};
```
ashleynolan, line 27: I'd potentially put a comment here about why we're using Sass variables for backdrop.

kevinrodrigues: Fixed.

packages/components/pie-modal/src/index.ts
```diff
@@ -44,20 +44,59 @@ export class PieModal extends RtlMixin(LitElement) {
...
+    /**
+     * We require this because toggling the prop `isOpen` itself won't
+     * allow the dialog to open in the correct way (with the default background),
+     * the method `showModal()` needs to be invoked.
+     *
+     * https://lit.dev/docs/components/lifecycle/#willupdate
+     *
+     * @param changedProperties
+     */
+    willUpdate (changedProperties: Map<string, any>) {
```
jamieomaguire, line 67: as a rule of thumb we should avoid using `any` in our code. [My pr](https://github.com/justeattakeaway/pie/blob/32c9184d60a517e7756ad64697dd9780f9a56c08/packages/components/pie-modal/src/index.ts#L34) proposes a way to avoid using `any` in this particular situation. You're welcome to add that to this PR instead if you want to get this in first?

kevinrodrigues: Yes I did think about that but wasn't sure as i took this directly from LIT. Will have a look 👍

kevinrodrigues: I've updated it so there would be a type checking now.

ashleynolan (approved):

FayeCarter (dismissed):

packages/components/pie-modal/src/index.ts
```diff
@@ -44,20 +44,62 @@ export class PieModal extends RtlMixin(LitElement) {
...
         this._onDialogClose();
     };
 
+    /**
+     * Dismiss modal via `_handleCloseDialog()` on backdrop click
+     * and will proceed to fire an `ON_MODAL_CLOSE_EVENT` event.
+     *
+     */
+    _handleDialogLightDismiss = () => {
+        this._dialog?.addEventListener('click', (event) => {
```
xander-marjoram, line 91: Will this add a new click handler every time the modal updates?

kevinrodrigues: Good point.

jamieomaguire: piggy backing this, do we need to consider removing any event listeners on the component being destroyed? I assume if the listener is on said component then perhaps not?

packages/components/pie-modal/src/modal.scss
jamieomaguire: I'd probably expand on this comment. It's a bit confusing because we know we can declare variables with the scope of `::backdrop`. Is this caused by a top layer/shadow DOM issue? 

Just mentioning it because at first glance I'm not sure this comment makes sense to me

packages/components/pie-modal/src/index.ts
kevinrodrigues: I'm still a bit uncertain on how lit deals with events if i'm honest. Their documentation suggests different ways to add them and some dont actually say you need to remove it either. They only mention removing listeners via the connected / discounted hooks and there is also a section about passing null into the event to remove the listener itself. I am fairly certainwe dont need to remove them.

`No need to remove internal event listeners. You don't need to remove event listeners added on the component's own DOM—including those added declaratively in your template. Unlike external event listeners, these won't prevent the component from being garbage collected.`

ashleynolan (dismissed):

jamieomaguire (approved):

packages/components/pie-modal/src/index.ts
jamieomaguire: Yeah I think in this instance you're all good!

xander-marjoram: I think there may still be an issue with adding too many event listeners but I don't mind if you want to test that post-merge

kevinrodrigues: I think they get removed but will have a play locally once merged. From what I see Lit destroys them for you.

jamieomaguire: I would move the event listener added to the constructor as outlined in [Lit's docs](https://lit.dev/docs/components/events/#adding-event-listeners-to-the-component-or-its-shadow-root)

packages/components/pie-modal/src/index.ts
```diff
@@ -44,20 +44,62 @@ export class PieModal extends RtlMixin(LitElement) {
...
+            const rect = this._dialog?.getBoundingClientRect();
+
+            if (typeof rect !== 'undefined') {
+                const isWithinDialogArea = (rect.top <= event.clientY && event.clientY <= rect.top + rect.height &&
```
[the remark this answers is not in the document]
kevinrodrigues: fixed.

packages/components/pie-modal/src/index.ts
kevinrodrigues: Moved to constructor.

xander-marjoram (approved):

ashleynolan (approved):

PR #564, opened 2023-06-22
feat(pie-modal): dsw-727 - implement scroll-locking for open modal

jamieomaguire:
---
---

[Added] - Scroll locking to modal

---
"pie-storybook": minor
---

[Added] - pie-modal story for scroll-locking and toggle button to open/close modal in stories

Screenshot 2023-06-22 at 17 06 16
Screenshot 2023-06-22 at 17 06 12


packages/components/pie-button/src/defs.ts
```diff
@@ -1,58 +1,11 @@
+export const sizes = ['xsmall', 'small-expressive', 'small-productive', 'medium', 'large'] as const;
```
xander-marjoram, line 1: 🚀

apps/examples/wc-next13/pages/index.tsx
```diff
@@ -12,8 +12,8 @@ export default function Home () {
     const decrement = () => setCount(count - 1);
     const switchVariant = () => {
         variantIndex.current += 1;
-        const variantsKeys: string[] = buttonVariants;
-        const variant: ButtonProps['variant'] = buttonVariants[variantIndex.current % variantsKeys.length];
+        const variantsKeys: readonly string[] = sizes;
```
xander-marjoram, line 15:
```suggestion
        const variantsKeys: readonly string[] = variants;
```

apps/examples/wc-angular12/src/app/app.component.ts
```diff
@@ -1,5 +1,5 @@
 import { Component } from '@angular/core';
-import { ButtonProps, buttonSizes, buttonVariants } from '@justeattakeaway/pie-button';
+import { ButtonProps, sizes, variants } from '@justeattakeaway/pie-button';
```
ashleynolan, line 2: Could this potentially get confusing if we end up importing other components with sizes and variants (for example, `pie-icon-button` has the same variables)?

ashleynolan: Just a small note as well – I'd recommend keeping refactoring like this in a separate PR. It makes it a bit harder to see what code is specifically to do with the scroll-locking.

packages/components/pie-button/README.md
```diff
@@ -65,13 +65,13 @@ import { PieButton } from '@justeattakeaway/pie-button/dist/react';
...
-| size        | `String`  | `medium`        | Size of the button, one of `buttonSizes` – `xsmall`, `small-expressive`, `small-productive`, `medium`, `large` |
-| type        | `String`  | `submit`        | Type of the button, one of `buttonTypes` – `submit`, `button`, `reset`, `menu`                                 |
-| variant     | `String`  | `primary`       | Variant of the button, one of `buttonVariants`– `primary`, `secondary`, `outline`, `ghost`                     |
-| disabled    | `Boolean` | `false`         | If `true`, disables the button.                                                                                |
-| isFullWidth | `Boolean` | `false`         | If `true`, sets the button width to 100% of it's container.                                                    |
+| Property    | Type      | Default         | Description                                                          |
+|-------------|-----------|-----------------|----------------------------------------------------------------------|
+| size        | `String`  | `medium`        | Size of the button, one of `sizes` |
```
ashleynolan, line 70: Would be good to keep the different sizes/types/variants in the README, so that consumers don't have to dig into the code to see the different options.

packages/components/pie-button/src/defs.ts
```diff
@@ -1,58 +1,11 @@
...
+export const types = ['submit', 'button', 'reset', 'menu'] as const;
+export const variants = ['primary', 'secondary', 'outline', 'ghost'] as const;
+
 export interface ButtonProps {
-    /**
```
ashleynolan, line 2: I kind of think we should include JSDoc comments like this as standard in these `defs`.

Completely understand that these are fairly self explanatory examples for these props, but we will have more complex props that these comments will become more useful for.

I think they either should live here, or in the `index.ts` file where they are used (happy for them to live inside the actual component potentially).

ashleynolan (commented):
Left a few comments – would it be feasible to split out the scroll-locking code from the refactor code?

packages/components/pie-modal/src/index.ts
```diff
@@ -24,7 +29,49 @@ export class PieModal extends RtlMixin(LitElement) {
         headingLevel: ModalProps['headingLevel'] = 'h2';
 
     @query('dialog')
-        _dialog: HTMLDialogElement;
+        _dialog!: HTMLDialogElement;
```
kevinrodrigues, line 32: Not a change required here but I've changed this in my PR to a `?` which will add a union type to allow undefined to hang off `_dialog`.

packages/components/pie-modal/src/index.ts
```diff
@@ -24,7 +29,49 @@ export class PieModal extends RtlMixin(LitElement) {
...
+            }
+        }
+    }
+
+    private _handleModalStateChanged (changedProperties: DependentMap<ModalProps>) {
+        if (changedProperties.has('isOpen')) {
+            const previousValue = changedProperties.get('isOpen') as boolean;
+            if (previousValue) {
+                this._dispatchModalCloseEvent();
+            } else if (previousValue === false) {
```
kevinrodrigues, line 70: Do you need to specifically check false here? Should be able to just remove the `if` and leave in the else.

packages/components/pie-webc-core/src/decorators/test/required-property.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+    });
+
+    afterEach(() => {
+        vi.restoreAllMocks();
+    });
+
+    // Mock class to test the decorator
+    class MockComponent {
+        @requiredProperty('mock-component')
+            color!: string;
```
kevinrodrigues, line 27: Will changing this to `color?` mean that the ts comments below can be removed?

apps/examples/wc-angular12/src/app/app.component.ts
jamieomaguire: I'll definitely split in future but going to keep this PR for now if that's ok.

I don't really think itis confusing, I actually think they should be named similarly. I see it more as consistency. You'd import from different namespaces and probs end up using aliases if you had to but I don't think we should be doing that preemptively. I see it as more of an external usage concern rather than internal one.

packages/components/pie-button/src/defs.ts
jamieomaguire: I think we should use them for the less obvious ones, but having comments like `the size of the button.` isn't particularly helpful in my opinion.

packages/components/pie-modal/src/index.ts
jamieomaguire: depending on who merges first I'll take your approach if it works

packages/components/pie-modal/src/index.ts
```diff
@@ -24,7 +29,49 @@ export class PieModal extends RtlMixin(LitElement) {
         headingLevel: ModalProps['headingLevel'] = 'h2';
 
     @query('dialog')
-        _dialog: HTMLDialogElement;
+        _dialog!: HTMLDialogElement;
+
+    firstUpdated (changedProperties: DependentMap<ModalProps>) {
```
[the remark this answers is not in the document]
jamieomaguire: This was a suggestion from Skip which I agree with. It adds better type checking. `unknown` isn't ideal and is a bit of a last resort when we really don't want to use `any`. 

It is saying the value of `changedProperties` is a map that is filled with `ModalProp` properties

packages/components/pie-modal/src/index.ts
jamieomaguire: it can also be `undefined`, in which case we don't want to dispatch anything

packages/components/pie-webc-core/src/decorators/test/required-property.spec.ts
jamieomaguire: It does! Thank you!

packages/components/pie-button/src/defs.ts
ashleynolan: It's more a consistency thing – it's better to have them on all props/definitions than only on a couple.

For example, the difference between type and variant may be completely obvious to us, as we work with this code a lot, but to someone new to the code, the extra comments clear it up straight away.

apps/examples/wc-angular12/src/app/app.component.ts
xander-marjoram: Regarding your first question @ashleynolan, I think that's a valid point, but in the scenario where someone needs to directly work with the different sizes and variants for multiple components they should be willing to rename the variables on import them. I don't think we need to make the names longer or more complex to cater for that.

```javascript
import { ButtonProps, sizes as buttonSizes, variants as buttonVariants } from '@justeattakeaway/pie-button';
import { IconButtonProps, sizes as iconButtonSizes, variants as iconButtonVariants } from '@justeattakeaway/pie-button';
```

packages/components/pie-modal/src/index.ts
xander-marjoram: How does `as boolean` behave when it could be `undefined`?
It could make sense to do an early `return` if it's `undefined`

packages/components/pie-modal/src/index.ts
```diff
@@ -44,7 +91,7 @@ export class PieModal extends RtlMixin(LitElement) {
                         variant="ghost-tertiary"
                         class="c-modal-closeBtn"></pie-icon-button>
                 </header>
-                <article>
+                <article >
```
xander-marjoram, line 94: Nit: whitespace

packages/components/pie-button/src/defs.ts
xander-marjoram: I partially agree, however having a JSDoc comment that only says "the variant of the button" doesn't exactly clear up much confusion 😄

jamieomaguire: Exactly :D

apps/examples/wc-angular12/src/app/app.component.ts
jamieomaguire: Yeah that's how I envisioned it too :)

packages/components/pie-modal/src/index.ts
jamieomaguire: I can't imagine `as boolean` doing us any favours, but an early return sounds like a much better way to do this. Far more explicit!

jamieomaguire: @xander-marjoram even though its cast it seems like it can be undefined 😢

jamieomaguire: Screenshot 2023-06-27 at 11 19 47

jamieomaguire: Screenshot 2023-06-27 at 11 22 19
All initial `changedProperties` values on first render

kevinrodrigues: Sorry further question because i think we may not even need the `as Boolean` type assertion. The `changedProperties.has()` already has an assertion that a boolean (primitive: true, false) will be returned so I think you just need to do a check against that without asserting again on the `has` method, effectively twice. That way i think it removes the `boolean | undefined` and you wont be left with making further checks later in the code.

You could just stick with the below then in the if statement continue as normal without checking anything else.
` const isClosed = changedProperties.get('isOpen') === false;`

packages/components/pie-modal/src/index.ts
jamieomaguire: SMH

jamieomaguire:
I think there's some work to do around the visual test. I'd like to address this, and the currently skipped component tests in an immediate follow-up PR if that's ok

packages/components/pie-button/src/defs.ts
jamieomaguire: @ashleynolan can I raise a separate PR to sort out the comments? I see no benefit in leaving them as they were as they weren't very helpful. I could add more descriptive ones in a dedicated PR if that's ok

package.json
```diff
@@ -72,6 +72,7 @@
     "allure-commandline": "2.21.0",
     "autoprefixer": "10.4.14",
     "babel-loader": "8.3.0",
+    "body-scroll-lock": "4.0.0-beta.0",
```
kevinrodrigues, line 75: Is there a reason to use the beta version?

packages/components/pie-modal/src/defs.ts
```diff
@@ -1,28 +1,21 @@
...
 }
 
 /**
- * Modal heading levels/tags
+ * Event name for when the modal is closed.
+ *
+ * @constant
  */
-export const headingLevels: Array<ModalProps['headingLevel']> = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6'];
+export const ON_MODAL_CLOSE_EVENT = 'pie-modal-close';
```
kevinrodrigues, line 14: Not for this PR but we should probably consider using the form events once we complete that ticket. If they work just as good we could remove these i think.

packages/components/pie-modal/src/index.ts
```diff
@@ -56,39 +122,19 @@ export class PieModal extends RtlMixin(LitElement) {
...
-            if (isClosed) {
-                dialog.showModal();
-            } else {
-                dialog.close();
-            }
-        }
-    }
-
-    _handleCloseDialog = () => {
+    private _handleCloseDialog = () : void => {
```
kevinrodrigues, line 125: There's two functions here that kinda do the same thing almost (one here and one on line 55), could we merge them so there's one `handleCloseDialog` I think we could move the enable scrolling or keep it in place as long as it's not being called unnecessarily. I would be tempted to delete this function and reuse the one that is on line 55 as they should handle the dispatch event too

packages/components/pie-modal/src/index.ts
```diff
@@ -31,17 +36,78 @@ export class PieModal extends RtlMixin(LitElement) {
...
+            if (previousValue === undefined && this.isOpen) {
+                this._dispatchModalOpenEvent();
+            }
+        }
+    }
+
+    // Handles changes to the modal isOpen property by dispatching any appropriate events
+    private _handleModalOpenStateChanged (changedProperties: DependentMap<ModalProps>) : void {
+        if (changedProperties.has('isOpen')) {
+            const previousValue = changedProperties.get('isOpen') as boolean;
```
kevinrodrigues, line 87: I still don't think we need to assert here but happy for you to leave it in. The method of the get will automatically return a boolean value.

packages/components/pie-modal/src/defs.ts
jamieomaguire: Yup one for the future I reckon :)

packages/components/pie-modal/src/index.ts
jamieomaguire: It won't automatically return a boolean, it will be bool or undefined. However the `as boolean` isn't needed and I'm pushing an update to remove that now :)

kevinrodrigues (dismissed):

packages/components/pie-modal/src/index.ts
```diff
@@ -31,21 +36,91 @@ export class PieModal extends RtlMixin(LitElement) {
...
+            this._dispatchModalOpenEvent();
+        }
+    }
+
+    // Handles changes to the modal isOpen property by dispatching any appropriate events
+    private _handleModalOpenStateChanged (changedProperties: DependentMap<ModalProps>) : void {
+        const previousValue = changedProperties.get('isOpen');
+
+        if (previousValue !== undefined) {
+            if (previousValue as boolean) {
```
xander-marjoram, line 102: Is the `as boolean` necessary here? I would have thought the undefined check above should leave a boolean as the only other possibility.

package.json
jamieomaguire: Just the latest version available. I notice the package hasn't been updated in a while so may be worth us finding an alternative in the future but we use this already in fozzie with a lot of success :)

packages/components/pie-modal/src/index.ts
```diff
@@ -57,60 +132,59 @@ export class PieModal extends RtlMixin(LitElement) {
     }
 
     /**
-     * We require this because toggling the prop `isOpen` itself won't
-     * allow the dialog to open in the correct way (with the default background),
-     * the method `showModal()` needs to be invoked.
-     *
-     * https://lit.dev/docs/components/lifecycle/#willupdate
+     * Dismiss modal via `_handleCloseDialog()` on backdrop click
```
xander-marjoram, line 135: I think this comment needs updating now?

packages/components/pie-modal/src/index.ts
jamieomaguire: Logically yes, but not for intellisense. I'm happy to remove it though

packages/components/pie-modal/src/index.ts
jamieomaguire: good spot

packages/components/pie-modal/src/index.ts
jamieomaguire: good spot - will remove

packages/components/pie-modal/src/index.ts
jamieomaguire: great spot - updated :)

packages/components/pie-webc-core/src/decorators/test/valid-property-values.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+        requestUpdate (propertyKey: string, oldValue: unknown) {
+            this._requestUpdateArgs = { propertyKey, oldValue };
+        }
+
+        requestUpdateCalledWith () {
+            return this._requestUpdateArgs;
+        }
+    }
+
+    it('should not change the value if it is valid', () => {
```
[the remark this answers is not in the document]
jamieomaguire: Oops, yes deffo - updating now

kevinrodrigues (approved):

packages/components/pie-webc-core/src/decorators/test/required-property.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+} from 'vitest';
+
+import { requiredProperty } from '../required-property';
+
+describe('requiredProperty', () => {
+    let consoleErrorSpy: unknown;
+
+    beforeEach(() => {
+        // eslint-disable-next-line @typescript-eslint/no-empty-function
+        consoleErrorSpy = vi.spyOn(console, 'error').mockImplementation(() => {});
```
xander-marjoram, line 17: For future reference you should be able to use `vi.fn()` instead of an empty arrow function which would avoid the need to disable eslint. Happy for you to try that in your next PR

xander-marjoram (approved):

packages/components/pie-webc-core/src/decorators/test/required-property.spec.ts
jamieomaguire: Screenshot 2023-06-28 at 13 19 37

I was hoping so too but I encountered this bad boy when trying, which led me to the sad code you now review

packages/components/pie-webc-core/src/decorators/test/valid-property-values.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+    it('should fallback to the default value if an invalid value is assigned', () => {
+        // Arrange
+        const mockComponent = new MockComponent();
+
+        // Act
+        mockComponent.color = 'yellow';
+
+        // Assert
+        expect(mockComponent.color).toBe('red');
+        expect(consoleErrorSpy).toHaveBeenCalled();
```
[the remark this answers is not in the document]
jamieomaguire: cheers - will do :)

packages/components/pie-webc-core/src/decorators/test/required-property.spec.ts
kevinrodrigues: You should be able to call mock implet... without anything i.e vi.spyOn(console, 'error').mockImplementation(). Usually that is fine. Or you can specifically return an anonymous method instead.  Or failing that you can just spy on the thing without invoking it that should work too `vi.spyOn(console, 'error')`

xander-marjoram: I think the only problem with `spyOn` without a `mockImplementation` is you still get the error message logging to the console which isn't so nice when you're running the tests locally. 

Your first suggestion would be ideal, however.

packages/components/pie-button/src/defs.ts
ashleynolan: Yes please 👍

packages/components/pie-modal/src/index.ts
```diff
@@ -57,60 +132,59 @@ export class PieModal extends RtlMixin(LitElement) {
...
      */
-    private _handleDialogLightDismiss = (event: MouseEvent) => {
-        const rect = this._dialog?.getBoundingClientRect();
-
-        if (typeof rect !== 'undefined') {
-            const isClickOutsideDialog = event.clientY < rect.top ||
-                event.clientY > rect.bottom ||
-                event.clientX < rect.left ||
-                event.clientX > rect.right;
+    private _dispatchModalCloseEvent = () : void => {
```
ashleynolan, line 167: Super late to this game (sorry, I missed you were looking for re-reviews).

Would be good at some point to have a `dispatchEvent` helper, to save with the repetition when dispatching events.

Not sure if we should just do this as part of the core helpers potentially as it'll be a cross-component thing I imagine?

jamieomaguire: yep fully agree with this

kevinrodrigues: We should be able to completely get rid of some of these dispatching event calls, im planning on cleaning some up in my PR. With regards to sharing I think we should move to a helper file which will be nice like you said.

PR #568, opened 2023-06-23
Fix modal test

jamieomaguire:
---
---

[Changed] - Bundle pie-icon-button with the modal and fix component test

### DIFF
--- a/packages/components/pie-modal/src/index.ts
+++ b/packages/components/pie-modal/src/index.ts
@@ -1,6 +1,6 @@
 import { LitElement, unsafeCSS } from 'lit';
 import { html, unsafeStatic } from 'lit/static-html.js';
-import { property, query } from 'lit/decorators.js'; // eslint-disable-line import/no-extraneous-dependencies
+import { property, query } from 'lit/decorators.js';
 import { RtlMixin, validPropertyValues, requiredProperty } from '@justeattakeaway/pie-webc-core';
 import '@justeattakeaway/pie-icon-button';
 import styles from './modal.scss?inline';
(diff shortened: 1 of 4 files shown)


JoshuaNg2332 (dismissed):

packages/components/pie-modal/src/index.ts
```diff
@@ -1,8 +1,8 @@
 import { LitElement, unsafeCSS } from 'lit';
 import { html, unsafeStatic } from 'lit/static-html.js';
-import { property, query } from 'lit/decorators.js'; // eslint-disable-line import/no-extraneous-dependencies
+import { property, query } from 'lit/decorators.js';
 import { RtlMixin, validPropertyValues, requiredProperty } from '@justeattakeaway/pie-webc-core';
-
+import '@justeattakeaway/pie-icon-button';
```
ashleynolan, line 5: Are we sure this is a good idea? I'd imagine this will lead to a lot of bloat in our components if we bundle smaller components into bigger components.

Don't mind this being a potentially short-term thing, but feel like we need to look at the performance impacts of doing it

jamieomaguire: I think that should feed into our workshop next week. For now I think we should do it. Spectrum do the same thing if that helps: https://github.com/adobe/spectrum-web-components/blob/60f54fb2c5f656cc775d52637e1ef95bde5949f5/packages/dialog/src/Dialog.ts#L29C41-L29C56

jamieomaguire:
Will investigate the module resolution error on Tuesday 👍

jamieomaguire:
Going to close this PR for now until we agree bundle strategies next week 👍

packages/components/pie-modal/src/index.ts
kevinrodrigues: This seems like the correct fix in my head because the test needs some context to the elements. I'm happy to pair up on working out the bundling issue too @jamieomaguire

PR #574, opened 2023-06-27
refactor(pie-icons-configs): renames iconSize to size

dandel10n:
BREAKING CHANGE:
"iconSize" prop now is "size" prop


raoufswe (approved):

packages/tools/pie-icons-configs/configs-vue.js
```diff
@@ -6,7 +6,7 @@ import {
 } from './configs.js'; // eslint-disable-line import/extensions
 
 /**
- * Object with iconSize props configuration for large and regular icons
+ * Object with size props configuration for large and regular icons
  */
 export const iconSize = {
```
ashleynolan, line 11: Should this be updated to `size`?

dandel10n: I think not as this const is imported in vue package and makes sense to leave it as `iconSize` as more descriptive.

jamieomaguire (approved):

PR #576, opened 2023-06-27
feat(pie-icons-webc): new icon package using web components

LTurns:
[Added] - package for creating and building icons using web components


packages/tools/pie-icons-configs/package.json
```diff
@@ -7,7 +7,8 @@
   "exports": {
     ".": "./configs.js",
     "./configs-react": "./configs-react.js",
-    "./configs-vue": "./configs-vue.js"
+    "./configs-vue": "./configs-vue.js",
+    "./configs": "./configs.js"
```
sebastienrocks, line 11: Do we need to export this twice? I assume `./configs.js` is already getting exported on line 8?

packages/tools/pie-icons-webc/README.md
```diff
@@ -0,0 +1,113 @@
...
+Within the context of a Vue app, that will look like:
+
+  ```js
+  <template>
+      <icon-calendar />
+      <icon-app-restaurant size="xs"/>
+  </template>
+
+  <script>
+  import { IconAlcoholFilled } from '@justeattakeaway/pie-icons-webc';
```
sebastienrocks, line 39: I think this should be: `{ IconCalendar, IconAppRestaurant }`

packages/tools/pie-icons-webc/icons/IconAlcohol.js
```diff
@@ -0,0 +1,51 @@
...
+
+export default class IconAlcohol extends HTMLElement {
```
sebastienrocks, line 7: It seems like the main difference between most of these 'components' is just the class name and SVG definition.

Is there anyway to improve this and maybe pass in these 2 things to generate the components instead? Just concerned with the repeated code across all of these icons if ever we want to make changes

packages/tools/pie-icons-webc/README.md
```diff
@@ -0,0 +1,113 @@
...
+export const IconAlcohol = createComponent({
+    displayName: 'IconAlcohol',
+    elementClass: IconAppRestaurant,
+    react: React,
+    tagName: 'icon-app-restaurant',
+    events: { },
+});
+
+  <template>
+      <IconAlcohol size="l"/>
```
jamieomaguire, line 59: I think this should be IconAppRestaurant

packages/tools/pie-icons-webc/build.js
```diff
@@ -0,0 +1,114 @@
...
+const template = document.createElement('template');
+template.innerHTML = '${svg}';
+
+export default class ${name} extends HTMLElement {
+    constructor () {
+        super();
+        const clone = template.content.cloneNode(true);
+
+        this.root = this.attachShadow({ mode: 'open' });
+        this.root.append(clone);
```
jamieomaguire, line 23: How come we aren't using Lit?

packages/tools/pie-icons-webc/README.md
LTurns: ah so should be IconAlcohol because I'm referencing the wrapper on line 50 in the template 😄

packages/tools/pie-icons-webc/README.md
LTurns: whoops thank you!

packages/tools/pie-icons-webc/README.md
```diff
@@ -0,0 +1,113 @@
...
+### Installation
+
+Add the module to your project
+
+```bash
+yarn add @justeattakeaway/pie-icons-webc
+```
+
+### Applications
+
```
ashleynolan, line 24: Could you add an example of how an icon can be pulled into a Lit Web Component as well?

packages/tools/pie-icons-webc/README.md
```diff
@@ -0,0 +1,113 @@
...
+The component extends [@justeat/browserslist-config-fozzie](https://github.com/justeat/browserslist-config-fozzie) package for the list of browsers to support.
+
+
+## Contributing
+
+Before starting please read our [contributing guide](https://pie.design/engineers/contributing/)
+
+### Adding new icons
+
+Icons should be added as SVGs to the main pie-icons package and published, before simply incrementing the dependency of `pie-icons` in the `pie-icons-vue` package, to generate the new set of Vue components.
```
ashleynolan, line 102:
```suggestion
Icons should be added as SVGs to the main pie-icons package and published, before simply incrementing the dependency of `pie-icons` in the `pie-icons-webc` package, to generate the new set of Web Components.
```

packages/tools/pie-icons-webc/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+The PIE iconset is managed by our PIE design team and new icon requests should go through them to ensure that they are designed inline with our standards and guildelines. Please reach out to PIE design system team using #help-designsystem slack channel.
```
ashleynolan, line 104:
```suggestion
The PIE iconset is managed by our PIE design team and new icon requests should go through them to ensure that they are designed in-line with our standards and guildelines. Please reach out to the PIE design system team using #help-designsystem slack channel.
```

packages/tools/pie-icons-webc/package.json
```diff
@@ -0,0 +1,63 @@
+{
+  "name": "@justeattakeaway/pie-icons-webc",
+  "version": "0.1.0",
```
ashleynolan, line 3: You'll need to set this to be `0.0.0` for the first release, as changesets will then bump it on merge to `v0.1.0`

packages/tools/pie-icons-webc/helpers/index.js
```diff
@@ -0,0 +1,5 @@
+const removeHyphenBeforeDigits = (name) => name.replace(/\-(\d+)/, '$1'); // eslint-disable-line no-useless-escape
```
ashleynolan, line 1: Doesn't need to be part of this PR, but would be good to tidy this up so that both the vue and webc packages just use the same function for this (if that's possible using the `pie-icons-configs` package)

packages/tools/pie-icons-configs/configs.js
```diff
@@ -70,7 +70,7 @@ export function validateGetRegularIconSize (sizeValue) {
  */
 
 export const getSvgProps = (svgClasses, staticClasses, sizeValue, componentName) => {
-    const isLargeIcon = svgClasses.endsWith('Large') || svgClasses.endsWith('-large');
+    const isLargeIcon = svgClasses.endsWith('Large') || svgClasses.endsWith('-large') || componentName.endsWith('Large');
```
ashleynolan, line 73: Do the outputted SVGs for the web component have classnames on them? I'd probably avoid using the componentName, simple as there's an element of user error built in (for instance, we found that some names had lowercase `l`). If you use the generated classnames to check for if the icon is large or not, these have already been normalised against error like that.

packages/tools/pie-icons-webc/build.js
ashleynolan: Nothing to add from me – just gets rid of a dependency for simple WCs when we probably don't need it here. If we find it causes us issues down the line though we can always revisit

packages/tools/pie-icons-configs/package.json
LTurns: resolved this [here](https://github.com/justeattakeaway/pie/pull/585/files#diff-3fa9cef765d67886e2775ac19cd61a60dcf31185b36c776537bb43292344f565R8) - I've renamed the export from `.` to the filename so it's clearer for the import 👍 Closing this PR as lots of changes had gone through and it was less confusing to start from most recent main 😄

packages/tools/pie-icons-webc/icons/IconAlcohol.js
LTurns: yeah there might be a better way to do this - at the moment the `build.js` generates all of these so it's still relatively easy to make changes. But I agree it might be worth seeing if we could reduce repetition. I've followed the same steps as pie-icons-react and pie-icons-vue here so it might be something we revisit later for all icon packages to keep things consistent 😄

packages/tools/pie-icons-configs/configs.js
LTurns: that makes sense, thank you! I've updated this to use the class [here](https://github.com/justeattakeaway/pie/pull/585/files#diff-beb4382d1fd71394d71c0f0f07f2e5111af3e72f86d94ce723a1a03b48251af9R13) - Closing this current PR as lots of changes had gone through and it was less confusing to start from most recent main 👍

PR #578, opened 2023-06-28
docs(pie-monorepo): add descriptive jsdoc comments to WC props

jamieomaguire:
Add descriptive jsdoc comments to component props

### DIFF
--- a/packages/components/pie-modal/src/defs.ts
+++ b/packages/components/pie-modal/src/defs.ts
@@ -7,10 +7,12 @@ export interface ModalProps {
     heading: string;
     /**
      * The HTML heading tag to use for the modal's heading. Can be H1-H6.
+     * @default "h2"
      */
     headingLevel: typeof headingLevels[number];
     /**
      * When true, the modal will be open.
+     * @default false
      */
     isOpen: boolean;
 }
--- a/packages/components/pie-modal/src/defs.ts
+++ b/packages/components/pie-modal/src/defs.ts
@@ -1,8 +1,17 @@
 export const headingLevels = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6'] as const;
 
 export interface ModalProps {
+    /**
+     * The text to display in the modal's heading.
+     */
     heading: string;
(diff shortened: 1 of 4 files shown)


packages/components/pie-modal/src/defs.ts
```diff
@@ -1,8 +1,19 @@
 export const headingLevels = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6'] as const;
 
 export interface ModalProps {
+    /**
+     * The text to display in the modal's heading.
+     */
     heading: string;
+    /**
```
LTurns, line 8: super nit but you're using full stops in modal but not in button or icon button 😄

ashleynolan: @xander-marjoram would be proud

jamieomaguire: Ugh. Removing them from here now

packages/components/pie-modal/src/defs.ts
```diff
@@ -1,8 +1,19 @@
 export const headingLevels = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6'] as const;
 
 export interface ModalProps {
+    /**
+     * The text to display in the modal's heading.
```
ashleynolan, line 5: Is it worth noting which of these are required and optional? (I think there's a `@required` directive if so).

If not required, I'd explicitly say here that the heading is optional.

jamieomaguire: Conflicted about this. Seems super smelly to not leverage actual type systems to achieve this. Not sure how we go about it though. Can we look into that separately? I can't see a `@required` directive available

jamieomaguire: Sounds good, but do you mind pointing me to an example of this in the wild? I honestly can't seem to see how we could achieve this!

raoufswe (approved):

JoshuaNg2332 (approved):

PR #579, opened 2023-06-28
refactor(pie-webc-core): [Changed] - Use factory pattern for decorators

xander-marjoram:
I recommend hiding whitespace changes when reviewing to limit the (nesting) diffs in the decorator files.

Testing the waters to see if we like the look of this pattern or if it adds an unnecessary layer.

### Reviewer 1

### Reviewer 2

### DIFF
--- a/.changeset/friendly-carrots-remain.md
+++ b/.changeset/friendly-carrots-remain.md
@@ -1,0 +1,7 @@
+---
+"@justeattakeaway/pie-icon-button": minor
+"@justeattakeaway/pie-button": minor
+"@justeattakeaway/pie-modal": minor
+---
+
+[Changed] - Use new decorator factories
--- a/packages/components/pie-modal/src/index.ts
+++ b/packages/components/pie-modal/src/index.ts
@@ -16,9 +16,9 @@ import {
 // Valid values available to consumers
 export { type ModalProps, headingLevels };
 
-const componentName = 'pie-modal';
-const validPropertyValues = validPropertyValuesDecoratorFactory(componentName);
-const requiredProperty = requiredPropertyDecoratorFactory(componentName);
+const componentSelector = 'pie-modal';
+const validPropertyValues = validPropertyValuesDecoratorFactory(componentSelector);
+const requiredProperty = requiredPropertyDecoratorFactory(componentSelector);
 
 export class PieModal extends RtlMixin(LitElement) {
     @property({ type: Boolean })
@@ -196,10 +196,10 @@ export class PieModal extends RtlMixin(LitElement) {
     static styles = unsafeCSS(styles);
 }
 
-customElements.define(componentName, PieModal);
+customElements.define(componentSelector, PieModal);
 
 declare global {
     interface HTMLElementTagNameMap {
-        [componentName]: PieModal;
+        [componentSelector]: PieModal;
     }
 }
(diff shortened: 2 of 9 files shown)


jamieomaguire (dismissed):

packages/components/pie-modal/src/index.ts
```diff
@@ -15,17 +17,19 @@ import {
...
+const validPropertyValues = validPropertyValuesDecoratorFactory(componentSelector);
+const requiredProperty = requiredPropertyDecoratorFactory(componentSelector);
 
 export class PieModal extends RtlMixin(LitElement) {
     @property({ type: Boolean })
         isOpen = false;
 
     @property({ type: String })
-    @requiredProperty(componentSelector)
+    @requiredProperty()
```
jamieomaguire, line 28: NIT: what if we just named it `required`?

packages/components/pie-modal/src/index.ts
```diff
@@ -15,17 +17,19 @@ import {
...
         heading!: string;
 
     @property()
-    @validPropertyValues(componentSelector, headingLevels, 'h2')
+    @validPropertyValues(headingLevels, 'h2')
```
jamieomaguire, line 32: Similar here, what about `ValidValues` or `AllowedValues`?

.changeset/friendly-carrots-remain.md
```diff
@@ -0,0 +1,7 @@
+---
+"@justeattakeaway/pie-icon-button": minor
+"@justeattakeaway/pie-button": minor
+"@justeattakeaway/pie-modal": minor
+---
+
+[Changed] - Use new decorator factories
```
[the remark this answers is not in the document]
xander-marjoram: Interesting, I think we should have a chat about this as a team to make sure we all know what constitutes a version change. I was under the impression that every PR required a changeset.

packages/components/pie-modal/src/index.ts
xander-marjoram: I initially wondered whether there might be anything else we need to decorate as `required`, but I'm not sure that's the case. We can probably get away with renaming it, yeah.

packages/components/pie-modal/src/index.ts
xander-marjoram: Similar to the other comment, when I wrote this initially I wasn't sure whether there might be other things that we need to decorate but I don't think that's the case now.

xander-marjoram: Looking back we did have `validValues`, but there was a PR comment about potential confusion with its first parameter having the same name. I quite like `allowedValues`.

.changeset/friendly-carrots-remain.md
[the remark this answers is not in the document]
ashleynolan: The way I see it with version bumps, we should always version on code change, mainly so that if we do break something, it's easy to roll back.

The only things that don't require a version bump would be:
- Test only changes (so adding tests, refactoring tests – but no underlying component change)
- Superficial config updates – so any config changes that only affect our repo management, and don't affect the component itself. So, updating an `editorconfig` could be an example of that for instance.

In general, anything that could unintentionally introduce a bug in the component should be versioned, so refactoring should definitely be versioned.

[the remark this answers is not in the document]
jamieomaguire: Yeah that all sounds reasonable to me!

ashleynolan (commented):
I'm not completely sold that this adds that much value for the amount of extra code going in. It seems like it abstracts things a little more, but is also a little more confusing (imo).

Would be interested to know what the main benefits of using the approach might be incase I'm missing something?

xander-marjoram:
I would argue that it doesn't add that much code, it just wraps a function around the existing decorator so that we don't need to pass in the `componentSelector` every time it's used.

However, the main point of raising this PR was to get some opinions about what people thought about the refactor, so if you think it's adding unnecessary complexity that's a valid opinion.

PR #585, opened 2023-06-30
feat(pie-icons-webc): new icon package using web components

LTurns:
[Added] - pie-icons-webc package for turning SVGs into web components
[Added] - small additional logic in `configs.js` for supporting `pie-icons-webc` package

tested in icon-button & some vue and react example apps 👍 

https://github.com/justeattakeaway/pie/assets/49618712/d56a52e5-a0a8-4670-a376-aa6181503f97


packages/tools/pie-icons-webc/package.json
```diff
@@ -0,0 +1,63 @@
...
+        "kebab-case": "1.0.2",
+        "pascal-case": "3.1.2",
+        "rollup": "3.20.2",
+        "rollup-plugin-delete": "2.0.0"
+    },
+    "volta": {
+        "extends": "../../../package.json"
+    },
+    "dependencies": {
+        "@open-wc/testing": "^3.2.0",
```
FayeCarter, line 60: nit, but do we want the `^`?

packages/tools/pie-icons-webc/icons/IconAlcohol.js
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    set size (value) {
+        this.setAttribute('size', value);
+    }
+
+    connectedCallback () {
+        const svg = this.root.querySelector('svg');
+        const defaultSize = getDefaultIconSize('c-pieIcon c-pieIcon--alcohol');
+        svg.setAttribute('width', iconSize[defaultSize]);
```
ashleynolan, line 29: Quick check on what's set on the SVG when it renders? Do we also add the `c-pieIcon c-pieIcon--alcohol` classnames onto the element?

We currently add these directly to the SVG for the React/Vue icon libraries, but I think it probably makes more sense to add these to the web component element, as these are there for consumers to help thm position the icons (or apply other styling like that)

packages/tools/pie-icons-webc/helpers/index.js
```diff
@@ -0,0 +1,5 @@
+const removeHyphenBeforeDigits = (name) => name.replace(/\-(\d+)/, '$1'); // eslint-disable-line no-useless-escape
```
ashleynolan, line 1: /**
 * Removes any hyphens from before digits in a string
 * This is used used example where we generate icon names from keys, such as over-18-filled-large
 * and without this the PascalCase generation returns Over-16Filled
 * @param {string} name - a string, for example an iconKey
 * @returns {string}
 */

packages/tools/pie-icons-configs/package.json
```diff
@@ -5,7 +5,7 @@
   "type": "module",
   "description": "",
   "exports": {
-    ".": "./configs.js",
+    "./configs": "./configs.js",
```
ashleynolan, line 8: Does this change mean that you'd need to import `@justeattakeaway/pie-icons-configs/configs/` rather than just `@justeattakeaway/pie-icons-configs`?

I think it makes more sense as `.` (without the `/configs`) – mainly as not many package don't have a default export, but interested to hear other thoughts potentially.

packages/tools/pie-icons-webc/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+  ```js
+import { IconAppRestaurant } from '@justeattakeaway/pie-icons-webc';
+export { IconAppRestaurant }
+
+  export class IconButton extends LitElement {
+    render () {
+      return html`
+        <button>
+          <icon-alcohol-filled size="xs" />
```
ashleynolan, line 73:
```suggestion
          <icon-app-restaurant size="xs" />
```

packages/tools/pie-icons-webc/tests/icon.test.js
```diff
@@ -0,0 +1,15 @@
+import { html, fixture, expect } from '@open-wc/testing';
+import { IconAlcoholFilled } from '../icons/IconAlcoholFilled';
+
+describe('IconAlcoholFilled', () => {
+    it('has a default width and height for a regular sized icon', async () => {
```
ashleynolan, line 5: Would be good to have a test for large icons too?

ashleynolan (commented):
Looks good – couple more comments but otherwise LGTM

packages/tools/pie-icons-configs/package.json
jamieomaguire: Might be worth repeating what we do in the [eslint config package](https://github.com/justeattakeaway/pie/blob/3865b0026261d4fe6db836f7dc57c85aa13e4c76/packages/tools/eslint-config-pie/package.json#L5C3-L9C5)

packages/tools/pie-icons-webc/helpers/index.js
dandel10n: Dow we have any icons with the digits in the name?

dandel10n: Interesting why it is not an issue for React and Vue packages that also use `pascalCase`

dandel10n: I checked that in React library 
```
const handleComponentName = (name) => name.replace(/\-(\d+)/, '$1');
pascalCase(handleComponentName(iconKey))
``` 
 does the trick.

Maybe worth updating there as well for consistency

LTurns: oh nice! Vue also uses the helper function so I'll update both so they're consistent with react 👍

packages/tools/pie-icons-configs/package.json
LTurns: ah yep I was getting an error earlier to do with the export so changed it to `./configs`, but @dandel10n helped me solve this this morning so I'll revert the change 👍

packages/tools/pie-icons-webc/build.js
```diff
@@ -0,0 +1,110 @@
...
+        this.setAttribute('size', value);
+    }
+
+    connectedCallback () {
+        const svg = this.root.querySelector('svg');
+        const defaultSize = getDefaultIconSize('${svgClasses}');
+        svg.setAttribute('width', iconSize[defaultSize]);
+        svg.setAttribute('height', iconSize[defaultSize]);
+        this.root.append(svg);
+    }
```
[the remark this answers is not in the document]
LTurns: no visible delay - tested in apps and pie-icon-button and the icons appear immediately with the buttons 👍

packages/tools/pie-icons-webc/icons/IconAlcohol.js
LTurns: I've added the classname to the element so they can use it for styling. If they add a different class though currently it overrides the default - do we want to append it instead? Can take more of a look on Tuesday if so, but I think everything else is resolved 😄

packages/tools/pie-icons-webc/helpers/index.js
ashleynolan: Would it not be worth changing the React one to use the helper? That way we don't have the same regex repeated 3 times across the libraries?

@dandel10n maybe let's chat about Monday? I'm happy to help make this change in the PR so we can help get this through.

packages/tools/pie-icons-vue/helpers/index.js
```diff
@@ -1,5 +0,0 @@
-const removeHyphenBeforeDigits = (name) => name.replace(/\-(\d+)/, '$1'); // eslint-disable-line no-useless-escape
```
ashleynolan, line 1: I would personally keep this and use it to avoid duplication of the regex across the 3 libraries. Just keeps it a bit drier.

ashleynolan: I've been playing around in a local branch with this – I'll raise a PR once this is merged so this is shared

packages/tools/pie-icons-webc/icons/IconAlcohol.js
ashleynolan: Yeh would be good if it was appended – happy for this to happen as a separate PR after this is merged though 👍

ashleynolan (approved):

kevinrodrigues (approved):

packages/tools/pie-icons-webc/build.js
```diff
@@ -0,0 +1,113 @@
...
+        this.setAttribute('class', '${svgClasses}');
+        this.root.append(svg);
+    }
+
+    attributeChangedCallback (attr, oldVal, newVal) {
+        const svg = this.root.querySelector('svg');
+        let svgSize;
+
+        if (attr === 'size') {
+            svgSize = getSvgProps('${svgClasses}', '', newVal, '${name}');
```
dandel10n, line 59: I see that second parameter here passed as an empty string meaning that no `staticClasses` will be applied to the component. Maybe in web components classes are appended by default and it is not needed but worth checking.

packages/tools/pie-wrapper-react/custom-elements.json
```diff
@@ -84,9 +84,9 @@
                     "kind": "variable",
                     "name": "sizes",
                     "type": {
-                        "text": "['xsmall', 'small-expressive', 'small-productive', 'medium', 'large']"
+                        "text": "['xsmall', 'small-productive', 'small-expressive', 'medium', 'large']"
```
dandel10n, line 87: Not sure why this file was changed as part of this PR

PR #588, opened 2023-06-30
feat(pie-modal): added `isDismissable` prop

kevinrodrigues:
[added] isDismissible prop and tests


sebastienrocks (dismissed):

packages/components/pie-modal/src/index.ts
```diff
@@ -165,34 +179,24 @@ export class PieModal extends RtlMixin(LitElement) {
             event.clientX < left ||
             event.clientX > right;
 
-        if (isClickOutsideDialog) {
+        if (isClickOutsideDialog && this.isDismissible) {
```
xander-marjoram, line 182: We could probably move the `isDismissible` check right to the beginning of the function to avoid doing all the calculations

packages/components/pie-modal/src/index.ts
```diff
@@ -165,34 +179,24 @@ export class PieModal extends RtlMixin(LitElement) {
...
             this.isOpen = false;
         }
     };
 
     /**
-     * Dispatch `ON_MODAL_CLOSE_EVENT` event.
-     * To be used whenever we close the modal.
+     * Note: We should aim to have a shareable event helper system to allow
```
xander-marjoram, line 188: Is there a ticket for this? Even if we don't fully define or refine it yet it would be good to have it created.

packages/components/pie-modal/test/component/pie-modal.spec.ts
```diff
@@ -83,10 +93,144 @@ test.describe('`Pie Modal is closed`', () => {
...
+                {
+                    props: {
+                        isOpen: true,
+                        isDismissible: true,
+                    },
+                },
+            );
+
+            // Act & Assert
+            await expect(component.locator('.c-modal-closeBtn')).toBeVisible();
```
xander-marjoram, line 120: I like the test, but we should probably hook onto a data-test-id rather than a class

packages/components/pie-modal/test/component/pie-modal.spec.ts
```diff
@@ -83,10 +93,144 @@ test.describe('`Pie Modal is closed`', () => {
...
+                {
+                    props: {
+                        isOpen: true,
+                        isDismissible: true,
+                    },
+                },
+            );
+
+            // Act
+            await component.locator('dialog').click();
```
xander-marjoram, line 155: I'm assuming that this clicks on the backdrop and not the modal then? Do we know why this works?

packages/components/pie-modal/test/component/pie-modal.spec.ts
```diff
@@ -83,10 +93,144 @@ test.describe('`Pie Modal is closed`', () => {
...
+                {
+                    props: {
+                        isOpen: true,
+                        isDismissible: false,
+                    },
+                },
+            );
+
+            // Act & Assert
+            await expect(component).not.toBeVisible();
```
xander-marjoram, line 195: Is this asserting that the modal isn't visible, rather than the close button?

packages/components/pie-modal/test/visual/pie-modal.spec.ts
```diff
@@ -88,3 +89,35 @@ sizes.forEach((size) => {
...
+            await mount(PieModal, {
+                props: {
+                    heading: 'This is a modal heading',
+                    headingLevel: 'h2',
+                    isOpen: true,
+                    isDismissible: true,
+                },
+            });
+
+            await percySnapshot(page, 'Modal with close button displayed - isDismissible: `truthy`');
```
xander-marjoram, line 105:
```suggestion
            await percySnapshot(page, 'Modal with close button displayed - isDismissible: `true`');
```

IMO we should only really use truthy/falsey when we want values like `5`/`"hello"`/`{}`/`[]` to be valid and `0`/`''`/`null`/`undefined` to be invalid. For a boolean let's stick to `true`/`false`.

packages/components/pie-modal/src/index.ts
kevinrodrigues: Yeah agree something i thought about when i was looking at this yesterday but forgot to actually do it 👍

packages/components/pie-modal/src/index.ts
kevinrodrigues: Maybe as part of the session we have today/tomorrow? I've noted it down.

packages/components/pie-modal/src/defs.ts
```diff
@@ -14,6 +14,12 @@ export interface ModalProps {
      * When true, the modal will be open.
      */
     isOpen: boolean;
+    /**
+     * (Optional) When true, the modal will be dismissible.
+     * @default true
```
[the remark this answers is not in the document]
kevinrodrigues: Fixed.

packages/components/pie-modal/test/component/pie-modal.spec.ts
kevinrodrigues: Good spot.

packages/components/pie-modal/test/component/pie-modal.spec.ts
kevinrodrigues: AFAIK you can't select a pseudo element. I saw there is a way to evaluate the page and pass in the pseudo element but im not sure that helps with the actual clicking part. Saying that though i think there is something wrong with this test as it seems to pass in both cases. I will double check it.

packages/components/pie-modal/test/component/pie-modal.spec.ts
kevinrodrigues: fixed.

.changeset/three-forks-smash.md
```diff
@@ -0,0 +1,5 @@
+---
+"@justeattakeaway/pie-modal": minor
+---
+
+[added] isDismissible prop and tests
```
ashleynolan, line 5: Super Nit

```suggestion
[Added] - isDismissible prop and tests
```

ashleynolan (approved):

packages/components/pie-modal/test/component/pie-modal.spec.ts
kevinrodrigues: I've updated them now. There isn't really a good way to selector the backdrop (in playwright). I tried targetting the `open` attribute which toggles but playwright doesn't have an API to check the attribute exists only the value of the attribute so felt hacky. I think the changes i added now works best imo now.

packages/components/pie-modal/src/defs.ts
```diff
@@ -14,6 +14,19 @@ export interface ModalProps {
...
+     *  1. The close button within the modal will be visible.
+     *  2. The user can dismiss the modal via the ESCAPE key, clicking the backdrop
+     *     or via a close button.
+     *
+     * When set to `false`:
+     *  1. The close button within the modal will be hidden.
+     *  2. The user can NOT dismiss the modal via the ESCAPE key or clicking the backdrop.
+     *
+     */
+    isDismissible: boolean;
```
jamieomaguire, line 28: I have been wondering about our usage of boolean properties and whether they would be more useful as attributes. [Spectrum takes this approach](https://opensource.adobe.com/spectrum-web-components/components/dialog/#dismissable) with their modal/components too.

If the attribute is present, that's treated as true. If the attribute is absent, that's treated as false.

@kevinrodrigues @ashleynolan @xander-marjoram I'm keen to hear what you all think?

jamieomaguire: All we'd need to do is set `Reflect: true` on the property here, I believe

packages/components/pie-modal/src/index.ts
```diff
@@ -144,55 +159,47 @@ export class PieModal extends RtlMixin(LitElement) {
...
-    /**
-     * Dispatch `ON_MODAL_OPEN_EVENT` event.
-     * To be used whenever we open the modal.
+     * Dispatch a custom event.
+     *
+     * To be used whenever we have behavioural events we want to
+     * bubble up through the modal.
      *
-     * @event
+     * @eventType
```
jamieomaguire, line 199: does this comment need some additional information?

packages/components/pie-modal/test/component/pie-modal.spec.ts
```diff
@@ -83,10 +93,162 @@ test.describe('`Pie Modal is closed`', () => {
...
             await page.locator('#dialog').click();
 
+            // Assert
             expect(messages).toHaveLength(1);
         });
     });
 });
 
+test.describe('Pie Modal `isDismissible`', () => {
+    test.describe('when `isDismissible` is truthy', () => {
```
jamieomaguire, line 106: I think it should be 

"when `isDismissible` is true"

Truthy implies 'a', 2 or true would all be valid values.

packages/components/pie-modal/test/component/pie-modal.spec.ts
```diff
@@ -83,10 +93,162 @@ test.describe('`Pie Modal is closed`', () => {
...
+                        isDismissible: false,
+                    },
+                },
+            );
+
+            // Act
+            await page.keyboard.press('Escape');
+
+            // Assert
+            await expect(component).not.toBeVisible();
```
jamieomaguire, line 186: Is there a reason we are using two different methods of asserting the modal is closed in this and the test above?

packages/components/pie-modal/test/component/pie-modal.spec.ts
```diff
@@ -83,10 +93,162 @@ test.describe('`Pie Modal is closed`', () => {
...
+        });
+    });
+
+    test.describe('when `isDismissible` is falsey', () => {
```
jamieomaguire, line 190: "when `isDismissible` is false" might be clearer, like above. Otherwise, it suggests values such as NaN, 0 and "" are all valid values

LTurns (approved):

packages/components/pie-modal/test/component/pie-modal.spec.ts
jamieomaguire: What would the value of the attribute be if it doesn't exists @kevinrodrigues? Could that potentially allow us to infer?

packages/components/pie-modal/test/component/pie-modal.spec.ts
kevinrodrigues: Yup. The backdrop is a bit of an issue to test. This one is fine because we can test it easily so we can use the `toBeVisible` helper. This helper does not work with the backdrop test above.

packages/components/pie-modal/src/index.ts
kevinrodrigues: Do you mean the overall method or just a comment about the eventType? I have a ticket around this which i will fill out soon, that ticket is to over our event system across components.

packages/components/pie-modal/test/component/pie-modal.spec.ts
kevinrodrigues: fixed

packages/components/pie-modal/test/component/pie-modal.spec.ts
kevinrodrigues: Fixed.

packages/components/pie-modal/src/defs.ts
kevinrodrigues: Fixed.

packages/components/pie-modal/src/index.ts
```diff
@@ -44,7 +49,20 @@ export class PieModal extends RtlMixin(LitElement) {
...
+    }
+
+    disconnectedCallback () : void {
+        document.removeEventListener(ON_MODAL_OPEN_EVENT, this._handleModalOpened.bind(this));
+        document.removeEventListener(ON_MODAL_CLOSE_EVENT, this._handleModalClosed.bind(this));
+        super.disconnectedCallback();
+    }
+
     firstUpdated (changedProperties: DependentMap<ModalProps>) : void {
+        this._dialog?.addEventListener('cancel', (event) => this._handleDialogCancelEvent(event));
```
jamieomaguire, line 65: Would the dialog event bubble up to `this`? If that's the case you could potentially move this listener to the connected callback above and not need to reference the dialog at all

packages/components/pie-modal/test/component/pie-modal.spec.ts
```diff
@@ -83,10 +93,162 @@ test.describe('`Pie Modal is closed`', () => {
...
+            // Act
             await page.locator('#dialog').click();
 
+            // Assert
             expect(messages).toHaveLength(1);
         });
     });
 });
 
+test.describe('Pie Modal `isDismissible`', () => {
```
jamieomaguire, line 105: may need to update these to `dismissible` now

kevinrodrigues: fixed.

packages/components/pie-modal/src/index.ts
kevinrodrigues: It does. I'm sure i added it here for a reason but it seems to work fine in the connected call back now so have moved it 👍

.changeset/three-forks-smash.md
```diff
@@ -0,0 +1,5 @@
+---
+"@justeattakeaway/pie-modal": minor
+---
+
+[Added] isDismissible prop and tests
```
jamieomaguire, line 5: just need to update the prop name here too :)

packages/tools/pie-wrapper-react/custom-elements.json
```diff
@@ -149,6 +149,14 @@
                     "description": "",
                     "name": "PieModal",
                     "members": [
+                        {
+                            "kind": "field",
+                            "name": "isDismissible",
```
jamieomaguire, line 154: annoyingly I think you need to re-build the component and push the newer version of this file

jamieomaguire (approved):

jamieomaguire (dismissed):

packages/components/pie-modal/test/component/pie-modal.spec.ts
sebastienrocks: I still see this using a class, did the changes to use a `data-test-id` get pushed?

packages/components/pie-modal/src/defs.ts
ashleynolan: Sorry I'm late to this! I'm personally of the opinion it's much clearer when Boolean props are named with `is`/`has` on them.

If we were to go down this route in the future of changing this approach, we'd want to do it everywhere, rather than just on a single instance here, otherwise we'll end up with inconsistencies.

Would it be a pain to change this back for now and we can discuss how we want to name these separately outside of this PR?

kevinrodrigues: I agree. Our components will undoubtedly get more complex as they grow so I think having a consistent group of props that signify a particular given value makes sense. Will change back for now.

packages/components/pie-modal/test/component/pie-modal.spec.ts
kevinrodrigues: The files is outdated so will show the old code still.

packages/components/pie-modal/src/index.ts
```diff
@@ -80,25 +98,23 @@ export class PieModal extends RtlMixin(LitElement) {
...
-    disconnectedCallback () : void {
-        document.removeEventListener(ON_MODAL_OPEN_EVENT, this._handleModalOpened.bind(this));
-        document.removeEventListener(ON_MODAL_CLOSE_EVENT, this._handleModalClosed.bind(this));
-        super.disconnectedCallback();
-    }
+    /**
+     * Prevents the user from dismissing the dialog via the `cancel`
+     * event (ESC key) when `isDismissible` is set to false.
+     *
+     * @param event
```
jamieomaguire, line 105: do we need some more information for the event param here?

packages/components/pie-modal/src/index.ts
```diff
@@ -80,25 +98,23 @@ export class PieModal extends RtlMixin(LitElement) {
...
+     */
+    private _handleDialogCancelEvent = (event: Event) : void => {
+        if (!this.isDismissible) event.preventDefault();
```
jamieomaguire, line 108: NIT: please can this use brackets and separate lines? I think single line `if` statements are kind of difficult to absorb for some reason

packages/components/pie-modal/src/index.ts
```diff
@@ -144,40 +157,47 @@ export class PieModal extends RtlMixin(LitElement) {
     }
 
     /**
-     * Dismisses the modal on backdrop click
+     * Dismisses the modal on backdrop click. (dependent on prop: `isDismissible`)
      *
      */
     private _handleDialogLightDismiss = (event: MouseEvent) : void => {
-        const rect = this._dialog?.getBoundingClientRect();
+        if (this.isDismissible) {
```
jamieomaguire, line 164: Keen to hear your thoughts on this - would changing this to simply do an early return if `isDismissible` is false be easier to read in the future?

That way you know you don't need to read ahead for any additional logic.

packages/components/pie-modal/src/index.ts
jamieomaguire: The event type jsdoc comment :)

packages/components/pie-modal/src/index.ts
```diff
@@ -186,19 +206,20 @@ export class PieModal extends RtlMixin(LitElement) {
...
      */
-    private _dispatchModalOpenEvent = () : void => {
-        const event = new CustomEvent(ON_MODAL_OPEN_EVENT, {
-            bubbles: true,
-            composed: true,
-        });
-
-        this.dispatchEvent(event);
-    };
+    protected renderCloseButton (): TemplateResult {
```
jamieomaguire, line 214: Out of interest why is this protected and not private?

packages/components/pie-modal/src/defs.ts
jamieomaguire: I do understand the sentiment but I wonder if this pattern is still as useful when we're using typescript? 

In terms of external consumers, if they're adding a prop to a component then they're aware of what that prop is. So I'm struggling to decide if this is still as useful as it was when we were using vanilla JS in our Vue library/Legacy JS packages.

Happy to leave it for now of course, but would be great to chat further!

.changeset/three-forks-smash.md
```diff
@@ -0,0 +1,5 @@
+---
+"@justeattakeaway/pie-modal": minor
+---
+
+[Added] dismissible prop and tests
```
ashleynolan, line 5:
```suggestion
[Added] - `isDismissible` prop and tests
```

packages/components/pie-modal/src/index.ts
ashleynolan: I'm not 100% sure, but our linting rules may force this atm. Worth checking as can't completely remember.

packages/components/pie-modal/src/index.ts
kevinrodrigues: I think this one is fairly straight forwards? The `event: Event` which is the `Event` interface passed through via the eventListener?

packages/components/pie-modal/src/index.ts
kevinrodrigues: Personally i think this reads better but have updated 👍

packages/components/pie-modal/src/index.ts
kevinrodrigues: fixed.

packages/components/pie-modal/src/index.ts
kevinrodrigues: fixed.

packages/components/pie-modal/src/index.ts
kevinrodrigues: fixed.

.changeset/three-forks-smash.md
kevinrodrigues: fixed.

packages/components/pie-modal/src/index.ts
kevinrodrigues: fixed.

packages/components/pie-modal/src/index.ts
```diff
@@ -144,10 +159,12 @@ export class PieModal extends RtlMixin(LitElement) {
     }
 
     /**
-     * Dismisses the modal on backdrop click
+     * Dismisses the modal on backdrop click. (dependent on prop: `isDismissible`)
      *
      */
     private _handleDialogLightDismiss = (event: MouseEvent) : void => {
+        if (this.isDismissible) return;
```
jamieomaguire, line 166: can we avoid inline ifs please?

kevinrodrigues: fixed.

packages/components/pie-modal/src/index.ts
kevinrodrigues: I guess the advantage here (although not tested it) is we have access via tests but since it's not being used I have removed.

jamieomaguire (approved):

raoufswe (approved):

packages/components/pie-modal/README.md
```diff
@@ -56,12 +57,12 @@ import { PieModal } from '@justeattakeaway/pie-modal/dist/react';
...
-| isOpen         | `Boolean` | `false` | Controls if the modal element is open or closed       |
-| heading*       | `String`  | -       | Sets the heading of the modal                         |
-| headingLevel   | `String`  | `h2`    | Allows you to set the heading tag (from `h1` to `h6`) |
-
+| Property      | Type      | Default | Description                                           |
+|---------------|-----------|---------|-------------------------------------------------------|
+| isOpen        | `Boolean` | `false` | Controls if the modal element is open or closed       |
+| heading*      | `String`  | -       | Sets the heading of the modal                         |
+| headingLevel  | `String`  | `h2`    | Allows you to set the heading tag (from `h1` to `h6`) |
+| isDismissible | `Boolean` | `false` | Controls if the modal element is dismissible or not.  |
```
raoufswe, line 65: nit: I just took a look at react.pie.design, snacks and fozzie and it seems the default value for the modal to show the close button, is there a reason we aren't doing that here?

raoufswe (approved):

packages/components/pie-modal/README.md
ashleynolan: I'm happy for the default to change to `true` if we think that's the most common usecase – I would check with @RLumleyJET in terms of what they consider the most used case and go for that.

ashleynolan (approved):
Looks good – happy for this to be merged and we can always update the default state of that prop based on design feedback

PR #597, opened 2023-07-04
refactor(pie-monorepo): set component props to public

jamieomaguire:
---
"@justeattakeaway/pie-icon-button": minor
---

[Changed] - Set all component props to public


LTurns (approved):

xander-marjoram:
I'm not sure this was necessary, as `public` is the default value, but I don't have a problem with it 😄

PR #599, opened 2023-07-04
feat(pie-modal): [Added] - Add `returnFocusAfterCloseSelector` prop

xander-marjoram:

- [Added] - Add `returnFocusAfterCloseSelector` prop
- [Changed] - Extract `renderTestPieModal` into test helper file so it can be used for component tests too
- [Added] - Add README documentation for new props
- [Changed] - Refactor modal story to reduce duplication
- [Fixed] - Fix console errors when switching between stories

One of the acceptance criteria was that if no value is passed in for this prop, then focus should be returned to the next element.
The native `dialog` component actually handles this behaviour itself - returning focus to where it was before the modal opened - so I have not specifically handled this in the code.

### Reviewer 1

### Reviewer 2


packages/components/pie-modal/test/component/pie-modal.spec.ts
```diff
@@ -88,5 +95,64 @@ test.describe('`Pie Modal is closed`', () => {
             expect(messages).toHaveLength(1);
         });
     });
-});
 
+    test('should return focus to specified element', async ({ mount, page }) => {
```
kevinrodrigues, line 99: I think these would be good to have a wrapping describe.

`test.describe('returnFocusAfterCloseSelector: when specified element is passed in')`
    `test('should return focus to that element')`

packages/components/pie-modal/test/component/pie-modal.spec.ts
```diff
@@ -88,5 +95,64 @@ test.describe('`Pie Modal is closed`', () => {
...
+        await page.locator(closeButtonSelector).click();
+
+        const focusedElement = await page.locator(':focus');
+        const focusedElementId = await focusedElement.getAttribute('id');
+
+        // Assert
+        expect(focusedElementId).toBe('focus-me');
+    });
+
+    test('should return focus to first matching element', async ({ page, mount }) => {
```
kevinrodrigues, line 129: Similar to above.

kevinrodrigues, line 129: Do you need a test for when the modal defaults it focus to the opening element?

packages/components/pie-modal/test/component/pie-modal.spec.ts
kevinrodrigues, line 99: Does ESC need a test too as that should also focus back?

webbertakken (dismissed):

packages/components/pie-modal/src/index.ts
```diff
@@ -115,6 +118,24 @@ export class PieModal extends RtlMixin(LitElement) {
...
+    private _returnFocus () {
+        const selector = this.returnFocusAfterCloseSelector?.trim();
+
+        if (selector) {
+            try {
+                const focalPoint = document.querySelector(selector) as HTMLElement;
+                focalPoint?.focus();
+            } catch (ignore) {
+                // Not necessary to throw an error here.
+            }
```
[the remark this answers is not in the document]
webbertakken: In more elaborate cases I have seen this rationale reduce the footprint of the code significantly.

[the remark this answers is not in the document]
raoufswe: i believe the try/catch here isn't needed as none of the logic will throw

raoufswe (approved):

packages/components/pie-modal/test/component/pie-modal.spec.ts
xander-marjoram: I did try this yesterday and couldn't get it working, but I've just made some changes around the focus so I'll give it another go and maybe it'll work this time

packages/components/pie-modal/src/index.ts
```diff
@@ -115,6 +118,24 @@ export class PieModal extends RtlMixin(LitElement) {
         }
     }
 
+    /**
+     * Return focus to the specified element, providing the selector is valid
+     * and the chosen element can be found.
+     * Fails silently.
+     */
+    private _returnFocus () {
```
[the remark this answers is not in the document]
xander-marjoram: `void` is actually inferred for all of our void functions so it's not necessary. I'll add it though just for consistency.

packages/components/pie-modal/test/component/pie-modal.spec.ts
xander-marjoram: Nice spot. This has actually opened a small can of worms which I need to debug 😄

xander-marjoram: Good shout, updated

packages/components/pie-modal/src/index.ts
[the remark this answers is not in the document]
xander-marjoram: So that method will throw in a couple of scenarios:
1. An empty string is passed in (although I catch this with `.trim()`
2. An invalid selector is passed in, e.g., `h1[`.

In case 2 I think we probably do want to throw an error so that the consuming app can fix whatever they're doing. What do you think?

In terms of the type assertion I'm doing here, `as HTMLElement`, that is actually necessary because `focus` doesn't exist on `Element` but it does on `HTMLElement`. Is that what you were referring to or was that more to do with the handling of the error?

packages/components/pie-modal/src/index.ts
```diff
@@ -59,14 +62,18 @@ export class PieModal extends RtlMixin(LitElement) {
     }
 
     disconnectedCallback () : void {
+        super.disconnectedCallback();
```
xander-marjoram, line 65: The Lit docs have the `super` call at the start of the method so we probably should too
https://lit.dev/docs/components/lifecycle/#disconnectedcallback

packages/components/pie-modal/src/index.ts
```diff
@@ -78,9 +85,10 @@ export class PieModal extends RtlMixin(LitElement) {
      */
     private _handleModalOpened () : void {
         disableBodyScroll(this);
-        //  We require this because toggling the prop `isOpen` itself won't
-        //  allow the dialog to open in the correct way (with the default background),
-        //  the method `showModal()` needs to be invoked.
+        if (this._dialog?.hasAttribute('open') || !this._dialog?.isConnected) {
```
xander-marjoram, line 88: This line resolves these two errors in Storybook, and it's probably good defensive practice regardless.

packages/components/pie-modal/src/index.ts
kevinrodrigues: It might not make a difference but i think it kinda makes sense calling it last because as i understand it the disconnected super call cleans up the overall class. If we call it first there might be other calls inside here that dont get cleaned up? Saying that happy for you to change it but would be interesting to ask the LIT team about this. I could put a message in the room and ask.

packages/components/pie-modal/test/component/pie-modal.spec.ts
```diff
@@ -101,11 +106,126 @@ test.describe('`Pie Modal is closed`', () => {
             expect(messages).toHaveLength(1);
         });
     });
+
+    test.describe('`returnFocusAfterCloseSelector` prop', () => {
```
kevinrodrigues, line 110: Nice

packages/components/pie-modal/src/index.ts
kevinrodrigues: LIT team:

`It won't matter at all in that case. The LitElement disconnectedCallback just notifies AsyncDirectives and Reactive Controllers that they're disconnected
My advice based on code style is to put the super.disconnectedCallback(); at the top of the method, because that's where the super() call in a constructor must go, but that's mainly for rease of reading`

packages/components/pie-modal/src/index.ts
```diff
@@ -164,8 +177,8 @@ export class PieModal extends RtlMixin(LitElement) {
     }
 
     /**
-     * Dismisses the modal on backdrop click. (dependent on prop: `isDismissible`)
-     *
+     * Dismisses the modal on backdrop click if `isDismissible` is `true`.
+     * @param {MouseEvent} event - the click event targetting the modal/backdrop
```
kevinrodrigues, line 181: 👍

kevinrodrigues (approved):

packages/components/pie-modal/src/index.ts
kevinrodrigues: Is it worth clearing the open state before re-opening it? i.e close it first, would that clear these errors. Just a thought i'd check.

packages/components/pie-modal/src/index.ts
```diff
@@ -78,9 +85,10 @@ export class PieModal extends RtlMixin(LitElement) {
      */
     private _handleModalOpened () : void {
         disableBodyScroll(this);
-        //  We require this because toggling the prop `isOpen` itself won't
-        //  allow the dialog to open in the correct way (with the default background),
-        //  the method `showModal()` needs to be invoked.
+        if (this._dialog?.hasAttribute('open') || !this._dialog?.isConnected) {
+            return;
+        }
+        // The ::backdrop pseudoelement is only shown if the modal is opened via JS
```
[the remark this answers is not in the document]
xander-marjoram: Thanks for the suggestion, I'll add that in my next PR if you don't mind

fernandofranca (approved):

PR #603, opened 2023-07-05
feat(pie-button): add loading state

jamieomaguire:

### Reviewer 1

### Reviewer 2

### DIFF
--- a/apps/pie-storybook/stories/pie-button.stories.ts
+++ b/apps/pie-storybook/stories/pie-button.stories.ts
@@ -26,6 +26,9 @@ export default {
         isFullWidth: {
             control: 'boolean',
         },
+        loading: {
+            control: 'boolean',
+        },
         slot: {
             control: 'text',
         },
@@ -36,6 +39,7 @@ export default {
         variant: 'primary',
         disabled: false,
         isFullWidth: false,
+        loading: false,
     },
     parameters: {
         design: {
@@ -48,13 +52,14 @@ export default {
 type ButtonProps = ButtonPropsBase & { slot: string }
 
 const Template = ({
-    size, variant, type, disabled, isFullWidth, slot,
+    size, variant, type, disabled, isFullWidth, loading, slot,
 }: ButtonProps): TemplateResult => html`
         <pie-button
             size="${size}"
             variant="${variant}"
             type="${type}"
             ?disabled="${disabled}"
+            ?loading="${loading}"
             ?isFullWidth="${isFullWidth}">
             ${slot}
         </pie-button>
@@ -66,6 +71,7 @@ const defaultArgs: ButtonProps = {
     variant: 'primary',
     disabled: false,
     isFullWidth: false,
+    loading: false,
     slot: 'This is Lit!',
 };
 
--- a/apps/pie-storybook/stories/pie-button.stories.ts
+++ b/apps/pie-storybook/stories/pie-button.stories.ts
@@ -26,7 +26,7 @@ export default {
         isFullWidth: {
             control: 'boolean',
         },
-        loading: {
+        isLoading: {
             control: 'boolean',
         },
         slot: {
@@ -39,7 +39,7 @@ export default {
         variant: 'primary',
         disabled: false,
         isFullWidth: false,
-        loading: false,
+        isLoading: false,
     },
     parameters: {
         design: {
@@ -52,14 +52,14 @@ export default {
 type ButtonProps = ButtonPropsBase & { slot: string }
 
 const Template = ({
-    size, variant, type, disabled, isFullWidth, loading, slot,
+    size, variant, type, disabled, isFullWidth, isLoading, slot,
 }: ButtonProps): TemplateResult => html`
         <pie-button
             size="${size}"
             variant="${variant}"
             type="${type}"
             ?disabled="${disabled}"
-            ?loading="${loading}"
+            ?isLoading="${isLoading}"
             ?isFullWidth="${isFullWidth}">
             ${slot}
         </pie-button>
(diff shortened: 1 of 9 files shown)


packages/components/pie-button/src/button.scss
```diff
@@ -161,18 +216,30 @@
         --btn-padding: 6px var(--dt-spacing-b);
         --btn-font-size: calc(var(--dt-font-size-14) * 1px);
         --btn-line-height: calc(var(--dt-font-size-14-line-height) * 1px);
+        --btn-min-inline-size: 55px;
```
ashleynolan, line 219: @ashleynolan to check these

jamieomaguire: Yes please! I noticed some buttons, even of the same variant, had differing heights (50px vs 55px for example) so would be good to clarify.

packages/components/pie-button/src/index.ts
```diff
@@ -32,12 +34,19 @@ export class PieButton extends LitElement {
     @property({ type: Boolean })
     public disabled = false;
 
+    @property({ type: Boolean, reflect: true })
+    public isLoading = false;
+
     @property({ type: Boolean })
     public isFullWidth = false;
 
+    private renderLoadingSpinner () {
```
kevinrodrigues, line 43: You could (not fully necessary tho) use the `TemplateResult` type check here from Lit since we are specifying a return value type of template.

packages/components/pie-button/src/index.ts
```diff
@@ -48,7 +57,7 @@ export class PieButton extends LitElement {
                 size=${size}
                 ?disabled=${disabled}
                 ?isFullWidth=${isFullWidth}>
-                <slot></slot>
+                ${isLoading ? this.renderLoadingSpinner() : html`<slot></slot>`}
```
kevinrodrigues, line 60: We can make use of the `until` directive here: https://lit.dev/docs/templates/directives/#until

packages/components/pie-button/src/index.ts
ashleynolan: I've removed this now – after speaking to design, they wanted to keep the height/width of the button dependent on the text in the button, so rather than rendering the loading spinner as markup, I changed it to add it as a pseudo-element (so this isn't needed now)

packages/components/pie-button/src/index.ts
ashleynolan: I've removed this now – after speaking to design, they wanted to keep the height/width of the button dependent on the text in the button, so rather than rendering the loading spinner as markup, I changed it to add it as a pseudo-element (so this isn't needed now)

packages/components/pie-button/src/button.scss
ashleynolan: This shouldn't be an issue now, as the button is the same height as the text. Design will check variant heights in more detail when they go through the design review of the button shortly.

ashleynolan (approved):

packages/components/pie-button/src/defs.ts
```diff
@@ -4,28 +4,27 @@ export const variants = ['primary', 'secondary', 'outline', 'ghost', 'inverse',
 
 export interface ButtonProps {
     /**
-     * (Optional) What size the button should be.
-     * @default "medium"
+     * What size the button should be.
```
raoufswe, line 7: are changes here intentional?

jamieomaguire: I'd recommend chatting with @xander-marjoram about this :)

raoufswe:
do you think we should remove the spinner/animation from the DOM completely when isLoading is set to false? considering the number of use cases of the loading button is a few, we might be overloading our apps/browsers with unnecessary animation computation. 
https://github.com/justeattakeaway/pie/assets/28605803/41ab4dc8-4e15-4e69-b59c-4373d7b6d88b

jamieomaguire:
> do you think we should remove the spinner/animation from the DOM completely when isLoading is set to false? considering the number of use cases of the loading button is a few, we might be overloading our apps/browsers with unnecessary animation computation. https://github.com/justeattakeaway/pie/assets/28605803/41ab4dc8-4e15-4e69-b59c-4373d7b6d88b

Fully agree - I didn't spot this but will update now 👍 thanks!

raoufswe:
based on the design, the size of the spinner should be 20px, and it's relatively bigger now as the border width is added on top of the width and height resulting in a larger spinner. 

maybe we need to add   `box-sizing: border-box;` or subtract the border width from the total height ane width? 

Screenshot 2023-07-11 at 10 04 42

raoufswe:
while the loading state for the primary variant looks correct, the other variants behave differently from the designs currently. not sure if it's meant to be handled in other PRs.

secondary, ghost, and outline should be like this based on designs:

Screenshot 2023-07-11 at 10 22 55

however, they inherit their own variant styles instead at the moment. 

https://github.com/justeattakeaway/pie/assets/28605803/6ae5f90c-996c-454f-ba15-1bad35a5d725

jamieomaguire:
> while the loading state for the primary variant looks correct, the other variants behave differently from the designs currently. not sure if it's meant to be handled in other PRs.
> 
> secondary, ghost, and outline should be like this based on designs:
> 
> Screenshot 2023-07-11 at 10 22 55
> however, they inherit their own variant styles instead at the moment.
> 
>  Screen.Recording.2023-07-11.at.10.34.11.mov

Very true - we're going to pick these up with design later on as we're unsure about them all using that particular style. So the difference is deliberate for now :)

raoufswe (approved):

apps/pie-storybook/stories/pie-button.stories.ts
```diff
@@ -26,6 +26,9 @@ export default {
         isFullWidth: {
             control: 'boolean',
         },
+        isLoading: {
```
kevinrodrigues, line 29: Could you add this to the readme please?

kevinrodrigues (approved):

LTurns (approved):

PR #612, opened 2023-07-12
fix(pie-wrapper-react): add events to custom-elements.json

jamieomaguire:
Currently when a component is build, the `cem analyser` will create a `custom-elements.json` entry for each component. This file is used to create a facade that React uses to interface with our web components.

Part of this facade adds callback hooks to our components where they emit events. This is because React is callback based rather than event-based. React is working to address this for improved integration with Web Components in the future.

A bug in our current implementation is that the `custom-elements.json` file is, for some reason, not detecting emitted events in our components. This means that the events never get added to the json file, never have callbacks added for React, meaning React cannot add callbacks to our components.

To aid in fixing this, I have found that adding `@jsdoc` comments declaring what events a component can emit at the top of the class ensures that they are added to the json file. 

To adhere to callback naming syntax I have read in the Lit docs, I then transform these event names to remove hyphens. So `pie-modal-open` becomes `onpiemodalopen`. [Example in the Lit docs showing this](https://lit.dev/docs/frameworks/react/#usage). Note - `on` is prepended to the event name.

I have also fixed the code so that event description comments are correctly added. Before, they would break the formatting of the `react.ts` files generated and thus would break the component build.

### Before
`react.ts`:

```
import * as React from 'react';
import { createComponent, EventName } from '@lit-labs/react';
import { PieModal as PieModalReact } from './index';

export const PieModal = createComponent({
    displayName: 'PieModal',
    elementClass: PieModalReact,
    react: React,
    tagName: 'pie-modal',
    events: { },
});
```

### After:
```

import * as React from 'react';
import { createComponent, EventName } from '@lit-labs/react';
import { PieModal as PieModalReact } from './index';

export const PieModal = createComponent({
    displayName: 'PieModal',
    elementClass: PieModalReact,
    react: React,
    tagName: 'pie-modal',
    events: {
        onpiemodalopen: 'pie-modal-open' as EventName<Event>, // when the modal is opened.
        onpiemodalclose: 'pie-modal-close' as EventName<Event>, // when the modal is closed.
    },
});

```

### DIFF
--- a/packages/tools/pie-wrapper-react/__tests__/__snapshots__/wrapper.test.js.snap
+++ b/packages/tools/pie-wrapper-react/__tests__/__snapshots__/wrapper.test.js.snap
@@ -3,16 +3,18 @@
 exports[`React Wrapper should be added from mock custom elements JSON 1`] = `
 "
 import * as React from 'react';
+import { createComponent, EventName } from '@lit-labs/react';
 import { MockComponent as MockComponentReact } from './index';
-import { createComponent } from '@lit-labs/react';
-import { EventName } from '@lit-labs/react';
 
 export const MockComponent = createComponent({
     displayName: 'MockComponent',
     elementClass: MockComponentReact,
     react: React,
     tagName: 'mock-component',
-    events: { onCustomEvent: 'CustomEvent' as EventName<CustomEvent>, onAnotherCustomEvent: 'AnotherCustomEvent' as EventName<CustomEvent>, },
+    events: {
+        onCustomEvent: 'CustomEvent' as EventName<CustomEvent>, 
+        onAnotherCustomEvent: 'AnotherCustomEvent' as EventName<CustomEvent>, 
+    },
 });
 "
 `;
@@ -20,15 +22,15 @@ export const MockComponent = createComponent({
 exports[`React Wrapper should be named the same as the component itself 1`] = `
 "
 import * as React from 'react';
-import { ButtonComponent as ButtonComponentReact } from './index';
 import { createComponent } from '@lit-labs/react';
+import { ButtonComponent as ButtonComponentReact } from './index';
 
 export const ButtonComponent = createComponent({
     displayName: 'ButtonComponent',
     elementClass: ButtonComponentReact,
     react: React,
     tagName: 'mock-component',
-    events: { },
+    events: {},
 });
 "
 `;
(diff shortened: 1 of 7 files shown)


packages/tools/pie-wrapper-react/__tests__/__snapshots__/wrapper.test.js.snap
```diff
@@ -3,48 +3,50 @@
...
-import { EventName } from '@lit-labs/react';
 
 export const MockComponent = createComponent({
     displayName: 'MockComponent',
     elementClass: MockComponentReact,
     react: React,
     tagName: 'mock-component',
-    events: { onCustomEvent: 'CustomEvent' as EventName<CustomEvent>, onAnotherCustomEvent: 'AnotherCustomEvent' as EventName<CustomEvent>, },
+    events: {
+        onCustomEvent: 'CustomEvent' as EventName<CustomEvent>,
```
jamieomaguire, line 15: I would suggest as a rule of thumb we keep our events to single words if possible. I think `pie-modal-open` for example could simply be `open`. Which would become `onopen`. Keeping them as short as possible will improve readability.

jamieomaguire: Bear in mind that native browser events follow lowercase, no hyphen formats. Such as `onreadystatechange`, `touchstart`, `fullscreenerror`. Depends how much we want to mirror native naming conventions

jamieomaguire: @raoufswe what we could always do is keep events in the web components like `modal-open` but camel case them for the react wrapper so it would become `onModalOpen`. Perhaps that would be a best of both solution?

ashleynolan: Agree with this – happy for this to be done in a follow up PR

packages/tools/pie-wrapper-react/scripts/add-react-wrapper.js
```diff
@@ -81,27 +81,33 @@ export function addReactWrapper (customElementsObject, folderName = process.argv
         components.forEach((component) => {
             const events = getEvents(component.class);
 
+            // Prepare the event names and comments
+            let eventNames = '';
+            if (component.class.events && component.class.events.length > 0) {
+                eventNames = events?.flat().reduce((pre, event) => {
```
xander-marjoram, line 87:
```suggestion
            if (component.class?.events?.length) {
                eventNames = events?.flat()?.reduce((pre, event) => {
```

jamieomaguire: Just chatted in person but for others viewing this, I think it would be preferable for us to avoid optional chaining and instead allow this process to error and break if there is any missing required data. This way we are less likely to introduce a possibility of silently failing when building the wrappers.

xander-marjoram (approved):

ashleynolan (approved):

packages/tools/pie-wrapper-react/__tests__/__snapshots__/wrapper.test.js.snap
jamieomaguire: Ta - will raise this shortly

PR #613, opened 2023-07-12
test(pie-modal): moved tests from visual to component

JoshuaNg2332:

- Moved some visual tests which were better suited to component tests.
- Added a `data-test-id` to modal

### Reviewer 1

### Reviewer 2


packages/components/pie-modal/test/component/pie-modal.spec.ts
```diff
@@ -373,3 +374,49 @@ test.describe('`isDismissible` prop', () => {
         });
     });
 });
+
+test.describe('isOpen prop', () => {
+    test('should not render when isOpen = false', async ({ mount, page }) => {
```
jamieomaguire, line 379: just a nit here, the component _will_ render. It just won't open.

packages/components/pie-modal/test/component/pie-modal.spec.ts
```diff
@@ -373,3 +374,49 @@ test.describe('`isDismissible` prop', () => {
...
+            },
+        },
+    );
+
+    // Scroll 800 pixels down the page
+    await page.mouse.wheel(0, 800);
+
+    await page.waitForTimeout(3000); // The mouse.wheel function causes scrolling, but doesn't wait for the scroll to finish before returning.
+
+    await expect(page.locator(modalSelector)).toBeInViewport();
```
jamieomaguire, line 421: This confuses me a bit. Even if you could scroll, the modal would still be in the viewport wouldn't it?

packages/components/pie-modal/test/component/pie-modal.spec.ts
JoshuaNg2332: @jamieomaguire  does `'should not render open when isOpen = false'` sound ok? Not sure how best to phrase it.

packages/components/pie-modal/test/component/pie-modal.spec.ts
JoshuaNg2332: Exactly. But we want to test the act of scrolling to signify that we're making sure that even if it did scroll, it would still display the modal in the viewport. 
So if something changed and broke the modal being locked in place, this test would cover that.

packages/components/pie-modal/src/index.ts
```diff
@@ -163,6 +163,7 @@ export class PieModal extends RtlMixin(LitElement) implements ModalProps {
             <dialog
                 id="dialog"
                 class="c-modal"
+                data-test-id="c-modal"
```
xander-marjoram, line 166: The `c-` prefix is more of a CSS thing. I would change this just to `modal` or `pie-modal`.

packages/components/pie-modal/test/component/pie-modal.spec.ts
xander-marjoram: Or "should not be visible ..." because that's the playwright method we're using

packages/components/pie-modal/test/component/pie-modal.spec.ts
jamieomaguire: So my concern is that I don't think this test verifies that you cannot scroll the page. With scrolling enabled/disabled the modal is always in viewport. So how is this testing that we cannot scroll?

JoshuaNg2332: Ohhh, I think I misunderstood this test. I think what we can do here is that we can test that X list item is not in the viewport when we scroll. I can change that.

JoshuaNg2332: I've pushed a change now that rectifies this test. Essentially, I added a last item in the list for it to assert against if it is/isn't in the viewport after scrolling. To expand the test to be more thorough, I also added a soft assert on the top item to check it's still in view and for the last item to not be in view.

packages/components/pie-modal/src/index.ts
```diff
@@ -164,6 +164,7 @@ export class PieModal extends RtlMixin(LitElement) implements ModalProps {
             <dialog
                 id="dialog"
                 class="c-modal"
+                data-test-id="pie-modal"
```
fernandofranca, line 167: Is this attribute something that all next components will have to implement?

packages/components/pie-modal/test/component/pie-modal.spec.ts
```diff
@@ -4,12 +4,13 @@ import { PieIconButton } from '@justeattakeaway/pie-icon-button';
...
     WebComponentTestWrapper,
 } from '@justeattakeaway/pie-webc-testing/src/helpers/components/web-component-test-wrapper/WebComponentTestWrapper.ts';
-import { renderTestPieModal } from '../helpers/index.ts';
+import { createScrollablePageHTML, renderTestPieModal } from '../helpers/index.ts';
 
 import { PieModal } from '@/index';
 import { headingLevels } from '@/defs';
 
 const closeButtonSelector = '[data-test-id="modal-close-button"]';
+const modalSelector = '[data-test-id="pie-modal"]';
```
fernandofranca, line 13: Maybe we could use the class `componentSelector` here and also in the dialog attribute.

packages/components/pie-modal/src/index.ts
xander-marjoram: This is just an attribute that is only used for locating the element during tests, it's a pattern we're used to following but if it doesn't seem logical then I'm sure we're happy to discuss it going forwards

packages/components/pie-modal/test/component/pie-modal.spec.ts
```diff
@@ -373,3 +374,58 @@ test.describe('`isDismissible` prop', () => {
...
+    // Act
+    // Scroll 800 pixels down the page
+    await page.mouse.wheel(0, 1000);
+
+    // The mouse.wheel function causes scrolling, but doesn't wait for the scroll to finish before returning.
+    await page.waitForTimeout(3000);
+
+    // Assert
+    await expect.soft(page.getByText('Test Page')).toBeInViewport();
+    await expect(page.getByText('Test Item')).not.toBeInViewport();
```
jamieomaguire, line 430: I think we can make the items a bit more descriptive here. Ideally I shouldn't need to check the other file to understand what test page and test item are doing.

Even renaming them to something like `top of page copy` and `bottom of page copy` might be useful

packages/components/pie-modal/test/component/pie-modal.spec.ts
```diff
@@ -373,3 +374,58 @@ test.describe('`isDismissible` prop', () => {
...
+                isOpen: true,
+            },
+        });
+
+        // Assert
+        await expect(page.locator(modalSelector)).toBeVisible();
+    });
+});
+
+test('Should not be able to scroll when modal is open', async ({ page, mount }) => {
```
jamieomaguire, line 404: in the previous tests we use the condition `when isOpen = true`. Should we stick to that for this test as well?

packages/components/pie-modal/test/component/pie-modal.spec.ts
```diff
@@ -373,3 +374,58 @@ test.describe('`isDismissible` prop', () => {
         });
     });
 });
+
+test.describe('isOpen prop', () => {
```
jamieomaguire, line 378: I think some tests checking the scroll enabling logic would be useful here.

Ideally, we want to check that scrolling is re-enabled when:
- the modal is closed
- the modal is removed from the DOM

packages/components/pie-modal/test/component/pie-modal.spec.ts
JoshuaNg2332: I get changing the class to `componentSelector` but don't feel like using `componentSelector` for the attribute name is the right approach seeing as this will only be used for testing and having it called a `data-test-id` is more clear to people what the attribute is used for.

packages/components/pie-modal/test/component/pie-modal.spec.ts
JoshuaNg2332: @jamieomaguire By this comment, when you say the modal is removed from the DOM, do you mean that it existed at the beginning of the test and we're just saying that `isOpen: false`? Or do you mean that we never loaded the modal into the dom as part of the test wrapper in the first place? 🤔

jamieomaguire (approved):

PR #615, opened 2023-07-12
feat(pie-wrapper-react): use the litelement flag for cem analyers

jamieomaguire:
Adds the `--litelement` flag to our `cem` command to enable special handling for LitElement syntax. It is recommended by Lit to use this flag.


ashleynolan (approved):
:shipit:

PR #619, opened 2023-07-13
feat(pie-modal): Add default buttons to footer

xander-marjoram:
- [Added] - Default modal footer
- [Added] - Height styles for the different sizes
- [Added] - Scroll shadow for modal contents
- [Removed] - Unnecessary `:focus-visible` outline for the modal
- [Fixed] - Modal width should be able to exceed 600px for `isFullWidthBelowMid`
- [Added] - Tests to validate leading/supporting action buttons
- [Added] - Missing tests to check that the modal closes when expected to
- [Changed] - Updated tests to verify that the correct event is being triggered

### Reviewer 1

### Reviewer 2


packages/components/pie-modal/src/modal.scss
```diff
@@ -57,7 +96,54 @@
...
+        padding-inline: var(--modal-content-padding);
+
+        overflow-y: auto;
+
+        &--scrollable {
+            background:
+                // Scroll shadow cover - a top-to-bottom opacity gradient from transparent to the component background colour
+                linear-gradient(to bottom, transparent, var(--dt-color-container-default) 75%) center bottom,
+                // Scroll shadow
+                linear-gradient(transparent, var(--dt-color-border-strong)) center bottom;
```
xander-marjoram, line 120: The design tokens we currently have for elevation drop shadows are not compatible with methods to make the shadow fade away as you scroll. We'll have to work with design to try and find a longer-term solution for this.

xander-marjoram:
There's a small issue on the Percy screenshot for the large and `isFullWidthBelowMid` modals showing a scrollbar on Firefox, but I can't replicate this locally.

packages/components/pie-modal/src/index.ts
```diff
@@ -154,33 +154,60 @@ export class PieModal extends RtlMixin(LitElement) implements ModalProps {
         }
     }
 
-    render () {
-        const {
-            heading,
-            headingLevel = 'h2',
-            size,
-            isFullWidthBelowMid,
```
ashleynolan, line 162: I quite like this destructuring to simplify the variables – was there a reason for removing it?

xander-marjoram: After extracting the header and footer creation into their own functions, there were only two variables left that were being used.

By removing the destructuring, it lets us use the arrow-body syntax which I think helps dissuade people from adding extra logic into the render function, instead moving it out into a separate function.

What do you think?

ashleynolan: I think it's probably worth catching up with @jamieomaguire as well when he's in tomorrow – he's done some restructuring of the render function in the `hasBackButton` PR as well which will definitely conflict with this, and I think we probably need to be careful not to abstract out everything into separate render functions (as I think it can have the opposite effect then whereby it makes things harder to find).

Think we likely will need to discuss and work out the best way to merge both of your PRs into main in the easiest way.

packages/components/pie-modal/src/modal.scss
```diff
@@ -23,31 +35,58 @@
...
+    --modal-margin-none: var(--dt-spacing-none);
+    --modal-margin-small: var(--dt-spacing-g);
+    --modal-margin-large: var(--dt-spacing-j);
+    --modal-margin-block: var(--modal-margin-small);
+
+    @media (min-width: $breakpoint-wide) {
+        --modal-margin-block: var(--modal-margin-large);
+    }
+
+    --modal-block-size: fit-content;
```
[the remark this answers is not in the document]
xander-marjoram: Not sure what you mean by this, unless you'd like an explicit comment to say that all styles are for medium by default?

packages/components/pie-modal/src/index.ts
```diff
@@ -154,33 +154,60 @@ export class PieModal extends RtlMixin(LitElement) implements ModalProps {
...
+     * Extracted to keep the render method succinct.
+     * @returns {TemplateResult} HTML template for the modal's footer
+    */
+    private _renderModalFooter = () : TemplateResult => html`
+        <pie-button
+            variant="primary"
+            type="submit"
+            @click="${() => this._dialog?.close('leading')}"
+            data-test-id="modal-leading-action">
+            Confirm
```
jamieomaguire, line 201: should these be user provided?

packages/components/pie-modal/src/index.ts
```diff
@@ -154,33 +154,60 @@ export class PieModal extends RtlMixin(LitElement) implements ModalProps {
...
+            ${this.isDismissible ? this.renderCloseButton() : nothing}`;
     }
 
+    /**
+     * Returns the HTML content for the footer.
+     * Extracted to keep the render method succinct.
+     * @returns {TemplateResult} HTML template for the modal's footer
+    */
+    private _renderModalFooter = () : TemplateResult => html`
+        <pie-button
```
kevinrodrigues, line 196: Not entirely necessary but you could store the individual aspects of the button in an array then pass it into one element (instead of having the two bits of markup) as a loop / map? (not tested this tho)

```javascript
private renderModalFooter(): TemplateResult {
  const buttonsValues = [
    { variant: 'primary', type: ... },
    { variant: 'ghost', type: ... }
  ];

  return html`
    ${buttons.map(button => html`
      <pie-button
        variant=${button.variant}....etc}
      >
        ${button.label}
      </pie-button>
    `)}
  `;
}
```

packages/components/pie-modal/test/component/pie-modal.spec.ts
```diff
@@ -71,39 +66,76 @@ test.describe('When modal is closed', () => {
...
+            // Arrange
+            await mount(PieModal, {
+                props: {
+                    isOpen: true,
+                    isDismissible: true,
+                },
+            });
+
+            const modal = page.locator(modalSelector);
+            expect(modal).toBeVisible();
```
kevinrodrigues, line 90: Do you need this expect?

packages/components/pie-modal/test/component/pie-modal.spec.ts
```diff
@@ -71,39 +66,76 @@ test.describe('When modal is closed', () => {
...
+            // Arrange
+            await mount(PieModal, {
+                props: {
+                    isOpen: true,
+                    isDismissible: true,
+                },
+            });
+
+            const modal = await page.locator(modalSelector);
+            expect(modal).toBeVisible();
```
kevinrodrigues, line 132: Similar question, do you need this one?

packages/components/pie-modal/test/component/pie-modal.spec.ts
```diff
@@ -350,26 +385,71 @@ test.describe('`isDismissible` prop', () => {
...
+                // Arrange
+                await mount(PieModal, {
+                    props: {
+                        heading: 'Modal Header',
+                        isOpen: true,
+                    },
+                });
+
+                const modal = await page.locator(modalSelector);
+                expect(modal).toBeVisible();
```
kevinrodrigues, line 425: Similar to above.

packages/components/pie-modal/test/component/pie-modal.spec.ts
```diff
@@ -350,26 +385,71 @@ test.describe('`isDismissible` prop', () => {
...
 
             // Assert
-            await expect(component.locator('dialog')).toBeVisible();
+            await expect(modal).toBeVisible();
+        });
+    });
+});
+
+test.describe('actions', () => {
+    ['leading', 'supporting'].forEach((actionName) => {
```
kevinrodrigues, line 411: Nice!

packages/components/pie-modal/test/component/pie-modal.spec.ts
xander-marjoram: About to have a chat with Josh about this, likely not, although it does prove the test works! 😄

packages/components/pie-modal/src/index.ts
xander-marjoram: Yes, I've now made a ticket to cover copy and i18n in general

packages/components/pie-modal/src/index.ts
xander-marjoram: This was just setting up the defaults, and I didn't want to make it too complicated.
With the upcoming tickets for the leading and supporting actions, there are some opportunities to try out some stuff like this 👍🏻

kevinrodrigues (approved):

packages/components/pie-modal/src/index.ts
```diff
@@ -209,36 +208,55 @@ export class PieModal extends RtlMixin(LitElement) implements ModalProps {
         `;
     }
 
-    render () {
+    public render () {
```
jamieomaguire, line 211: I'm unsure about `public`. I've seen examples from Lit using `protected` too - Doesn't need to block this PR but would be good to understand the 'recommended' access modifier here

xander-marjoram: That's a good point. `public` is the default so it's probably helpful to have it there as a reminder, in case we do want to change it in future

packages/components/pie-modal/test/component/pie-modal.spec.ts
```diff
@@ -1,31 +1,47 @@
...
-    );
-
-    // Removing the element so it's not present in the tests (but is still registered in the DOM)
-    await page.evaluate(() => {
-        const element : Element | null = document.querySelector('pie-icon-button');
-        element?.remove();
+test.beforeEach(async ({ mount }) => {
+    await (await mount(PieButton)).unmount();
+    await (await mount(PieIconButton)).unmount();
+});
```
jamieomaguire, line 28: I have absolutely no idea why `unmount` didn't work for me when I initial wrote this!

jamieomaguire (approved):

kevinrodrigues (dismissed):

packages/components/pie-modal/src/modal.scss
```diff
@@ -22,6 +22,18 @@
     --modal-bg-color: var(--dt-color-container-default);
     --modal-elevation: var(--dt-elevation-04);
 
+    &:focus-visible {
```
ashleynolan, line 25: I didn't notice this the first time around, but the ordering of definitions here makes it pretty hard to scan (imo).

I'm pretty relaxed on ordering for the most part, but I do think we should keep to the structure of:

- Custom props defined at the start of a block
- Any block-specific style declarations
- Media queries for that block
- and then any other descendent or pseudo-selectors after that

Could you put in a PR to move this around based on that?

PR #625, opened 2023-07-17
feat(pie-monorepo): DSW-801 jira ticket ids automation

fernandofranca:
- Adds verification of the branch name format
- Adds verification and automation for adding the Jira ticket id to the commit message
- Sets up GitHub workflow for checking the PR title for the Jira ticket id

### Branch name

The branch name must have the Jira ticket id prepended in lowercase, followed by a dash. Example: `dsw-123-branch-name`, `dsw-1234-branch-name`.

If the branch doesn't have an associated Jira ticket, it must be prepended with `dsw-000`. Example: `dsw-000-branch-name`.

This is validated by the `pre-commit` and `pre-push` git hooks. If the branch name doesn't follow the pattern, the commit or push will be aborted.

### Commit message

The commit message must contain the Jira ticket in uppercase, appended between the `type(scope)` and the commit subject. Example: `type(scope): DSW-123 your commit message`.

Commitizen now validates the presence of the ticket id in the commit message. By consequence, it will be asked in `yarn cz` prompt, and it can be automatically filled in by simply typing 0 as the scripts will infer the ticket id from the branch name.

This is validated by the `commit-msg` hook. As mentioned above, if the commit message doesn't contain a Jira ticket id, it will be inferred from the branch name and added to the commit message.

The same applies to manually added commits. E.g.: Running `git commit -m "feat(pie-monorepo): your commit message"` will have the commit message automatically updated to `feat(pie-monorepo): DSW-123 your commit message`.

Note: Originally the idea was to have this completely automated, however, the `commitlint` prompt is not as flexible, and in order to make it understand that commit messages now have a Jira ticket id, it has to ask for it in the prompt. [commitlint docs link](https://commitlint.js.org/#/reference-prompt). If we consider it too annoying we can keep the validation part and create our own prompt for replacing the standard one.

### PR title

The PR title must contain follow the same pattern as the branch name. Example: `type(scope): DSW-123 your PR title`.

A GitHub action will validate the title when the PR is created or after direct editing. It will add and update a comment showing the validation status, as well as showing a check in the PR checks (see screenshots). This can be useful for guiding new contributors.

The Github action uses a regex that also matches other patterns to accommodate for exceptions like `Version Packages` and `[Snyk]`. The regex can be easily updated in the GitHub workflow file `pr-title-linter.yml`.

## cz script optimization

This is unrelated to the task itself but it helped me a lot during development and I believe it will be useful for all of us since it helps to cut time on every occasion we use the `yarn cz` script.

The idea to optimize this came in a conversation with @siggerzz. `.cz-config.js` uses a Turborepo script to list all monorepo packages, but it takes around 5~10 seconds to run. It's not a huge issue but it becomes inconvenient when running `yarn cz` multiple times during development.

I've updated the script to use a glob pattern instead, which is faster. Now it takes around 1 second to run.

## Screenshots

Updated Commitlint prompt:

image

Added PR title check (success/error):

image

image

PR title status comment (success/error):

image

image


.cz-config.js
```diff
@@ -1,21 +1,25 @@
-const { execSync } = require('child_process');
+const fs = require('fs')
+const { globSync } = require('glob')
+
 
 const getPackages = () => {
-  let outputPackages;
+    const ignore = [
+        'apps/examples/**',
```
sebastienrocks, line 7: Why do we need to ignore `apps/examples/**`?

fernandofranca:
My bad! I forgot to update the branch after requesting review 🤦

.cz-config.js
fernandofranca: I'm not sure, but I believe it's being filtered for convenience since we don't update them as often... Likely @siggerzz  can help clarify that.

jamieomaguire:
Screenshot 2023-07-21 at 12 35 04
@fernandofranca I noticed already it seems to run git secrets twice on commit.

Does this PR make it run 3 times now? If so would be really good to try and prevent if possible. If its a lot of work we could do as part of a separate ticket.

.github/workflows/pr-title-linter.yml
```diff
@@ -0,0 +1,19 @@
...
+    steps:
+      - uses: morrisoncole/pr-lint-action@v1.7.0
+        with:
+          repo-token: "${{ secrets.GITHUB_TOKEN }}"
+          title-regex: "^[a-z]+?\\([a-z0-9-]+?\\): [A-Z]+?-\\d+? \\w.*|^Version Packages|^\\[Snyk\\].*"
+          on-failed-regex-fail-action: true
+          on-failed-regex-create-review: true
+          on-failed-regex-request-changes: true
+          on-failed-regex-comment: "[Lint PR Title] ❌ Failed.\nThe PR title should follow the same convention used for commits, e.g.:\n`type(scope): DSW-123 title` where `DSW-123` is the Jira ticket id or `DSW-000` when one is not available."
+          on-succeeded-regex-dismiss-review-comment: "[Lint PR Title] ✅ All good!"
```
jamieomaguire, line 19: Do we really need to add a comment on success? Feels a bit noisy

scripts/git-hooks/__tests__/git-hooks-utils.test.js
```diff
@@ -0,0 +1,103 @@
...
+                const expected = null;
+
+                expect(getTicketIdFromBranchName('random-string')).toBe(expected);
+                expect(getTicketIdFromBranchName('jira000-branch-name')).toBe(expected);
+            });
+        });
+    });
+
+    describe('when called with a valid argument', () => {
+        it('should return the ticket id', () => {
```
jamieomaguire, line 30:
```suggestion
        it('should return the ticket id in upper case', () => {
```

.github/workflows/pr-title-linter.yml
fernandofranca: In fact, it only updates the existing one, so no one gets spammed.

scripts/git-hooks/__tests__/git-hooks-utils.test.js
fernandofranca: Good point!

scripts/git-hooks/__tests__/git-hooks-utils.test.js
```diff
@@ -0,0 +1,103 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
```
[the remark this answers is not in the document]
fernandofranca: Well noted!

jamieomaguire:
Everything looks great! Just left a few small comments/questions :)

fernandofranca:
@jamieomaguire [That](https://github.com/justeattakeaway/pie/pull/625#issuecomment-1645444319) would be better, but I'm not trying to fix it. 

I checked how much time it consumes, and it's at most 1 sec. Most of the processing time in fact is spent **once** by `commitlint` in `commit-msg` hook.

image

.github/workflows/pr-title-linter.yml
jamieomaguire: I appreciate it's only 1 comment but do we really need one if we can already see the check is green?

jamieomaguire: It just feels a little unnecessary and doesn't really add anything useful

.cz-config.js
ashleynolan: Yeh, that's right – we don't add changesets for the example apps, we add a changeset to the root monorepo for these changes (mainly as they start to clog up the commitizen UI otherwise with lots of apps that we only use for testing atm).

.github/workflows/pr-title-linter.yml
ashleynolan: I think this only shows if it's previously failed - it won't show if it passes first time (looking at the github action).

It looks like if you don't customise this, it'll update it anyway and just display `All good!` so I don't think you can get rid of this behaviour.

.cz-config.js
```diff
@@ -1,21 +1,25 @@
-const { execSync } = require('child_process');
+const fs = require('fs')
```
ashleynolan, line 1: This change makes me so happy btw – the slight delay every time I committed made a little part of me die each time 🤣

package.json
```diff
@@ -14,6 +14,7 @@
     "packages": [
       "apps/**/*",
       "configs/**/*",
+      "scripts/**/*",
```
ashleynolan, line 17: I'd be tempted to move this to `packages/tools`, as this seems like it fits in that category of packages potentially (and keen not to have too many folders at the root of the repo).

Could even use `packages/tools/scripts` if we think we're likely to have more than one github action script there in the future

ashleynolan (commented):
Couple of comments, but otherwise looks good 👍

package.json
fernandofranca: I think that also makes sense. 👍 

I don't know if it's a concern only I have but I wished we had a clearer separation in directories for consumer packages and internal ones.

.github/workflows/pr-title-linter.yml
fernandofranca: We have the option of not displaying it at all. If it fails one can check the job output.

I'd say we can turn it off in the near future while we all adapt to the new pattern.

image

package.json
ashleynolan: I don't think we've looked at it yet, but if we wanted to rearrange directories to reflect this in the future we could do that.

If you've got any opinions on how it could work, might be worth opening a Github discussion and can think it through there?

sebastienrocks (approved):

raoufswe (approved):

PR #632, opened 2023-07-19
feat(pie-modal): Add leadingAction prop

kevinrodrigues:

add prop for leadingAction

### Reviewer 1

### Reviewer 2


packages/components/pie-modal/src/index.ts
```diff
@@ -63,6 +63,13 @@ export class PieModal extends RtlMixin(LitElement) implements ModalProps {
     @validPropertyValues(componentSelector, sizes, 'medium')
     public size: ModalProps['size'] = 'medium';
 
+    @property()
+    public leadingAction!: {
```
ashleynolan, line 67: What does the `!` do here (sorry for the noob questions!)

packages/components/pie-modal/src/defs.ts
```diff
@@ -56,6 +56,27 @@ export interface ModalProps extends RTLComponentProps {
...
+        /**
+         * The text to display for the leading action button.
+         */
+        text: string;
+
+        /**
+         * The variant of the leading action button.
+         * Default: 'primary'
+         */
+        variant?: string;
```
jamieomaguire, line 73: I think we should utilise the PieButton prop type for variant here. We certainly want some validation on the variant.

packages/components/pie-modal/src/defs.ts
```diff
@@ -56,6 +56,27 @@ export interface ModalProps extends RTLComponentProps {
      * The size of the modal; this controls how wide it will appear on the page.
      */
     size: typeof sizes[number];
+
+    /**
+     * The leading action configuration for the modal.
+     */
+    leadingAction?: {
```
jamieomaguire, line 63: I would potentially create a type for action to reuse between this and the upcoming trailing action

packages/components/pie-modal/src/index.ts
```diff
@@ -211,6 +218,31 @@ export class PieModal extends RtlMixin(LitElement) implements ModalProps {
...
+     * @param action
+     * @private
+     */
+    private renderLeadingAction (action: { text: string; variant?: string; ariaLabel?: string }): TemplateResult {
+        const { text, variant = 'primary', ariaLabel } = action;
+
+        return html`
+              <pie-button
+                  variant="${variant}"
+                  aria-label="${ariaLabel}"
```
jamieomaguire, line 237: what happens to the aria label attribute if no aria-label is provided?

jamieomaguire, line 237: have you tested that screen readers narrate the aria label correctly when applied to pie button?

packages/components/pie-modal/src/index.ts
```diff
@@ -211,6 +218,31 @@ export class PieModal extends RtlMixin(LitElement) implements ModalProps {
...
+     * Render leadingAction button depending on prop availability.
+     *
+     * 1. If the prop `leadingAction` is not provided, the button is not rendered.
+     * 2. If the prop `leadingAction` is provided but any of the optional properties
+     * are not provided, they fall back to their default values.
+     *
+     * @param action
+     * @private
+     */
+    private renderLeadingAction (action: { text: string; variant?: string; ariaLabel?: string }): TemplateResult {
```
jamieomaguire, line 231: Can we just access `this.leadingAction` rather than passing it into the render function? It's a class method so I would have thought it's in scope

apps/pie-storybook/stories/pie-modal.stories.ts
```diff
@@ -136,7 +144,8 @@ const BaseStoryTemplate = (props: ModalProps): TemplateResult => {
             returnFocusAfterCloseSelector="${ifDefined(returnFocusAfterCloseSelector)}"
             ?isOpen="${isOpen}"
             dir="${dir}"
-            size="${size}">
+            size="${size}"
+            .leadingAction="${leadingAction}">
```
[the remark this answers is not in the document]
kevinrodrigues: It will bind the to the prop leadingAction to itself. There's a bit here about it: https://lit.dev/docs/v1/components/templates/#bind-properties-to-templated-elements

packages/components/pie-modal/src/index.ts
kevinrodrigues: NPs. 

The `!` is a typescript thing. What it does is forces a property to always have a value. By default TS makes properties optional so they fall back to `undefined` or `null`. By adding the `!` it forces the compiler to think that the property will always have a non-null value (I think of it as filtering/excluding out the null values)

packages/components/pie-modal/test/visual/pie-modal.spec.ts
```diff
@@ -149,30 +154,89 @@ test.describe('`hasBackButton`', () => {
...
-test('long heading renders correctly', async ({ page, mount }) => {
-    await mount(PieModal, {
-        props: {
-            heading: 'This is a modal heading but super long and should span multiple lines, hopefully this should never happen on production!',
-            isOpen: true,
-            size: 'medium',
-            hasBackButton: true,
-            isDismissible: true,
-        },
+test.describe('Prop: `heading`', () => {
```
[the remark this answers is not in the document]
kevinrodrigues: Ideally, we should wrap all related props in just one describe for `Props`. For now I've just updated them individually.

.changeset/fifty-geckos-count.md
```diff
@@ -0,0 +1,7 @@
+---
+"@justeattakeaway/pie-modal": minor
+"pie-storybook": minor
+"@justeattakeaway/pie-wrapper-react": patch
+---
+
+add prop for leadingAction
```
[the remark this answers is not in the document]
kevinrodrigues: Fixed.

packages/components/pie-modal/src/index.ts
kevinrodrigues: Fixed.

packages/components/pie-modal/test/visual/pie-modal.spec.ts
```diff
@@ -149,30 +154,89 @@ test.describe('`hasBackButton`', () => {
...
+                },
+            });
+
+            await percySnapshot(page, 'Modal fallback to default property `primary`');
+        });
     });
 
-    await percySnapshot(page, `Modal displays loading spinner - isLoading: ${true}`);
+    test.describe('when prop is NOT passed into component', () => {
+        test('should NOT display `leadingAction`', async ({ mount, page }) => {
```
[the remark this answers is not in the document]
kevinrodrigues: Fixed.

packages/components/pie-modal/src/defs.ts
kevinrodrigues: Yeah I was thinking this when adding it. This ticket actually suggested working on both tickets at the same time so i was hoping to pick that up next and then change this. Would that be ok if i left it as it is now then add types in the second ticket?

packages/components/pie-modal/src/index.ts
kevinrodrigues: Fixed.

kevinrodrigues: Hadn't checked but have now and seems to work fine for me using voiceover.

packages/components/pie-modal/README.md
```diff
@@ -72,6 +72,7 @@ import { PieModal } from '@justeattakeaway/pie-modal/dist/react';
 | isLoading | `Boolean` | `false` | When true, displays a loading spinner in the modal |
 | size | `String` | `medium` | Determines the maximum width of the modal. Large modals will expand to fill the entire page width at narrow viewports. Can be `small`, `medium` or `large`. |
 | returnFocusAfterCloseSelector | `String` | `undefined` | If provided, focus will be sent to the first element that matches this selector when the modal is closed. If not provided, the `dialog` element will return focus to the element that opened the modal. |
+| leadingAction | `Object` | `undefined` | An object representing the leading action of the modal. It has the following properties:<br>- text (required): The text to display on the leading action button.<br>- variant: The variant of the leading action button. It can be one of primary, secondary, ghost, or ghost-secondary. Defaults to primary if not provided.<br>- ariaLabel: The ARIA label for the leading action button. |
```
jamieomaguire, line 75: Its worth bearing in mind we will need to manually keep this up to date with the button variant type in this readme. I won't if a separate prop table would be useful for object properties like this. Otherwise it's quite easy to miss. Eventually I think we'd want to automate this

packages/components/pie-modal/src/defs.ts
kevinrodrigues: Fixed.

packages/components/pie-modal/src/index.ts
ashleynolan: Makes sense – does this mean that `leadingAction` must always have a value then? Or just that it should always receive an object (whether empty or with values).

Just wondering how this works for when we don't want to display a `leadingAction`

packages/components/pie-modal/README.md
ashleynolan: I think it would be better to just link to the prop table for pie-button here:

I've added a suggestion below

packages/components/pie-modal/README.md
```diff
@@ -73,6 +73,7 @@ import { PieModal } from '@justeattakeaway/pie-modal/dist/react';
 | size | `String` | `medium` | Determines the maximum width of the modal. Large modals will expand to fill the entire page width at narrow viewports. Can be `small`, `medium` or `large`. |
 | position | `String` | `center` | The position of the modal; this controls where it will appear on the page. Can be `top` or `center`. |
 | returnFocusAfterCloseSelector | `String` | `undefined` | If provided, focus will be sent to the first element that matches this selector when the modal is closed. If not provided, the `dialog` element will return focus to the element that opened the modal. |
+| leadingAction | `Object` | `undefined` | An object representing the leading action of the modal. It has the following properties:<br>- text (required): The text to display on the leading action button.<br>- variant: The variant of the leading action button. It can be one of primary, secondary, ghost, or ghost-secondary. Defaults to primary if not provided.<br>- ariaLabel: The ARIA label for the leading action button. |
```
ashleynolan:
```suggestion
| leadingAction | `Object` | `undefined` | An object representing the leading action of the modal. It has the following properties:<br>- text (required): The text to display on the leading action button.<br>- variant: The variant of the leading action button. For a list of variants, see the [pie-button documentation](https://github.com/justeattakeaway/pie/tree/main/packages/components/pie-button#props). Defaults to primary if not provided.<br>- ariaLabel: The ARIA label for the leading action button. |
```

apps/pie-storybook/stories/pie-modal.stories.ts
```diff
@@ -144,6 +152,7 @@ const BaseStoryTemplate = (props: ModalProps): TemplateResult => {
             ?isOpen="${isOpen}"
             dir="${dir}"
             size="${size}"
+            .leadingAction="${leadingAction}"
```
sebastienrocks, line 155: Just for my learning: what is the `.` here for?

packages/components/pie-modal/README.md
```diff
@@ -61,18 +61,19 @@ import { PieModal } from '@justeattakeaway/pie-modal/dist/react';
...
+| returnFocusAfterCloseSelector | `String` | `undefined` | If provided, focus will be sent to the first element that matches this selector when the modal is closed. If not provided, the `dialog` element will return focus to the element that opened the modal.                                                                                                                                                            |
+| leadingAction | `Object` | `undefined` | An object representing the leading action of the modal. It has the following properties:<br>- text (required): The text to display on the leading action button.<br>- variant: The variant of the leading action button. Please see [pie-button](https://github.com/justeattakeaway/pie/blob/main/packages/components/pie-button/README.md#props) for options. Defaults to primary if not provided.<br>- ariaLabel: The ARIA label for the leading action button. |
```
sebastienrocks, line 76: Is it possible here to add an example?

```
leadingAction = {
  text: "Example Text",
  variant: "secondary",
  ariaLabel: "Example Aria Label"
}
```

sebastienrocks (approved):

apps/pie-storybook/stories/pie-modal.stories.ts
kevinrodrigues: There's a bit about it here and a reference to the doc 👍 https://github.com/justeattakeaway/pie/pull/632#discussion_r1268568805

packages/components/pie-modal/README.md
kevinrodrigues: Fixed.

packages/components/pie-modal/src/index.ts
kevinrodrigues: So for displaying the `leadingAction` if it isn't provided then nothing renders. The `!` is only specific to the TS compiler itself :)

sebastienrocks (dismissed):

ashleynolan (dismissed):

fernandofranca (dismissed):

packages/components/pie-modal/README.md
```diff
@@ -61,18 +61,19 @@ import { PieModal } from '@justeattakeaway/pie-modal/dist/react';
...
+| Property | Type | Default | Description                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  |
+|---|---|---|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
```
jamieomaguire, line 65: is this necessary?

packages/components/pie-modal/src/defs.ts
```diff
@@ -59,6 +60,27 @@ export interface ModalProps extends RTLComponentProps {
     size: typeof sizes[number];
 
     /**
+     * The leading action configuration for the modal.
+     */
+    leadingAction: {
+        /**
+         * The text to display for the leading action button.
+         */
+        text: string;
```
jamieomaguire, line 69: should text have the `?` too?

packages/components/pie-modal/src/index.ts
```diff
@@ -216,6 +223,30 @@ export class PieModal extends RtlMixin(LitElement) implements ModalProps {
...
+     *
+     * @private
+     */
+    private renderLeadingAction (): TemplateResult {
+        const { text, variant = 'primary', ariaLabel } = this.leadingAction;
+
+        return html`
+            <pie-button
+                  variant="${variant}"
+                  aria-label="${ariaLabel || nothing}"
```
jamieomaguire, line 241: Does this narrate correctly when its applied to the host?

packages/components/pie-modal/test/visual/pie-modal.spec.ts
```diff
@@ -143,32 +143,91 @@ test.describe('`hasBackButton`', () => {
...
+                    hasBackButton: true,
+                    isDismissible: true,
+                    isOpen: true,
+                    leadingAction: {
+                        text: 'Confirm',
+                    },
+                } as ModalProps,
+            });
+
+            await percySnapshot(page, 'Modal fallback to default property `primary`');
```
jamieomaguire, line 213:
```suggestion
            await percySnapshot(page, 'Modal falls back to default property `primary`');
```

jamieomaguire:
Screenshot 2023-07-21 at 13 27 56

@kevinrodrigues should I be able to pass an empty string into the leading action text?

packages/components/pie-modal/src/index.ts
kevinrodrigues: Do you mean via screen reader? I did test it on voice over and it was working fine.

apps/pie-storybook/stories/pie-modal.stories.ts
```diff
@@ -31,6 +31,10 @@ const defaultArgs: ModalProps = {
     position: 'center',
     slot: 'This is Lit!',
     dir: 'ltr',
+    leadingAction: {
+        text: 'Confirm',
+        variant: 'primary',
```
jamieomaguire, line 36: please add aria label to this object otherwise it isn't clear when visiting storybook that this is a property we can pass in

kevinrodrigues: Fixed.

packages/components/pie-modal/README.md
kevinrodrigues: Fixed.

packages/components/pie-modal/test/visual/pie-modal.spec.ts
kevinrodrigues: Fixed.

kevinrodrigues:
> Screenshot 2023-07-21 at 13 27 56
> @kevinrodrigues should I be able to pass an empty string into the leading action text?

Good spot. So there's a couple things here. 

1. I think we should use the `@validPropertyValues` like we discussed for the text specific value so it errors in the console - i don't think that will work as it is now tho.
2. I've updated it so the template renders nothing if text isn't supplied, the other option would have been to fall back to another default bit of text like "confirm".

ashleynolan (approved):

fernandofranca (approved):

packages/components/pie-modal/src/defs.ts
```diff
@@ -59,6 +60,27 @@ export interface ModalProps extends RTLComponentProps {
...
+        /**
+         * The variant of the leading action button.
+         * Default: 'primary'
+         */
+        variant?: Variant;
+
+        /**
+         * The ARIA label for the leading action button.
+         */
+        ariaLabel?: string;
```
raoufswe, line 80: I think it makes more sense for us to pass an aria object instead of a string primitive here as aria properties are not limited to a label only. for example, the button should accept the following aria attributes to provide means for an accessible button component (referenced from snacks button)
```
     aria-disabled={disabled}
      aria-label={aria.label}
      aria-labelledby={aria.labelledby}
      aria-describedby={aria.describedby}
      aria-expanded={aria.expanded}
      aria-controls={aria.controls}
      aria-live="polite"
      aria-busy={aria.loading ? 'true' : 'false'}
      aria-atomic="true"
```

```
  aria: {
    label: string,
    labelledby: string,
    describedby: string,
    loading: boolean,
    expanded: boolean,
    controls: string,
  },
```

this will require an update to the button itself as well, so feel free to address it in another PR if that makes more sense.

raoufswe (approved):

packages/components/pie-modal/src/index.ts
xander-marjoram: Does it actually force it to have a value or just tell TS that it should? Do we get console errors if there is no value for `leadingAction`?

packages/components/pie-modal/src/index.ts
```diff
@@ -255,13 +291,7 @@ export class PieModal extends RtlMixin(LitElement) implements ModalProps {
...
             </article>
             <footer class="c-modal-footer">
-                <pie-button
-                    variant="primary"
-                    type="submit"
-                    @click="${() => this._dialog?.close('leading')}"
-                    data-test-id="modal-leading-action">
-                    Confirm
-                </pie-button>
+                ${leadingAction ? this.renderLeadingAction() : nothing}
```
xander-marjoram, line 294: `renderLeadingAction` can already return nothing so can we just have the function call here, or simplify the function instead?

packages/components/pie-modal/src/index.ts
kevinrodrigues: Just TS. We won't get console errors if it isn't supplied.

packages/components/pie-modal/src/index.ts
kevinrodrigues: It would only return nothing if `text` isn't supplied otherwise it would return a button. The `nothing` on this particular line covers  the `leadingAction` prop. I could move it all into a method and just have a function here?

xander-marjoram: Actually I think you're right about us needing `nothing` in both places, all good.

packages/components/pie-modal/src/index.ts
```diff
@@ -216,6 +223,34 @@ export class PieModal extends RtlMixin(LitElement) implements ModalProps {
...
+     * Render leadingAction button depending on prop availability.
+     *
+     * 1. If the prop `leadingAction` is not provided, the button is not rendered.
+     * 2. If the prop `leadingAction` is provided but any of the optional properties
+     * are not provided, they fall back to their default values.
+     *
+     * @private
+     */
+    private renderLeadingAction (): TemplateResult | typeof nothing {
+        const { text = 'Confirm', variant = 'primary', ariaLabel } = this.leadingAction;
```
xander-marjoram, line 236: I think having this default value of `"Confirm"` means the button still renders if no text is passed in?

packages/components/pie-modal/test/visual/pie-modal.spec.ts
```diff
@@ -143,32 +183,119 @@ test.describe('`hasBackButton`', () => {
...
-            isOpen: true,
-            isLoading: true,
-        } as ModalProps,
+test.describe('Prop: `leadingAction`', () => {
+    test.describe('when prop is passed into component', () => {
+        test('should display `leadingAction` when props are passed correctly', async ({ mount, page }) => {
+            await mount(PieModal, {
+                props: {
+                    heading: 'This is a modal heading',
+                    hasBackButton: true,
```
xander-marjoram, line 234: I don't think we need `hasBackButton` or `isDismissible` in any of these tests, those should be covered by our other tests

packages/components/pie-modal/src/index.ts
kevinrodrigues: Ah yes that's my bad. I'll remove in the next pr.

PR #637, opened 2023-07-21
feat(pie-docs): added new header design

LTurns:

### Reviewer 1

### Reviewer 2


apps/pie-docs/src/_includes/home-page.njk
```diff
@@ -0,0 +1,30 @@
...
+                </div>
+                <div class="u-showAboveMid c-homePage-image">
+                    <img src={{ heroSrc }} alt="" />
+                </div>
+            </div>
+        </div>
+    </div>
+    <div class="razor"></div>
+    <div class="c-homePage-content">
+        <div class="c-contentPage-content">
```
ashleynolan, line 24: The class naming here is a bit confusing – using `c-contentPage-content` when this isn't a `contentPage`.

The intention would be that this class would be a child of `c-contentPage` but this is a child of `c-homePage`, so may need a bit of renaming here to make this a bit clearer. I'd personally update them so that they are less page dependent (maybe if this is styling content across pages, it could be named `c-content` and then `c-content-container` makes sense being used here too.

apps/pie-docs/src/_includes/home-page.njk
```diff
@@ -0,0 +1,30 @@
+{% set heroSrc = ['/assets/img/heroes/', heroImageSrc] | join | url %}
+
+<div class="c-homePage" data-test-id="site_content">
+    <div class="c-homePage-header">
+        <div class="c-content-container">
```
ashleynolan, line 5: Same here with the classnames – I think these need to be tidied up a little to be a bit clearer.

apps/pie-docs/src/assets/styles/components/_homePage.scss
```diff
@@ -0,0 +1,123 @@
+@use '../fozzie-config' as f;
+
+.c-homePage {
```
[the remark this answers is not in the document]
ashleynolan: Just for some ideas looking at the preview, I'm guessing some of this relates to the homepage banner, so you could call that component `c-heroBanner`.

And for content straight after a `c-heroBanner` (as it looks like you're adding styles to the content part of the page based on if there's a heroBanner or not) you could use `c-content` but use sibling selectors – `c-heroBanner + .c-content`.

If it'd be easier to pair on this, let me know – I know naming isn't always the easiest thing to think of!

ashleynolan (changes requested):
Would be good to tidy up the classnames - they're a bit mixed up atm

LTurns:
closed due to updating branch name to include Jira ticket

PR #643, opened 2023-07-24
feat(pie-modal): DSW-931 added supportingAction prop

kevinrodrigues:

[Added] SupportingAction prop to modal

### Reviewer 1

### Reviewer 2


packages/components/pie-modal/src/index.ts
```diff
@@ -233,7 +240,7 @@ export class PieModal extends RtlMixin(LitElement) implements ModalProps {
      * @private
      */
     private renderLeadingAction (): TemplateResult | typeof nothing {
```
raoufswe, line 242: i might be missing something here, but don't we want to allow users to trigger an on click callballback for the action buttons here? they both close the modal at the moment.

packages/components/pie-modal/src/index.ts
```diff
@@ -251,6 +258,41 @@ export class PieModal extends RtlMixin(LitElement) implements ModalProps {
...
+
+        if (!this.leadingAction) {
+            console.warn('Use `leadingAction` instead of `supportingAction`. `supportingAction` is being ignored.');
+            return nothing;
+        }
+
+        return html`
+              <pie-button
+                variant="${variant}"
+                aria-label="${ariaLabel || nothing}"
```
raoufswe, line 287: can we use ifDefined here?
https://lit.dev/docs/templates/directives/#ifdefined

kevinrodrigues: I tried a few things before i went this route initially. The issue is with `ifDefined` and `guard` they still render the attribute itself if the value is empty i.e `ariaLabel: ''`. With `nothing` if the ariaLabel is empty or doesn't exist then the attribute won't render.

packages/components/pie-modal/src/index.ts
kevinrodrigues: Not sure if there is a ticket for this but the markup that was there before both closed the dialog, I'm guessing this is just intentional until we have a tickets for this work. @xander-marjoram @fernandofranca do you know if we have one? If not i can add the changes here.

fernandofranca (approved):

PR #648, opened 2023-07-25
feat(pie-modal): DSW-932 add isFooterPinned prop

jamieomaguire:
---
"pie-storybook": minor
---

[Added] - isFooterPinned property to modal which offers more scrolling options to consumers

Example: isFooterPinned = false
Screenshot 2023-07-27 at 16 58 42

Example: isFooterPinned = false (scrolled to bottom)
Screenshot 2023-07-27 at 16 59 02

Example: isFooterPinned = true
Screenshot 2023-07-27 at 16 58 54


packages/components/pie-modal/src/defs.ts
```diff
@@ -44,6 +44,11 @@ export interface ModalProps extends RTLComponentProps {
      */
     hasBackButton: boolean;
 
+    /**
+     * When true, the modal footer will scroll with the content inside the modal body.
```
ashleynolan, line 48:
```suggestion
     * When false, the modal footer will scroll with the content inside the modal body.
```

apps/pie-storybook/stories/pie-modal.stories.ts
```diff
@@ -24,6 +24,7 @@ const defaultArgs: ModalProps = {
     headingLevel: 'h2',
     isDismissible: true,
     hasBackButton: true,
+    isFooterPinned: true,
```
kevinrodrigues, line 27: Could you add this to the readme please.

packages/components/pie-modal/src/modal.scss
```diff
@@ -137,18 +151,31 @@
...
         padding: var(--modal-footer-padding);
 
         @media (min-width: $breakpoint-wide) {
             --modal-footer-padding: var(--dt-spacing-e);
         }
     }
 
+
+    & .c-modal-scrollContainer,
+    &[isfooterpinned] .c-modal-content {
```
sebastienrocks, line 172: Are we not creating a double scrollbar with this?

Does this not result in:

<div class=".c-modal-scrollContainer"> -> Scrollable
 <div isFooterPinned>
  <div class="c-modal-content"></div> -> Scrollable
  </div>
</div>

packages/components/pie-modal/src/modal.scss
```diff
@@ -48,10 +49,23 @@
...
-        flex-direction: column;
+        display: grid;
+    }
+
+    grid-template-areas:
+        'header'
+        'content';
+    grid-template-rows: min-content minmax(var(--modal-content-min-height), 1fr);
+
+    &[isfooterpinned] {
```
ashleynolan, line 60: Nit – Can you move this to be after the other base styling declarations? (so somewhere after line 84)

Not sure if it's just me but I like to have all the base styles together before hitting any modifiers (just makes it easier to scan the base styles)

ashleynolan (commented):
Looks good – some super small comments but otherwise LGTM

packages/components/pie-modal/test/visual/pie-modal.spec.ts
```diff
@@ -414,3 +414,51 @@ test.describe('`position`', () => {
         });
     });
 });
+
+test.describe('Prop: `isFooterPinned`', () => {
```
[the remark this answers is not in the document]
jamieomaguire: perhaps yeah - I think we should chat with a few others about it

packages/components/pie-modal/src/index.ts
```diff
@@ -328,15 +351,16 @@ export class PieModal extends RtlMixin(LitElement) implements ModalProps {
...
-            <article class="c-modal-content c-modal-content--scrollable">
-                <div class="c-modal-contentInner">
-                    <slot></slot>
-                </div>
-            </article>
-            <footer class="c-modal-footer">
-                ${leadingAction ? this.renderLeadingAction() : nothing}
-                ${supportingAction ? this.renderSupportingAction() : nothing}
-            </footer>
+            ${
```
kevinrodrigues, line 354: You could potentionally shorten this and make use of the isFooterPinned prop to render a class or a generic class instead. That way you dont require the ternary and can have a shorter bit of code.
```html
<div class=${isFooterPinned ? "c-modal-footer-pinned" : "c-modal-scrollContainer"}>
  ${this.renderModalContentAndFooter()}
</div>

Comment (jamieomaguire) [reply packages/components/pie-modal/src/index.ts:354]: I was trying to avoid a needless div! What would `c-modal-footer-pinned` be used for in this case?

File: packages/components/pie-modal/src/index.ts [hunk 7]
```diff
@@ -317,7 +338,9 @@ export class PieModal extends RtlMixin(LitElement) implements ModalProps {
             size="${size}"
             position="${position}"
             ?hasBackButton=${hasBackButton}
+            ?hasActions=${!!leadingAction || !!supportingAction}
```
kevinrodrigues, line 341: Could you try the below instead.

`?hasActions=${ifDefined(leadingAction || supportingAction)}`

ashleynolan (dismissed):

packages/components/pie-modal/test/visual/pie-modal.spec.ts
```diff
@@ -414,3 +414,51 @@ test.describe('`position`', () => {
...
+                    isOpen: true,
+                    isFooterPinned,
+                    leadingAction: {
+                        text: 'Confirm',
+                        variant: 'primary',
+                        ariaLabel: 'Confirmation text',
+                    },
+                } as ModalProps,
+                slots: {
+                    default: `Lorem ipsum dolor sit amet consectetur adipisicing elit. Deleniti fugit id exercitationem repellendus in magni
```
kevinrodrigues, line 433: We might want to look at generating this. I know there are tools like faker or lorem ipsum that allow you to pull in content via a simple object so we wont have lots of fake data in our tests.

packages/components/pie-modal/src/index.ts
jamieomaguire: Screenshot 2023-07-25 at 15 43 38
@kevinrodrigues does this seem right?

jamieomaguire: Screenshot 2023-07-25 at 15 44 17

@kevinrodrigues does this seem right?

packages/components/pie-modal/src/modal.scss
jamieomaguire: If `isFooterPinned` is false, the markup is:

```html
<dialog class="c-modal">
    <header class="c-modal-header">
    <!--  -->
    </header>
    <div class="c-modal-scrollContainer">  <-- this is the only element with a scroll bar
        <article class="c-modal-content c-modal-content--scrollable">
        <!--  -->
        </article>
        <footer class="c-modal-footer">
        <!--  -->
        </footer>
    </div>
</dialog>
```

When `isFooterPinned` is true,
```html
<dialog class="c-modal">
    <header class="c-modal-header">
    <!--  -->
    </header>
    <article class="c-modal-content c-modal-content--scrollable">   <-- this is the only element with a scroll bar
    <!--  -->
    </article>
    <footer class="c-modal-footer">
    <!--  -->
    </footer>
</dialog>
```

packages/components/pie-modal/src/index.ts
kevinrodrigues: Fair enough. More of a suggestion. I think the bigger this component gets the harder it is going to be to follow / read. Maybe even moving it out to a separate method would help and then rendering that method as a single thing here would just make it clearer.

jamieomaguire: I do agree, but I'm unsure if adding another layer of abstraction (i.e. another method) would add any coherence here. I'd prefer to go with this for now and optimize once all the markup is added to the modal and we have a clearer understanding of the final 'shape' of the html

packages/components/pie-modal/src/index.ts
kevinrodrigues: It doesn't look right. On second thoughts and out of interest why do you need the coerson boolean conversion? Doesn't ${leadingAction || supportingAction} work?

jamieomaguire: Removing the `!!` would probably work but I felt this was clearly testing the truthiness and the way you suggested made it look a bit more like the props themselves could be booleans

kevinrodrigues: Our props are named so indicate if it is a boolean or not `isFooBar` = Boolean vs `leadingAction` = something else (object). I dont think the conversion should be applied here.

jamieomaguire: Ok, my opinion is that the line I've added is far more explicit. It's telling you that either prop, as long as it is truthy, will meet the criteria for the attribute to be added.

It's all the benefit of `Boolean()` with less code. I think `?hasActions=${leadingAction || supportingAction}` looks as though you're trying to use the value for the attribute.

raoufswe: `leadingAction` and `supportingAction ` are optional objects. if they aren't passed, their value is undefined and I believe just checking leadingAction || supportingAction would be enough.

raoufswe: tho I would be extra defensive here as regardless if we use Boolean or !!, this will still pass if users pass an empty object `{}` and that's misleading. 

// Boolean({}) == true

I think lodash has a function called isEmpty that we can use for this case, but we don't have lodash added to our web components / alternatively, create a helper  and import it 

https://www.freecodecamp.org/news/check-if-an-object-is-empty-in-javascript/

jamieomaguire: @kevinrodrigues given this has gone on a bit I'm going to take your suggestion. Please can you re-review?

kevinrodrigues: Just had another thought.  If leadingAction isn't supplied i.e is `undefined` then `supportingAction` isn't rendered in the dom (even if it is supplied) so how does this check work in that scenario when the modal has no actions rendered but here we are saying `hasActions` gets a value back from either leadingAction or supportingAction?

fernandofranca: My 2 cents: those expressions not so easy to read could be extracted to a constant. It also makes debugging easier, since you can add a breakpoint or log the expression value.

packages/components/pie-modal/src/modal.scss
```diff
@@ -137,18 +150,31 @@
         --modal-button-spacing: var(--dt-spacing-d);
         --modal-footer-padding: var(--dt-spacing-d);
 
+        grid-area: footer;
+        min-block-size: var(--dt-spacing-e);
```
xander-marjoram, line 154: Would it be preferable to update the variable here instead?
```suggestion
        --modal-block-size: var(--dt-spacing-e);
```

xander-marjoram (dismissed):

sebastienrocks (approved):

sebastienrocks (dismissed):

LTurns (approved):

LTurns (approved):

ashleynolan (approved):

PR #650, opened 2023-07-25
feat(pie-modal): DSW-933 add hasStackedActions to pie-modal

ashleynolan:
## Describe your changes

- [Added] - hasStackedActions to pie-modal action buttons

This bumps the versions of pie-modal and pie-button, as styles were needed in both to make this prop work as required.

## Author Checklist

### Reviewer 1

### Reviewer 2


packages/components/pie-button/src/button.scss
```diff
@@ -304,3 +304,12 @@
...
 }
+
+// Ensure that any component context that sets hasStackedActions="true"
+// makes the button full width on narrow viewports
+:host-context([hasstackedactions]) .o-btn {
+    // TODO: Move breakpoint sizes into shared CSS component utilties
+    @media (max-width: 768px) {
+        inline-size: 100%;
+    }
+}
```
raoufswe, line 315: just curious: do we have a strict opinion about overriding component styles this way?

I believe our components should be flexible enough to achieve these cases without relying on custom styles when possible. For example, the button component already supports `isFullWidth`, and setting this behavior through the prop ensures that our components only change from the source and avoid any cascading/overrides.

so in theory, we could do something like 

```
<pie-button isFullWIdth={hasStackActions}></pie-button> 

```
and wrap the footer with flex columns on narrow screens

```
.hasStackedActions footer {
display:flex;
flex-direction: column
}
```

packages/components/pie-modal/test/visual/pie-modal.spec.ts
```diff
@@ -414,3 +414,33 @@ test.describe('`position`', () => {
         });
     });
 });
+
+test.describe('Prop: `hasStackedActions`', () => {
+    test.describe('when true', () => {
```
kevinrodrigues, line 419: Is there a test for false?

packages/components/pie-modal/src/index.ts
```diff
@@ -43,27 +43,23 @@ export class PieModal extends RtlMixin(LitElement) implements ModalProps {
...
 
-    @property({ type: Boolean, reflect: true })
-    public isDismissible = false;
-
     @property({ type: Boolean })
     public hasBackButton = false;
 
     @property({ type: Boolean })
-    public isFullWidthBelowMid = false;
+    public hasStackedActions = false;
```
kevinrodrigues, line 50: I think you'll need to add this to the `defs` file too.

packages/components/pie-modal/test/visual/pie-modal.spec.ts
```diff
@@ -414,3 +414,33 @@ test.describe('`position`', () => {
...
+        (['small', 'medium', 'large'] as Array<ModalProps['size']>)
+            .forEach((size) => {
+                test(`actions should be full width (at narrow viewports) for a modal with size = ${size}`, async ({ page, mount }) => {
```
kevinrodrigues, line 422: Would you mind changing this to `should ....` maybe `should display actions at full width ....`

packages/components/pie-modal/test/visual/pie-modal.spec.ts
ashleynolan: I haven't made one, as after chatting it through with a couple of others, it seems like the visual tests for `false` would be covered by every other test being run (as that's the default, so the default modal test effectively tests the `false` state for instance). So didn't want to duplicate and add another set of tests that could be unecessary.

packages/components/pie-modal/test/visual/pie-modal.spec.ts
ashleynolan: Fixed

packages/components/pie-modal/src/index.ts
ashleynolan: Fixed

packages/components/pie-modal/README.md
```diff
@@ -57,22 +57,22 @@ import { PieModal } from '@justeattakeaway/pie-modal/dist/react';
...
-| aria                          | `Object` | `undefined` | An object representing the aria labels for the `close` & `back` buttons within the modal as well as the `isLoading` state labels (`aria-label` & `aria-busy`).                                                                                                                                                                                                                                                                                                                                                                                                                        |
+| Property                      | Type      | Default        | Description |
```
ashleynolan, line 60: Just a note – most of these changes are ordering the properties to be alphabetical, so it's easier to scan them (plus adding `hasStackedActions` prop)

packages/components/pie-button/src/button.scss
jamieomaguire: I think this is going to become a more prevalent issue as we add more components where their prop-driven styling is screen-size dependent. I.e. A button needs it's full-width prop to only be true on smaller screens.

@ashleynolan @raoufswe I think this would be an excellent candidate for a GitHub discussion, would either of you be happy to set one up?

apps/pie-storybook/stories/pie-modal.stories.ts
```diff
@@ -147,40 +151,42 @@ const createFocusableElementsPageHTML = () : TemplateResult => html`
...
     return html`
         <pie-button @click=${toggleModal}>Toggle Modal</pie-button>
         <pie-modal
+            .aria="${aria}"
+            dir="${dir}"
             heading="${heading}"
             headingLevel="${headingLevel}"
-            ?isDismissible="${isDismissible}"
             ?hasBackButton="${hasBackButton}"
+            ?hasStackedActions="${hasStackedActions}"
```
jamieomaguire, line 179: I think for any property here that is not `reflected` we should use a `.` instead of a `?`. 

Using `?` is great for attributes as it means it won't be added if the condition is falsey. However for a `property` I think we'd prefer `.` which will set the value to the JS value passed in (so actual booleans or other js data types). 

I discovered this on my recent PR in which I tried using `?isFooterPinned=${isFooterPinned}` (our only default true boolean) which seemed to lead to some funky behaviour.

(I've shared a post about this internally if you want a bit more info!)

packages/components/pie-modal/README.md
jamieomaguire: In the future it would be great to automate these prop tables based off of the `custom-elements.json` file

packages/components/pie-modal/src/index.ts
jamieomaguire: I'd recommend updating `defs` first when adding new properties. That will then ensure you have to update SB and the component together thanks to typescript :)

packages/components/pie-modal/src/index.ts
```diff
@@ -102,21 +105,21 @@ export class PieModal extends RtlMixin(LitElement) implements ModalProps {
         this.addEventListener('click', (event) => this._handleDialogLightDismiss(event));
     }
 
-    connectedCallback () : void {
+    connectedCallback (): void {
```
jamieomaguire, line 108: certainly intrigued as to why the linter either:

- Doesn't flag these
- Doesn't fix these
- Accepts both the previous and new formats as valid (i.e your editor config differs from someone else's but it thinks both are fine).

It feels like we're playing whitespace ping pong between PRs at the moment 😂

packages/components/pie-modal/src/index.ts
```diff
@@ -343,8 +347,9 @@ export class PieModal extends RtlMixin(LitElement) implements ModalProps {
             class="c-modal"
             size="${size}"
             position="${position}"
-            ?hasBackButton=${hasBackButton}
             ?hasActions=${leadingAction || supportingAction}
+            ?hasBackButton=${hasBackButton}
+            ?hasStackedActions=${hasStackedActions}
             ?isDismissible=${isDismissible}
             ?isFooterPinned=${isFooterPinned}
             ?isFullWidthBelowMid=${isFullWidthBelowMid}
```
jamieomaguire, line 355: I am starting to wonder if using attributes for these vs class modifiers makes our lives easier or harder as the number of attributes grows. I suppose we'd be adding bloat either way but I'm not sure which approach makes more sense

packages/components/pie-modal/test/visual/pie-modal.spec.ts
jamieomaguire: We should probably be careful how many visual tests we add as well as it increases both CI duration and usage

jamieomaguire (approved):

kevinrodrigues (dismissed):

LTurns (approved):

packages/components/pie-button/src/button.scss
raoufswe: I took a look and I believe we can achieve the desired behavior by making the footer buttons fluid (width 100% of their parent container) and changing the direction of the footer flex container to be `column` if hasStackedAction is true for narrow screens. 


https://github.com/justeattakeaway/pie/assets/28605803/fcf03304-9cf6-4692-8999-69ec6f0a481c


I reverted the changes made to the button locally and added the following: 

```
  &[hasstackedactions] .c-modal-footer  {
        @media (max-width: 768px) {
          flex-direction: column;
        }
    }
```

and added isFullWidth="true" to both footer actions. 

this will ensure that the flex children (footer actions) will expand to span the full width of the parent container along the column direction, and still only span the space they need by default if its set to row, (currently, the default value of footer is (row-reverse).  do you reckon this would work?

packages/components/pie-modal/src/index.ts
ashleynolan: I personally prefer attributes, but it's totally just a coin-flip preference.

I don't think there's any major pro's and cons to either approach (as far as I can see anyway) but interested to know if others have any hard reasons for one or the other.

apps/pie-storybook/stories/pie-modal.stories.ts
ashleynolan: Makes sense – I've made a ticket to investigate this so we can work out the best course of action going forward with these kind of props

jamieomaguire (approved):

packages/components/pie-button/src/button.scss
ashleynolan: I'll try it out on Monday and test if it works with those changes across different browsers/widths 👍

LTurns (dismissed):

packages/components/pie-button/src/button.scss
ashleynolan: This works perfectly 🚀  I've just pushed those updates 👍

raoufswe (approved):

sebastienrocks (approved):

apps/pie-storybook/stories/pie-modal.stories.ts
```diff
@@ -159,43 +163,45 @@ const createFocusableElementsPageHTML = () : TemplateResult => html`
...
         slot,
-        dir,
-        leadingAction,
         supportingAction,
     } = props;
     return html`
         <pie-button @click=${toggleModal}>Toggle Modal</pie-button>
         <pie-modal
+            .aria="${aria}"
+            dir="${dir}"
```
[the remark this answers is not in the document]
ashleynolan: Honestly not sure (I didn't add it in this PR, I just changed the ordering). I can check it out after this is merged and see if we still need it or not.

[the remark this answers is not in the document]
ashleynolan: Honestly not sure (I didn't add it in this PR, I just changed the ordering). I can check it out after this is merged and see if we still need it or not.

kevinrodrigues (approved):

apps/pie-storybook/stories/pie-modal.stories.ts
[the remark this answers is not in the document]
kevinrodrigues: Just had a look it is required for the writing mode in SB. 👍

sebastienrocks (approved):

PR #658, opened 2023-07-28
feat(pie-toggle-switch): DSW-990 add empty package skeleton

jamieomaguire:
"@justeattakeaway/pie-toggle-switch": minor
"@justeattakeaway/pie-wrapper-react": minor
"pie-storybook": minor
"pie-monorepo": minor
---

[Added] - Created package skeleton for pie-toggle-switch component

NOTE: I've removed the test commands as they fail due to no tests existing yet. Will add them back in on the next PR and there is a ticket to set up generated tests to prevent this in future 👍

### DIFF
--- a/packages/components/pie-toggle-switch/package.json
+++ b/packages/components/pie-toggle-switch/package.json
@@ -1,0 +1,34 @@
+{
+  "name": "@justeattakeaway/pie-toggle-switch",
+  "description": "PIE Design System Toggle Switch built using Web Components",
+  "version": "0.0.0",
+  "type": "module",
+  "main": "dist/index.js",
+  "module": "dist/index.js",
+  "types": "dist/types/index.d.ts",
+  "scripts": {
+    "build": "yarn build:wrapper pie-toggle-switch && run -T vite build",
+    "lint:scripts": "run -T eslint .",
+    "lint:scripts:fix": "yarn lint:scripts --fix",
+    "lint:style": "run -T stylelint ./src/**/*.{css,scss}",
+    "lint:style:fix": "yarn lint:style --fix",
+    "watch": "run -T vite build --watch",
+    "test": "echo \"Error: no test specified\" && exit 0",
+    "test:ci": "yarn test",
+    "test:browsers": "npx playwright test -c ./playwright-lit.config.ts",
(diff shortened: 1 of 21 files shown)


raoufswe (approved):

kevinrodrigues (dismissed):

raoufswe (approved):

kevinrodrigues (approved):

raoufswe (approved):

packages/components/pie-toggle-switch/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "license": "Apache-2.0",
+  "devDependencies": {
+    "@justeattakeaway/pie-components-config": "workspace:*"
+  },
+  "dependencies": {
+    "@justeattakeaway/pie-webc-core": "workspace:*"
+  },
+  "volta": {
+    "extends": "../../../package.json"
+  }
```
ashleynolan, line 29: I think we need to add the browserslist config to our web components by the looks of it? Can't see it being referenced here.

Could you make a ticket to add this if so?

packages/tools/generator-pie-component/README.md
```diff
@@ -18,34 +18,31 @@
...
 
-Install `yeoman` and the pie-component generator globally:
+Install `yeoman` globally:
 
 ```sh
-# Via NPM
-$ npm install --global yo @justeattakeaway/generator-pie-component
-
-# Using Yarn
-$ yarn add --global yo @justeattakeaway/generator-pie-component
```
ashleynolan, line 28: Would be good to keep the yarn global option here potentially, as I'm guessing either could be used?

Also, may be worth keeping the other docs around installation for use outside our repo, as I can see other teams wanting to spin up Lit components in the future. Could have a section for use in our monorepo and one for external use?

jamieomaguire: the yarn option didn't work, so I'd suggest we create a small ticket to add the correct suggestions for yarn :) I don't see why using NPM at the OS/Global level would be an issue as it doesn't affect a project that uses yarn

To be clear, everything that was removed here was because it was incorrect/didn't work :)

PR #666, opened 2023-08-01
refactor(pie-icons-react): DSW-1007 switch to mjs to enable usage in commonjs and esm

dandel10n:
Changes to make sure the ability to use the library in projects that use commonjs or esm:

- Switches react-icons-gen.js and rollup.config.js to .mjs.
- removes `"type": "module"` from package.json and switches "main" to direct to dist instead of esm.

Tested by yalcing the package to snacks and then using snacks dev package in CW that also uses this version of pie-icons-react. CW bundle size increased only as expected.
```
├ chunks/pie.07cc7e.js                                          1.21 kB
├ chunks/snacks.415295.js                                       114 kB
```
snacks bundle in CW before icons update:
```
├ chunks/snacks.3cbf48.js                                       115 kB
```


ashleynolan (approved):

fernandofranca:
The changes look good! We just need to check how it will behave on CW in terms of bundling.

Also, @JoshuaNg2332 recently introduced updates to ESLint and file extensions. I think we need to add `.mjs` to that change.

LTurns (approved):

PR #667, opened 2023-08-01
feat(pie-icons-webc): DSW-984 Convert pie-icons-webc from native web components to Lit

xander-marjoram:
#### New PR because I broke the old one 😄 

Recommended reviewing using dev mode (press `.`) due to the large number of icon files being deleted!

>**Note**
> There is an existing issue with the webc icons where they will not apply custom classes on top of the base classes. I've raised a separate ticket to cover this (DSW-1034).

- [Fixed] - Invisible icons
- [Fixed] - Prevent tree-shaking of components in storybook
- [Removed] - Built webc icons from source control
- [Added] - Type declaration files for components
- [Added] - Types for pie-icons
- [Added] - TS version of pie-icons-configs/config.js (Will be used after DSW-1025)
- [Added] - Webc icon tests for width, height and base classes
- [Changed] - Update pie-icons-webc build to generate a slightly different template for regular and large icons (using different types, etc.)
- [Changed] - Update pie-icons-webc rollup config to remove commonjs build
- [Changed] - Use `just-kebab-case` and `just-pascal-case` instead of `kebab-case` and `pascal-case` to simplify usage (and they're more recently maintained)

### Reviewer 1

### Reviewer 2


packages/components/pie-modal/src/index.ts
```diff
@@ -8,10 +8,10 @@ import {
 } from '@justeattakeaway/pie-webc-core';
 import type { DependentMap } from '@justeattakeaway/pie-webc-core';
 import { disableBodyScroll, enableBodyScroll } from 'body-scroll-lock';
-import '@justeattakeaway/pie-icons-webc/icons/IconClose';
-import '@justeattakeaway/pie-icons-webc/icons/IconChevronLeft';
-import '@justeattakeaway/pie-icons-webc/icons/IconChevronRight';
 import { Variant } from '@justeattakeaway/pie-button/src/defs.ts';
+import '@justeattakeaway/pie-icons-webc/dist/icons/IconClose.js';
+import '@justeattakeaway/pie-icons-webc/dist/icons/IconChevronLeft.js';
+import '@justeattakeaway/pie-icons-webc/dist/icons/IconChevronRight.js';
```
raoufswe, line 14: is there a reason why we are importing the dist folder and js files here?

xander-marjoram: Yes, because it doesn't work with the TS files 😅
Ash has agreed that we can look more into this in future

raoufswe: could you please create a ticket for it so we don't lose track? 🔆

raoufswe (dismissed):

packages/components/pie-modal/src/index.ts
xander-marjoram: I don't mind doing this, but what do you think the ticket should do? IMO if we do make one it should be more of an investigation than a task.
Ultimately they're the same code, just one is TS and one is JS. Also, [Spectrum seem to import components as JS](https://github.com/adobe/spectrum-web-components/blob/main/packages/dialog/src/Dialog.ts#L28-L31) so I don't think it's a bad thing to be doing, especially as the browser will be wanting JS anyway.

ashleynolan: I'll make a ticket for this today, as I wanted to have a play around with the paths once this is merged. I'll link it here once I've created it

raoufswe: sorry if it wasn't clear, my point isn't about using js files only, more about the import itself as Ash mentioned.

raoufswe (approved):

ashleynolan (approved):

PR #671, opened 2023-08-01
feat(pie-css): DSW-1019 import design tokens and add testing

jamieomaguire:
---
"@justeattakeaway/pie-css": minor
"pie-monorepo": minor
---

[Added] - import design tokens into CSS file and add snapshot and W3C validity testing


packages/tools/pie-css/build-css.ts
```diff
@@ -0,0 +1,26 @@
...
+
+// css to be processed
+const css = fs.readFileSync('css/input.css', 'utf8');
+
+// process css
+postcss()
+    .use(atImport())
+    .process(css, {
+        // `from` option is needed here
+        from: 'css/input.css',
```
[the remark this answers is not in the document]
jamieomaguire: Sounds good, lets figure out a better structure post-merge :)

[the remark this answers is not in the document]
jamieomaguire: @ashleynolan I've moved it to a dist folder for now but we can fine-tune later on 👍

ashleynolan (approved):

xander-marjoram (approved):

packages/tools/pie-css/package.json
```diff
@@ -7,11 +7,15 @@
     "index.css",
     "scss"
   ],
-  "type": "module",
-  "main": "index.css",
+  "main": "dist/index.css",
   "scripts": {
-    "lint:style": "run -T stylelint ./**/*.{css,scss} --allow-empty-input",
-    "lint:style:fix": "yarn lint:style --fix"
+    "build:css": "run -T ts-node ./build-css.ts",
```
siggerzz, line 12: Would you mind renaming this to `build` just so it's consistent with the other packages?

Each time we deviate, it requires another entry into `turbo.json`, as well as CI changes as we currently only execute the `build` command. 👍🏼

raoufswe, line 12: do you mind sharing why we didn't use vite for this package and decided to use ts-node instead?

packages/tools/pie-css/test/css/css.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+    describe,
+    it,
+    expect,
+} from 'vitest';
+
+const readFile = util.promisify(fs.readFile);
+
+async function getCssContent (filePath: string): Promise<string> {
+    try {
+        const data = await readFile(filePath, 'utf8');
```
siggerzz, line 15: Nit: We don't need a variable here as it's not being used anywhere.

IMO it's already clear what's being returned given the function name / return type.

siggerzz (changes requested):
Looking good! Just a couple minor comments :)

packages/tools/pie-css/package.json
jamieomaguire: we aren't packaging any javascript for consumers @raoufswe

raoufswe: I believe `vite` isn't limited to packaging for consumers only. it can be used to transpile your typescript code to javascript code and I think this is what we are using ts-node for here.

the flow would be, similar to the other packages, have `build-css` as a typescript file, when you build / or run, run vite to transpile your code to javascript and run the transpiled code from the dist folder. you can find more details in the generator setup, but my concern is we shouldn't add two packages to the repo that do the same thing if possible.

jamieomaguire: I think there's a bit of a (very understandable) misunderstanding here. The only dist output is an `index.css` file :) We aren't compiling the TS. It's just running in Node to build the CSS file. Therefore we don't need a vite config

packages/tools/pie-css/test/css/css.spec.ts
jamieomaguire: hope you don't mind but I'd prefer to keep this for now :)

raoufswe (approved):

siggerzz (approved):

PR #675, opened 2023-08-02
feat(pie-css): DSW-1020 add font-size scss mixin

jamieomaguire:
---
"@justeattakeaway/pie-css": minor
---

[Added] - SCSS font-size mixin and scss output testing


raoufswe (approved):

packages/tools/pie-css/test/scss/unit/validity.spec.ts
```diff
@@ -0,0 +1,36 @@
+import {
+    describe,
+    it,
+    expect,
+} from 'vitest';
+
+import cssValidator from 'w3c-css-validator';
+import { compileCss } from '../../../utilities/compileCss';
```
ashleynolan, line 8: Would this path work?

```suggestion
import { compileCss } from './utilities/compileCss';
```

packages/tools/pie-css/test/scss/unit/mixins/fontSize.spec.ts
```diff
@@ -0,0 +1,41 @@
+import {
+    describe,
+    it,
+    expect,
+} from 'vitest';
+
+import { compileCss, stripCSSWhitespace } from '../../../../utilities/compileCss';
```
ashleynolan, line 7: Same here with the path? Could you use `./…`

ashleynolan (commented):
Looks good – small comment on the path but otherwise :shipit:

packages/tools/pie-css/test/scss/unit/mixins/fontSize.spec.ts
sebastienrocks: Just out of curiosity, are we not able to leverage aliases like with webpack? `@/`, `@icons/`

jamieomaguire: I think we do something similar in `tsconfigs` elsewhere in this repo - let me see if I can do it here. We don't use webpack here @sebastienrocks

jamieomaguire: will update in the next pr!

raoufswe (approved):

ashleynolan (approved):

PR #683, opened 2023-08-03
feat(pie-icons-vue): DSW-862 replaced bili with rollup

LTurns:
"@justeattakeaway/pie-icons-vue": minor
[Changed] - JS bundler in package from `Bili` to `Rollup`

"wc-vue3": minor
[Added] - style links to `index.html` so buttons are styled


apps/examples/wc-nuxt2/pages/index.vue
```diff
@@ -1,42 +1,42 @@
-<template>
```
LTurns, line 1: no changes in this file 👍 I used it to test but not sure why it's removed and re-added the code

raoufswe (approved):

fernandofranca:
💯 Everything seems perfect, though I see a main difference between the Vue and React icons packages.

The React one generates barrel files for tying the independent icons files together, so they can be imported or required like the documentation describes:  `import { IconCalendar, IconAlertTriangleLarge, ... } from '@justeattakeaway/pie-icons-***';`.

Should we keep the consistency and provide the barrel files?

LTurns:
> 💯 Everything seems perfect, though I see a main difference between the Vue and React icons packages.
> 
> The React one generates barrel files for tying the independent icons files together, so they can be imported or required like the documentation describes: `import { IconCalendar, IconAlertTriangleLarge, ... } from '@justeattakeaway/pie-icons-***';`.
> 
> Should we keep the consistency and provide the barrel files?

@fernandofranca thank you 🎉 is the barrel file in the react one the `index.tsx` that gets generated on build? I believe the same is happening in pie-icons-vue inside the `generated` folder. I have tested the changes in the nuxt example app by importing them like above and all is working 😄

kevinrodrigues (approved):

packages/tools/pie-icons-vue/rollup.config.js
```diff
@@ -0,0 +1,39 @@
+import del from 'rollup-plugin-delete';
+import { nodeResolve } from '@rollup/plugin-node-resolve';
+
+export default [
```
kevinrodrigues, line 4: You could share this config and override what you need doesn't have to happen in this pr tho.

You could share this and the next one in a shared config and then pass it into each.

```javascript
const sharedConfig = {
    input: ['generated/index.js'],
    output: {
        exports: 'named',
        dir: '',
        format: '',
        name: '@justeattakeaway/pie-icons-vue',
        preserveModulesRoot: 'icons',
        entryFileNames: '[name].[format]',
        preserveModules: true,
    },
    plugins: [],
};
```
```javascript

export default [
    {
        ...sharedConfig,
        output: {
            ...sharedConfig.output,
            dir: 'esm' ...
        }
    },
    {
       ...sharedConfig,
        output: {
            ...sharedConfig.output,
            dir: 'dist' ...
        }
     }
]
```

LTurns: yeah good point actually, would be worth looking into making it shared now rollup is being used by all icon packages 👍 I'll tackle this in a separate PR. Thank you!

PR #696, opened 2023-08-07
feat(pie-toggle-switch): DSW-999 add aria props

FayeCarter:

[Added] - aria props

### Reviewer 1

### Reviewer 2

https://github.com/justeattakeaway/pie/assets/24740015/b909321c-4e73-4280-b8b2-01b55d0ce01e


packages/components/pie-toggle-switch/src/index.ts
```diff
@@ -35,28 +38,34 @@ export class PieToggleSwitch extends RtlMixin(LitElement) implements ToggleSwitc
...
+        const toggleSwitchId = 'toggle-switch-description';
+
         return html`
             <label
                 data-test-id="toggle-switch-component"
                 class="c-toggleSwitch"
                 ?isChecked="${isChecked}"
-                ?isDisabled=${isDisabled}>
+                ?isDisabled=${isDisabled}
+                aria-label="${aria?.label || null}">
```
kevinrodrigues, line 54: I think instead of null you can use `nothing` that won't render the aria-label if it's not provided i believe.

packages/components/pie-toggle-switch/src/index.ts
```diff
@@ -35,28 +38,34 @@ export class PieToggleSwitch extends RtlMixin(LitElement) implements ToggleSwitc
...
                 <input
                     role="switch"
                     type="checkbox"
                     class="c-toggleSwitch-input"
                     .checked="${isChecked}"
                     .disabled="${isDisabled}"
-                    @change="${this.onToggleChange}">
+                    @change="${this.onToggleChange}"
+                    aria-describedby="${aria?.describedBy ? toggleSwitchId : ''}">
```
kevinrodrigues, line 62: Similar to above

kevinrodrigues, line 62: Would be nice to include some tests for these too (component).

packages/components/pie-toggle-switch/src/defs.ts
```diff
@@ -1,6 +1,15 @@
 import { RTLComponentProps } from '@justeattakeaway/pie-webc-core';
 
+export type AriaProps = {
+    label?: string,
+    describedBy?: string
+};
+
 export interface ToggleSwitchProps extends RTLComponentProps {
+    /**
+     * The ARIA labels used for the mtoggle switch.
```
jamieomaguire, line 10: think `mtoggle` might be a typo

FayeCarter: i believe that 'toggle switch' might have been 'modal' before the copy paste 😬

packages/components/pie-toggle-switch/src/index.ts
```diff
@@ -35,28 +38,35 @@ export class PieToggleSwitch extends RtlMixin(LitElement) implements ToggleSwitc
...
                 <input
+                    data-test-id="toggle-switch-input"
                     role="switch"
                     type="checkbox"
                     class="c-toggleSwitch-input"
                     .checked="${isChecked}"
                     .disabled="${isDisabled}"
-                    @change="${this.onToggleChange}">
+                    @change="${this.onToggleChange}"
+                    aria-describedby="${aria?.describedBy ? toggleSwitchId : ''}">
```
jamieomaguire, line 63: above we use `nothing`. here an empty string - should they be consistent? Is this adding an empty attribute if the prop is empty?

apps/pie-storybook/stories/pie-toggle-switch.stories.ts
```diff
@@ -63,6 +72,7 @@ const Template = (props: ToggleSwitchProps): TemplateResult => {
         <pie-toggle-switch
             ?isChecked=${isChecked}
             ?isDisabled=${isDisabled}
+            .aria=${aria}
```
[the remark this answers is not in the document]
FayeCarter: image
nope

packages/components/pie-toggle-switch/src/index.ts
FayeCarter: ah I thought I'd updated that 😕. I'll fix it up

packages/components/pie-toggle-switch/src/toggle-switch.scss
```diff
@@ -1,3 +1,5 @@
+@use '@justeat/pie-design-tokens/dist/jet.scss' as dt;
```
jamieomaguire, line 1: I don't think we need this. The expectation is the consumer will have the tokens already loaded into the consuming application

packages/components/pie-toggle-switch/src/index.ts
```diff
@@ -35,28 +38,35 @@ export class PieToggleSwitch extends RtlMixin(LitElement) implements ToggleSwitc
...
 
                 <div class="c-toggleSwitch-control">
                     ${isChecked ? html`<icon-check></icon-check>` : nothing}
                 </div>
             </label>
+            ${aria?.describedBy ? html`<div id="${toggleSwitchId}" data-test-id="${toggleSwitchId}" class="c-toggleSwitch-description">${aria?.describedBy}</div>` : nothing}
```
jamieomaguire, line 69: How come we're using html rather than an attribute with the prop value?

raoufswe: not sure why we might need this. the input element can accept `aria-describedby` that we can use. 
maybe it's worth checking how it's done in snacks? the component is called switch.

FayeCarter: Hey,

So I spoke to Jamie. `aria-describedby` takes an id for a description element. So what I have here is correct if we still want to allow an `aria.describedBy` prop

@kevinrodrigues I've added a screen recording to the pr description for adding the following as a prop. It reads the value, the label, the description and states that it is a switch.

```
  aria: {
    label: 'label',
    describedBy: 'description'
 }
 ```
 
 @raoufswe in snacks we just pass the id and the consumer, I'm guessing, needs to create their own description element.
 
 I've done similar to how we have in fozzie. Happy to change it to whatever you all prefer

jamieomaguire: @FayeCarter is totally right - I completely mixed up `aria-describedby` with `aria-label` 🤦

sebastienrocks:
Can we please add the prop to the readme as well?

sebastienrocks (approved):

fernandofranca (approved):
Awesome work!

PR #697, opened 2023-08-08
feat(pie-toggle-switch): DSW-1040 add pie-css to next13 example app

jamieomaguire:
Adds `pie-css` to the next-13 example app

### DIFF
--- a/apps/examples/wc-next13/pages/index.tsx
+++ b/apps/examples/wc-next13/pages/index.tsx
@@ -4,6 +4,7 @@ import { ButtonProps, sizes, variants } from '@justeattakeaway/pie-button';
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore
 import { PieButton } from '@justeattakeaway/pie-button/dist/react';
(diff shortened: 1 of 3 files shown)


apps/examples/wc-next13/pages/index.tsx
```diff
@@ -4,6 +4,7 @@ import { ButtonProps, sizes, variants } from '@justeattakeaway/pie-button';
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore
 import { PieButton } from '@justeattakeaway/pie-button/dist/react';
+import '@justeattakeaway/pie-css';
```
raoufswe, line 7: I think this should be added to the root entry of the app (_app.tsx) just in case we wanna add more pages in the future as that will ensure that all pages will have the package available.

jamieomaguire: That's entirely reasonable. We are actually going to overhaul all of these example apps very soon. Would you mind if we left this as-is for now and refactor as a part of that future work?

raoufswe (approved):

PR #699, opened 2023-08-08
feat(pie-monorepo): DSW-1040 integrate pie-css with angular app

jamieomaguire:
Adds `pie-css` to the Angular example app

### DIFF
--- a/apps/examples/wc-angular12/angular.json
+++ b/apps/examples/wc-angular12/angular.json
@@ -100,5 +100,8 @@
             }
         }
         }
+    },
+    "cli": {
+      "analytics": false
     }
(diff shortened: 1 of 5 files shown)


raoufswe (approved):

apps/examples/wc-angular12/angular.json
```diff
@@ -100,5 +100,8 @@
             }
         }
         }
+    },
```
jamieomaguire, line 103: This was caused by running yarn dev and opting out of recording analytics - I wouldn't worry about it :)

LTurns (approved):

PR #705, opened 2023-08-08
feat(pie-toggle-switch): DSW-998 label prop - DNMY

kevinrodrigues:

[Added] - Label Prop

I'm not sure i like the fact the label isn't clickable in a sense that the toggle won't toggle. I checked snacks and it is similar there so might be good to get some thoughts on that (can sort that out in another ticket).

### Reviewer 1

### Reviewer 2


webbertakken:
Personally I almost always use labels to interact with inputs. Because they're often easier to click. Websites that don't allow clickable labels are very frustrating (and feel badly done) to me.

webbertakken (approved):

packages/components/pie-toggle-switch/src/defs.ts
```diff
@@ -1,6 +1,14 @@
 import { RTLComponentProps } from '@justeattakeaway/pie-webc-core';
 
-export interface ToggleSwitchProps extends RTLComponentProps {
+export type LabelProps = {
+    text: string;
+    position: {
+        leading: boolean;
+        trailing: boolean;
```
raoufswe, line 7: what do you think about having it as position=[leading, trailing]?
i believe this will bring more clarity than having to flip the boolean logic for each

kevinrodrigues: I was thinking this too, ill take a look

kevinrodrigues:
> Personally I almost always use labels to interact with inputs. Because they're often easier to click. Websites that don't allow clickable labels are very frustrating (and feel badly done) to me.

Yup i agree I use it in the same way. I mentioned it in the description too so will create a ticket around this. 👍

packages/components/pie-toggle-switch/src/defs.ts
kevinrodrigues: Thanks, this actually helped clean up a bit of code too 👍

packages/components/pie-toggle-switch/src/defs.ts
```diff
@@ -1,6 +1,11 @@
 import { RTLComponentProps } from '@justeattakeaway/pie-webc-core';
 
-export interface ToggleSwitchProps extends RTLComponentProps {
+export type LabelProps = {
+    text: string;
+    position: string;
```
raoufswe, line 5: is okay if we can define the position type as we do with size in the button defs for consistency?

this way we can export the label positions as a const, and can be used in the story.

defs.ts
```
export const labelPositions = ['leading', 'trailing'] as const;

export type LabelProps = {
    text: string;
    position: labelPositions types[number];
}
```

and in the story, we can do 

```
    label: {
            description: 'The label configuration for the toggle switch,
            control: {
                type: 'object',
                defaultValue: {
                    text: 'Label',
                    options: labelPoisitons,
                },
            },
        },

```

raoufswe: we probably we wanna export labelPositiosns from index.ts too

packages/components/pie-toggle-switch/src/index.ts
```diff
@@ -25,38 +29,67 @@ export class PieToggleSwitch extends RtlMixin(LitElement) implements ToggleSwitc
...
+     * Renders the label for a toggle switch if provided.
+     * You can only pass in either a leading or trailing label, if you pass both in
+     * as true nothing renders.
+     *
+     * @private
+     */
+    private renderToggleSwitchLabel (): TemplateResult {
+        const { label } = this;
+
+        if (label?.text && (label?.position === 'leading' || label?.position === 'trailing')) {
```
raoufswe, line 53: if we go with my suggestion above, this could become

```suggestion
        if (label?.text && (labelPoisitons.include(label.position))) {
```

packages/components/pie-toggle-switch/src/index.ts
```diff
@@ -25,38 +29,67 @@ export class PieToggleSwitch extends RtlMixin(LitElement) implements ToggleSwitc
...
+            return html`
+                <span
+                    data-test-id="toggle-switch-label"
+                    class="c-toggleSwitch-label ${label.position === 'leading' ? 'c-toggleSwitch--leading' : 'c-toggleSwitch--trailing'}">
```
raoufswe, line 57:
```suggestion
                    class="c-toggleSwitch-label c-toggleSwitch--${label.position} ">
```

packages/components/pie-toggle-switch/test/visual/pie-toggle-switch.spec.ts
```diff
@@ -19,3 +19,25 @@ import { PieToggleSwitch } from '@/index';
         await percySnapshot(page, `ToggleSwitch - isChecked = ${isChecked} and isDisabled = ${isDisabled}`);
     });
 });
+
+test.describe('Prop: `Label`', () => {
+    test.describe('when passed in', () => {
+        const labelPositions = ['leading', 'trailing'];
```
raoufswe, line 25: this should be imported from defs.ts

packages/components/pie-toggle-switch/src/index.ts
kevinrodrigues: fixed.

raoufswe (approved):

packages/components/pie-toggle-switch/README.md
```diff
@@ -59,8 +59,9 @@ import { PieToggleSwitch } from '@justeattakeaway/pie-toggle-switch/dist/react';
 
 | Property | Type | Default | Description |
 |---|---|---|---|
-| isChecked | `Boolean` | false | Indicates whether the switch is on or off |
+| isChecked  | `Boolean` | false | Indicates whether the switch is on or off |
 | isDisabled | `Boolean` | false | Indicates whether the switch is disabled or not |
+| label      | `Object`  | `undefined` | Provides a label for the component. You can set the label position using options: `leading` | `trailing` e.g object: `label: { text: 'Label', position: 'leading' }` |
```
sebastienrocks, line 64: Seems like the default value is incorrect?

https://github.com/justeattakeaway/pie/pull/705/files#diff-4c058adb36e0f8491d5e07aec4d39d31af1d2bac947533eb1b865b7ea54c3869R51

packages/components/pie-toggle-switch/src/defs.ts
```diff
@@ -1,6 +1,15 @@
 import { RTLComponentProps } from '@justeattakeaway/pie-webc-core';
 
-export interface ToggleSwitchProps extends RTLComponentProps {
+export const labelPositions = ['leading', 'trailing'] as const;
+
+export type LabelPosition = typeof labelPositions[number];
+
+export type LabelProps = {
```
jamieomaguire, line 7: I actually don't know if I'd prefer a label position prop separate from the label text. I'm really not a fan of the complexity added by these object prop definitions but I appreciate its a matter of preference

sebastienrocks (approved):

kevinrodrigues:
> It also seems like null values (what storybook adds if you remove the object) break the component:
> 
> https://pr705.d17ja0ul7nrdy0.amplifyapp.com/?path=/story/toggle-switch--default&args=label:!null
> 
> Please can we avoid using objects for properties by default, or can we add some more guards against invalid values?
> 
> Screenshot 2023-08-09 at 16 42 47

Will take a look at this once i'm back Monday if that's ok. Ill mark it as DNMY for now.

raoufswe:
the PR should be rebased and its comments are resolved here https://github.com/justeattakeaway/pie/pull/719

PR #718, opened 2023-08-11
ci(pie-monorepo): DSW-1077 upgrade to latest turbo

xander-marjoram:
- [Added] - `turbo.json` for some individual packages (storybook, docs, icons-react, icons-vue and icons-webc)
- [Changed] - refactor root turbo config
- [Changed] - refactor package scripts to better leverage turbo caching

### Reviewer 1

### Reviewer 2


apps/pie-docs/package.json
```diff
@@ -11,9 +11,8 @@
   "scripts": {
     "build": "DEPLOYMENT_ENVIRONMENT='production' eleventy",
     "build:dev": "DEPLOYMENT_ENVIRONMENT='development' eleventy",
-    "clean:dist": "run -T rimraf dist",
-    "cz": "run -T cz",
```
xander-marjoram, line 15: Removed as this should be run from the package root.

apps/pie-docs/package.json
```diff
@@ -11,9 +11,8 @@
...
-    "dev": "yarn clean:dist && npx @11ty/eleventy --serve",
+    "clean": "run -T rimraf dist",
```
xander-marjoram, line 14: Aligned with naming convention from other packages, which also allows this to be run from the root

apps/pie-storybook/package.json
```diff
@@ -6,8 +6,7 @@
     "dev": "storybook dev -p 6006",
     "build": "storybook build --output-dir dist",
     "lint:scripts": "run -T eslint .",
-    "lint:scripts:fix": "run -T eslint . --fix",
-    "serve": "serve --no-request-logging dist"
```
xander-marjoram, line 10: This was not being used

packages/components/pie-webc-core/package.json
```diff
@@ -9,7 +9,8 @@
   "scripts": {
     "lint:scripts": "run -T eslint .",
     "lint:scripts:fix": "run -T eslint . --fix",
-    "test": "run -T vitest run"
+    "test": "run -T vitest run",
+    "test:watch": "run -T vitest"
```
xander-marjoram, line 13: I've added `:watch` commands wherever we use vitest for consistency

packages/tools/pie-icons-webc/package.json
```diff
@@ -30,13 +30,11 @@
     "PIE"
   ],
   "scripts": {
-    "build": "yarn clean && yarn build:icons && yarn build:dist",
```
xander-marjoram, line 33: We're now making use of dependent tasks via turborepo so that we can benefit from caching.

packages/tools/pie-icons-webc/package.json
```diff
@@ -30,13 +30,11 @@
...
-    "build:dist": "rollup -c",
+    "build": "rollup -c",
     "build:icons": "node ./build.js",
     "clean": "run -T rimraf -rf ./dist/ ./icons/",
     "lint:scripts": "run -T eslint .",
     "lint:scripts:fix": "yarn lint:scripts --fix",
-    "prebuild:example": "yarn build:icons",
```
xander-marjoram, line 39: This was not being used

apps/pie-storybook/turbo.json
```diff
@@ -0,0 +1,14 @@
+{
+  "$schema": "https://turborepo.org/schema.json",
+  "extends": [
+    "//"
+  ],
+  "pipeline": {
+    "dev": {
```
fernandofranca, line 7: `pie-storybook` also has a `build` script. Should it be included here?

apps/pie-docs/turbo.json
```diff
@@ -0,0 +1,43 @@
+{
+  "$schema": "https://turborepo.org/schema.json",
+  "extends": [
+    "//"
```
fernandofranca, line 4: Does the `//` refers to the root `turbo.json`?

xander-marjoram: Yes indeed! https://turbo.build/repo/docs/core-concepts/monorepos/configuring-workspaces#how-it-works

See the note below the code snippet

apps/pie-storybook/turbo.json
xander-marjoram: Good question. I just tried this and it runs just fine without it (because its definition would be identical to the one in the root `turbo.json`). I think this means that I should be able to remove the one that I've defined in `pie-docs/turbo.json` though!

packages/tools/pie-icons-webc/turbo.json
```diff
@@ -0,0 +1,31 @@
...
+      "cache": false,
+      "dependsOn": [
+        "clean"
+      ],
+      "outputs": [
+        "icons"
+      ]
+    },
+    "build": {
+      "cache": false,
```
fernandofranca: Why not cache `pie-icons-webc` build scripts?

fernandofranca (dismissed):
Good stuff!

packages/tools/pie-icons-webc/turbo.json
xander-marjoram: This was something I was looking at. Hopefully resolved now.

packages/tools/pie-icons-react/rollup.config.mjs
```diff
@@ -17,7 +16,6 @@ export default [
             preserveModules: true,
         },
         plugins: [
-            del({ targets: ['esm/*'] }),
```
xander-marjoram, line 20: These are already being deleted by the `clean` script so we don't need to do it again

.changeset/calm-suits-return.md
```diff
@@ -0,0 +1,13 @@
...
+"pie-monorepo": minor
+"pie-storybook": minor
+"@justeattakeaway/pie-wrapper-react": patch
+---
+
+[Added] - `turbo.json` for some individual packages (storybook, docs, icons-react, icons-vue and icons-webc)
+[Removed] - generated vue icons from source control
+[Changed] - use latest turbo version
+[Changed] - refactor root turbo config
+[Changed] - update custom-elements.json
```
ashleynolan, line 12: I think this has been removed for version control now, so can likely get rid of this line?

apps/pie-docs/package.json
```diff
@@ -11,9 +11,8 @@
...
+    "dev": "npx @11ty/eleventy --serve",
```
ashleynolan, line 15: Out of interest - why remove the clean before serving here?

Looks like you may need to add clean as a dependency in the turbo confg for `dev` (it's added to `build:dev` atm, but not the straight up `dev` script?)

apps/pie-storybook/turbo.json
```diff
@@ -0,0 +1,14 @@
...
+      "cache": false,
+      "dependsOn": [
+        "^build"
```
ashleynolan, line 10: I thought with storybook, dev built in non prod mode, and build built in prod mode ready for deploy? So not sure they're a dependency on one another?

packages/tools/pie-icons-react/turbo.json
```diff
@@ -0,0 +1,35 @@
...
+        "clean"
+      ],
+      "outputs": [
+        "icons/**"
+      ]
+    },
+    "build": {
+      "cache": true,
+      "dependsOn": [
+        "^build",
```
ashleynolan, line 23: More as a noob turbo q – why do you need `^build` as a depends on of `build` here?

ashleynolan (commented):
Couple of comments/q's from me – looks good though

apps/pie-docs/package.json
xander-marjoram: Good question! `dev` depends on `build:dev`, which itself depends on `clean`.

apps/pie-storybook/turbo.json
xander-marjoram: The syntax here can be a bit confusing. What this is saying is that `dev` depends on the same things as `build`, i.e., run what is required for `build` (but not `build` itself) before running `dev`.

packages/tools/pie-icons-react/turbo.json
xander-marjoram: This is a good question, I would have thought that this would be implied. I'll try removing it.

xander-marjoram: I don't think turbo implies this relationship actually, so it seems to be necessary.

jamieomaguire (approved):

apps/pie-storybook/turbo.json
ashleynolan: Makes sense 👍

PR #719, opened 2023-08-13
feat(pie-toggle-switch): DSW-998 add label

raoufswe:

[Added] - Label and labelplacement to the toggle switch 

Some of the work of this PR is carried out from here https://github.com/justeattakeaway/pie/pull/705

### Reviewer 1

### Reviewer 2

### DIFF
--- a/packages/components/pie-toggle-switch/src/defs.ts
+++ b/packages/components/pie-toggle-switch/src/defs.ts
@@ -1,11 +1,24 @@
 import { RTLComponentProps } from '@justeattakeaway/pie-webc-core';
 
+export const labelPlacements = ['leading', 'trailing'] as const;
+
 export type AriaProps = {
     label?: string,
     describedBy?: string
 };
 
-export interface ToggleSwitchProps extends RTLComponentProps {
+export type LabelProps = {
+    label: string;
+    labelPlacement: typeof labelPlacements[number];
+  } | {
+    label: string;
+    labelPlacement?: never;
+  } | {
+    label?: never;
+    labelPlacement?: never;
+  };
+
+export type ToggleSwitchProps = RTLComponentProps & LabelProps & {
     /**
      * The ARIA labels used for the toggle switch.
      */
@@ -13,11 +26,11 @@ export interface ToggleSwitchProps extends RTLComponentProps {
     /**
      * Same as the HTML checked attribute - indicates whether the switch is on or off
      */
-    isChecked: boolean;
+    isChecked?: boolean;
     /**
      * Same as the HTML checked attribute - indicates whether the switch disabled or not
      */
-    isDisabled: boolean;
+    isDisabled?: boolean;
 }
 
 /**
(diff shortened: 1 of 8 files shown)


packages/components/pie-toggle-switch/src/index.ts
```diff
@@ -33,11 +43,34 @@ export class PieToggleSwitch extends RtlMixin(LitElement) implements ToggleSwitc
     onToggleChange (event: Event) {
         const target = event?.target as HTMLInputElement;
         this.isChecked = target.checked;
-        this.dispatchEvent(new CustomEvent(EVENT_TOGGLE_SWITCH_CHANGED, { detail: this.isChecked }));
+        const changedEvent = new CustomEvent(ON_TOGGLE_SWITCH_CHANGED_EVENT, { detail: this.isChecked });
+        this.dispatchEvent(changedEvent);
```
raoufswe, line 47: for some reason, defining the custom event and dispatching it in the same line creates two events in the react wrapper.

Screenshot 2023-08-13 at 23 29 19

packages/components/pie-toggle-switch/src/defs.ts
```diff
@@ -13,16 +15,24 @@ export interface ToggleSwitchProps extends RTLComponentProps {
...
+    labelPlacement?: typeof labelPlacements[number];
 }
 
 /**
  * Event name for when the toggle switch checked state is changed.
  *
  * @constant
  */
-export const EVENT_TOGGLE_SWITCH_CHANGED = 'pie-toggle-switch-changed';
+export const ON_TOGGLE_SWITCH_CHANGED_EVENT = 'pie-toggle-switch-changed';
```
fernandofranca, line 38: More of a general TS question, shouldn't everything be typed?

fernandofranca: I mean, not only for the sake of declaring types, but also because TS might complain about where types might be lacking.

raoufswe: typed as specifying that this variable should be a string? 

```
export const ON_TOGGLE_SWITCH_CHANGED_EVENT: string = 'pie-toggle-switch-changed';
```

if so, I'm not a big fan of redeclaring that the type should be string here as the compiler will infer that automatically as far as I know and we aren't reassigning it anywhere so I believe we don't have to.

packages/components/pie-toggle-switch/src/index.ts
fernandofranca: It might sound like a dumb question... Are those events the same and why do we have both?

packages/components/pie-toggle-switch/src/defs.ts
raoufswe: oh no, it should be automatically inferred by the compiler that's a string. 

if you try to reassign the variable to something else, it will complain.

packages/components/pie-toggle-switch/src/index.ts
raoufswe: yep, and the first one is wrong and shouldn't be added. 
when you inline the custom event and dispatch it as it was done previously, the first event is created, which shouldn't happen.

fernandofranca (approved):

packages/components/pie-toggle-switch/src/index.ts
fernandofranca: Got it. I thought that was in runtime! 😨

PR #722, opened 2023-08-15
fix(pie-wrapper-react): DSW-1053 set up types and exports for react components

jamieomaguire:
## Changes
- Ensures that TS types are available for TS-based react consumers of our components. Before the `*.d.ts` files were unavailable when React apps used our components and Typescript.

- Export all of the `defs.ts` files from our components so that the generated `react.ts` file and `index.ts` file export the same things (except for the component being wrapped for React).

### The dist folder:
dist

### Example of `react.ts`
reactts


raoufswe (approved):
image

fernandofranca (approved):
Nice stuff!!!

PR #744, opened 2023-08-21
feat(pie-toggle-switch): DSW-1081 adds support for toggling switch via label

kevinrodrigues:

[Added] label toggling support. 

- [X] If it is a core feature, I have added thorough tests

### Reviewer 1

### Reviewer 2

### DIFF
--- a/.changeset/yellow-radios-glow.md
+++ b/.changeset/yellow-radios-glow.md
@@ -2,4 +2,4 @@
 "@justeattakeaway/pie-toggle-switch": minor
 ---
 
-[Added] label toggling support
+[Added] - label toggling support
--- a/.changeset/yellow-radios-glow.md
+++ b/.changeset/yellow-radios-glow.md
@@ -1,0 +1,5 @@
+---
+"@justeattakeaway/pie-toggle-switch": minor
+---
+
+[Added] label toggling support
--- a/packages/components/pie-toggle-switch/src/index.ts
+++ b/packages/components/pie-toggle-switch/src/index.ts
@@ -55,11 +55,12 @@ export class PieToggleSwitch extends RtlMixin(LitElement) implements ToggleSwitc
 
         if (label && labelPlacement && (labelPlacements.includes(labelPlacement))) {
             return html`
-                <span
+                <label
+                    for="toggle-switch"
                     data-test-id="toggle-switch-label"
                     class="c-toggleSwitch-label c-toggleSwitch--${labelPlacement}">
                     ${label}
-                </span>`;
+                </label>`;
         }
 
         return html``;
@@ -84,6 +85,7 @@ export class PieToggleSwitch extends RtlMixin(LitElement) implements ToggleSwitc
                     ?isChecked="${isChecked}"
                     ?isDisabled=${isDisabled}>
                     <input
+                        id="toggle-switch"
                         data-test-id="toggle-switch-input"
                         role="switch"
                         type="checkbox"
--- a/packages/components/pie-toggle-switch/test/component/pie-toggle-switch.spec.ts
+++ b/packages/components/pie-toggle-switch/test/component/pie-toggle-switch.spec.ts
@@ -45,7 +45,7 @@ test.describe('Component: `Pie toggle switch`', () => {
         expect(pieToggleSwitchComponent).toBe(false);
     });
 
-    test('should set `id` to `toggle-switch` on the input element', async ({ mount }) => {
+    test('should have an `id` of `toggle-switch` on the input element', async ({ mount }) => {
         // Arrange
         const component = await mount(PieToggleSwitch);
 
--- a/packages/components/pie-toggle-switch/test/component/pie-toggle-switch.spec.ts
+++ b/packages/components/pie-toggle-switch/test/component/pie-toggle-switch.spec.ts
@@ -67,7 +67,7 @@ test.describe('Component: `Pie toggle switch`', () => {
             });
 
             // Act
-            const pieToggleSwitchLabel = await component.locator(toggleSwitchLabel);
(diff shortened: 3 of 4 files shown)


packages/components/pie-toggle-switch/src/index.ts
```diff
@@ -55,11 +55,12 @@ export class PieToggleSwitch extends RtlMixin(LitElement) implements ToggleSwitc
 
         if (label && labelPlacement && (labelPlacements.includes(labelPlacement))) {
             return html`
-                <span
+                <label
```
xander-marjoram, line 58: It looks to me like the structure of the label(s) will be like this:
```html
<label for="toggle-switch">Leading label</label>
<label>
    <input id="toggle-switch" />
</label>
<label for="toggle-switch">Trailing label</label>
```

Is that right? I know the leading and trailing labels won't be showing at the same time but I'm not sure we need the label wrapping the input too.

raoufswe:
looks good! 
one question, should we make the cursor `pointer` to indicate it's clickable?

`cursor: pointer;`  if not disbaled ~

xander-marjoram:
@raoufswe Do you mean over the label? I like the idea but I think native labels don't use `cursor: pointer` so we should probably match that

kevinrodrigues:
> @raoufswe Do you mean over the label? I like the idea but I think native labels _don't_ use `cursor: pointer` so we should probably match that

I wasn't sure either so left as is.

packages/components/pie-toggle-switch/test/component/pie-toggle-switch.spec.ts
```diff
@@ -30,7 +31,7 @@ test.describe('Component: `Pie toggle switch`', () => {
         const pieToggleSwitchComponent = await component.locator(componentSelector).isChecked();
 
         // Assert
-        await expect(pieToggleSwitchComponent).toBe(false);
+        expect(pieToggleSwitchComponent).toBe(false);
```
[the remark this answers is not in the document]
kevinrodrigues: Yes so awaiting doesn't do anything here as it has already resolved. I will remove the others 👍  p.s no errors just something i noticed.

packages/components/pie-toggle-switch/src/index.ts
kevinrodrigues: Will take a look to see if removing the wrapping works ok tomorrow 👍

.changeset/yellow-radios-glow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@justeattakeaway/pie-toggle-switch": minor
+---
+
+[Added] label toggling support
```
ashleynolan, line 5:
```suggestion
[Added] - label toggling support
```

packages/components/pie-toggle-switch/src/index.ts
ashleynolan, line 58: Part of me wonders if we're building a lot of label functionality into the toggle when it's due to be pulled out as a separate component before we tackle any of the other inputs.

Would it make sense to make the tickets for the label now (after you've merged this PR) rather than wait, as it may mean a bit of consideration in terms of how you structure this component pulling it in?

kevinrodrigues: Happy to create some tickets for this. Will speak to you and Raouf after this one is merged 👍

packages/components/pie-toggle-switch/test/component/pie-toggle-switch.spec.ts
```diff
@@ -41,7 +42,36 @@ test.describe('Component: `Pie toggle switch`', () => {
         const pieToggleSwitchComponent = await component.locator(componentSelector).isDisabled();
 
         // Assert
-        await expect(pieToggleSwitchComponent).toBe(false);
+        expect(pieToggleSwitchComponent).toBe(false);
+    });
+
+    test('should set `id` to `toggle-switch` on the input element', async ({ mount }) => {
```
xander-marjoram, line 48: IMO not sure this test is required because it's not really dynamic. The test below is valid though because it's conditional

packages/components/pie-toggle-switch/src/index.ts
kevinrodrigues: I think i will leave this as is for now. Snacks follow a similar structure although they use a span for for the actual label where text appears.

packages/components/pie-toggle-switch/test/component/pie-toggle-switch.spec.ts
kevinrodrigues: I sort of agree here, I wasn't going to add this but i felt that as it's tied to the `for` part of the label it might be worth keeping it so it's not messed with or changed in future & not updated in the input element, probably unlikely to happen but added as a belt and braces thing. Happy to remove it tho.

kevinrodrigues: Maybe the description is misleading. "should have an attribute ID toggle-switch" or something like this instead?

packages/components/pie-toggle-switch/src/index.ts
```diff
@@ -84,6 +85,7 @@ export class PieToggleSwitch extends RtlMixin(LitElement) implements ToggleSwitc
                     ?isChecked="${isChecked}"
                     ?isDisabled=${isDisabled}>
                     <input
+                        id="toggle-switch"
```
jamieomaguire, line 88: nothing weird happens if we have multiple toggle switches on a single page all with the same ID, does it?

raoufswe (approved):

jamieomaguire (approved):

PR #748, opened 2023-08-22
feat(pie-link): DSW-1121 allow the link component to behave like a button

raoufswe:

[Added] - allow the link component to behave like a button

[Changed] - fixed visual tests for the link and divider components

### Reviewer 1

### Reviewer 2

### DIFF
--- a/apps/pie-storybook/stories/pie-link.stories.ts
+++ b/apps/pie-storybook/stories/pie-link.stories.ts
@@ -21,7 +21,7 @@ const defaultArgs: LinkProps = {
     target: '_blank',
     isBold: false,
     isStandalone: false,
-    slot: 'This is Lit!',
+    slot: 'Link',
 };
 
 const linkStoryMeta: LinkStoryMeta = {
--- a/apps/pie-storybook/stories/pie-link.stories.ts
+++ b/apps/pie-storybook/stories/pie-link.stories.ts
@@ -1,7 +1,8 @@
 import { html, TemplateResult, nothing } from 'lit';
 import { type StoryObj as Story } from '@storybook/web-components';
 import {
-    PieLink, LinkProps as LinkBaseProps, sizes, variants, iconPlacements,
+    PieLink, LinkProps as LinkBaseProps, sizes, variants,
+    iconPlacements, tags,
 } from '@justeattakeaway/pie-link';
 import type { StoryMeta, SlottedComponentProps } from '../types';
 
@@ -13,6 +14,7 @@ type LinkProps = SlottedComponentProps<LinkBaseProps>;
 type LinkStoryMeta = StoryMeta<LinkProps>;
 
 const defaultArgs: LinkProps = {
+    tag: 'a',
     variant: 'default',
     size: 'medium',
     href: 'https://pie.design',
@@ -26,6 +28,14 @@ const linkStoryMeta: LinkStoryMeta = {
     title: 'Link',
     component: 'pie-link',
     argTypes: {
+        tag: {
+            description: 'Set the element tag of the link.',
+            control: 'select',
+            options: tags,
+            defaultValue: {
+                summary: 'a',
+            },
+        },
         variant: {
             description: 'Set the variant of the link.',
             control: 'select',
@@ -50,14 +60,17 @@ const linkStoryMeta: LinkStoryMeta = {
         href: {
             description: 'The URL that the hyperlink should point to',
             control: 'text',
+            if: { arg: 'tag', eq: 'a' },
         },
         target: {
             description: 'Set where to display the linked URL',
             control: 'text',
+            if: { arg: 'tag', eq: 'a' },
         },
         rel: {
             description: 'Set what the relationship of the linked URL is',
             control: 'text',
+            if: { arg: 'tag', eq: 'a' },
         },
         isBold: {
             description: 'If `true`, makes the link text bold',
@@ -93,6 +106,7 @@ const linkStoryMeta: LinkStoryMeta = {
 export default linkStoryMeta;
 
 const Template = ({
+    tag,
     href,
     target,
     rel,
@@ -104,6 +118,7 @@ const Template = ({
     iconPlacement,
 }: LinkProps): TemplateResult => html`
         <pie-link
+            tag="${tag}"
             variant="${variant}"
             size="${size}"
             iconPlacement="${iconPlacement || nothing}"
--- a/apps/pie-storybook/stories/pie-link.stories.ts
+++ b/apps/pie-storybook/stories/pie-link.stories.ts
@@ -1,8 +1,8 @@
 import { html, TemplateResult, nothing } from 'lit';
 import { type StoryObj as Story } from '@storybook/web-components';
 import {
-    PieLink, LinkProps as LinkBaseProps, sizes, variants,
-    iconPlacements, tags,
+    PieLink, LinkProps as LinkBaseProps, sizes,
+    variants, iconPlacements, tags, buttonTypes,
 } from '@justeattakeaway/pie-link';
 import type { StoryMeta, SlottedComponentProps } from '../types';
 
@@ -18,7 +18,7 @@ const defaultArgs: LinkProps = {
     variant: 'default',
     size: 'medium',
     href: 'https://pie.design',
-    target: '_target',
+    target: '_blank',
     isBold: false,
     isStandalone: false,
     slot: 'This is Lit!',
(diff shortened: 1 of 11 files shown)


apps/pie-storybook/stories/pie-link.stories.ts
```diff
@@ -50,14 +60,17 @@ const linkStoryMeta: LinkStoryMeta = {
         href: {
             description: 'The URL that the hyperlink should point to',
             control: 'text',
+            if: { arg: 'tag', eq: 'a' },
```
ashleynolan, line 63: This is super cool – didn't know you could have if statements in storybook like this.

packages/components/pie-link/src/index.ts
```diff
@@ -49,27 +51,38 @@ export class PieLink extends LitElement implements LinkProps {
...
-                ?isStandalone=${isStandalone}>
+                ?isStandalone=${isStandalone}
+                href=${!isButton && href ? href : nothing}
+                target=${!isButton && target ? target : nothing}
+                rel=${!rel && rel ? target : nothing}>
                     <div class="c-link-content">
-                      ${iconPlacement === 'leading' ? html`<slot name="icon"></slot>` : nothing}
-                      <slot></slot>
-                      ${iconPlacement === 'trailing' ? html`<slot name="icon"></slot>` : nothing}
+                        ${iconPlacement === 'leading' ? html`<slot name="icon"></slot>` : nothing}
```
ashleynolan, line 81: I don't think it's valid to have a `div` inside a `button` – may need to update it to be a `span` now?

apps/pie-storybook/stories/pie-link.stories.ts
raoufswe: yes it's neat 💯

packages/components/pie-link/src/index.ts
raoufswe: yep makes sense 👍

packages/components/pie-link/README.md
```diff
@@ -57,16 +57,18 @@ import { PieLink } from '@justeattakeaway/pie-link/dist/react';
...
-| rel           | `String`  | `undefined` | Native html `rel` attribute                                                                  |
-| target        | `String`  | `undefined` | Native html `target` attribute                                                               |
-| isBold        | `Boolean` | `false`     | If `true`, sets the link text bold                                                           |
-| isStandalone  | `Boolean` | `false`     | If `true`, sets the link as a block element                                                  |
-| iconPlacement | `String`  | `leading`   | Icon placements of the icon slot, if provided, one of `iconPlacements` - `leading`, `trailing`         |
+| Property      | Type        | Default       | Description                                                                                          |
+| ------------- | ----------- | ------------- | ---------------------------------------------------------------------------------------------------- |
+| tag           | `String`  | `a`         | The renderd HTML element of the link, one of `tags` – `a`, `button`                         |
```
kevinrodrigues, line 62: Rendered.

packages/components/pie-link/src/defs.ts
```diff
@@ -1,8 +1,14 @@
...
 export const sizes = ['small', 'medium'] as const;
 export const iconPlacements = ['leading', 'trailing'] as const;
+export const tags = ['a', 'button'] as const;
+export const buttonTypes = ['submit', 'button', 'reset', 'menu'] as const;
 
 export interface LinkProps {
+     /**
+     * What HTML element the link should be such as a or button.
+     */
+    tag: typeof tags[number];
```
jamieomaguire, line 11: could we perhaps use an `isButton` prop instead?

packages/components/pie-link/src/index.ts
```diff
@@ -20,6 +18,10 @@ const componentSelector = 'pie-link';
  */
 
 export class PieLink extends LitElement implements LinkProps {
+    @property()
+    @validPropertyValues(componentSelector, tags, 'a')
```
jamieomaguire, line 22: I think using `isButton` could be simpler and remove the need for this validation

packages/components/pie-link/src/link.scss
```diff
@@ -18,6 +18,18 @@
     text-decoration: var(--link-text-decoration);
     cursor: pointer;
 
+    &[tag='a'] {
+        /* Same as default styles */
+    }
+
+    &[tag='button'] {
+        outline: none;
```
jamieomaguire, line 26: I can't seem to find any focus styles in storybook. Could this be removing them?

packages/components/pie-link/src/defs.ts
raoufswe: I don't mind changing to that but the team agreed to follow the generic approach of specifying tags during refinement and GitHub discussion.

jamieomaguire: Fully appreciate that, it just feels more complicated and I'm not sure I understand what benefits it gives us over a boolean

packages/components/pie-link/src/link.scss
raoufswe: focus, active, and hover states are not done yet as designers are still looking into them 💯 
DSW-1057

packages/components/pie-link/src/index.ts
```diff
@@ -47,29 +49,46 @@ export class PieLink extends LitElement implements LinkProps {
...
-                ?isStandalone=${isStandalone}>
-                    <div class="c-link-content">
-                      ${iconPlacement === 'leading' ? html`<slot name="icon"></slot>` : nothing}
-                      <slot></slot>
-                      ${iconPlacement === 'trailing' ? html`<slot name="icon"></slot>` : nothing}
-                    </div>
-                </div>
-            </a>`;
+                ?isStandalone=${isStandalone}
+                href=${!isButton && href ? href : nothing}
```
kevinrodrigues, line 82: Just a thought. Is it worth separating the button and anchor attributes out and then passing them in on a condition instead, it might make better for reading? Something like below (not tested it tho):

```javascript
    const anchorValues = !isButton ? { href, target, rel } : {};
    const buttonValues = isButton ? { type } : {};

    return html`
        <${element} ...${commonAttrs} ...${anchorAttrs} ...${buttonAttrs}>
            `;
......
.....
}
```

raoufswe: I have thought about the same thing but unfortunately lit doesn't support spreading attributes yet

https://github.com/lit/lit/issues/923

packages/components/pie-link/src/defs.ts
raoufswe: in case it clears a bit, the reasoning behind using `tag` instead of a boolean is to provide enough flexibility in cases we wanna render something else in the future other than a button or an anchor such as passing a framework-specific navigation element as a base for our component.

for example, Nextjs provides its own link component for in-app routing and you can't just use a native anchor tag to do that. So if our component is to be used there, we need to support something similar to what chakra-ui does [here](https://chakra-ui.com/docs/components/link/usage#usage-with-nextjs) 

I understand it might be less complex if we refactor to boolean, but passing `tag`, `as`, `renderAs` or `component` is quite common for components libraries.

packages/components/pie-link/src/index.ts
kevinrodrigues: Dam. OK didn't know that.

packages/components/pie-link/src/defs.ts
jamieomaguire: @raoufswe that makes sense, I kind of see it as a case of [YAGNI](https://martinfowler.com/bliki/Yagni.html) to make it this flexible prematurely.

I don't feel super strongly about this but I wanted to highlight I think it could be simpler :)

maledr5: `isButton` sounds for sure easier to implement and kind of easier for consumers to understand (in my opinion). But I also could se we needing to use different tags in other components, such as the `header` where per a11y there might be the need to render an `h1` style on an `h2` tag. We may want to keep the `tag` prop so that its use is consistent among components?

packages/components/pie-link/README.md
raoufswe: fixed

packages/components/pie-link/src/defs.ts
jamieomaguire: Fair enough! PR looks good 👍 🎉

jamieomaguire (approved):

FayeCarter (approved):

apps/pie-storybook/stories/pie-link.stories.ts
```diff
@@ -13,19 +14,28 @@ type LinkProps = SlottedComponentProps<LinkBaseProps>;
 type LinkStoryMeta = StoryMeta<LinkProps>;
 
 const defaultArgs: LinkProps = {
+    tag: 'a',
     variant: 'default',
     size: 'medium',
     href: 'https://pie.design',
-    target: '_target',
+    target: '_blank',
```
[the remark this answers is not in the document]
raoufswe: target is an optional attribute, and this is only for storybook as I thought it's cooler to open the link in a new tab, otherwise, the story iframe will be replaced by the docs site 🤣

ashleynolan (approved):

PR #780, opened 2023-09-01
feat(pie-card-container): DSW-1156 add basic markup, styling and link functionality)

jamieomaguire:
---
"@justeattakeaway/pie-card-container": minor
"pie-storybook": minor
---

[Added] - Base markup, styles and interactive states for the card in it's Link implementation

### DIFF
--- a/packages/components/pie-card-container/README.md
+++ b/packages/components/pie-card-container/README.md
@@ -59,16 +59,22 @@ import { PieCardContainer } from '@justeattakeaway/pie-card-container/dist/react
 
 | Property | Type | Default | Description |
 |---|---|---|---|
-| - | - | - | - |
+| variant | `string` | `default` | What style variant the card should be such as default or inverse. |
+| disabled | `boolean` | `false` | When true, the card container is disabled. |
+| href | `string` | `undefined` | The URL that the card should point to (this will not take effect unless the card is a link). |
+| target | `string` | `undefined` | Where to display the linked URL such as _self, _blank, _parent or _top (this will not take effect unless the card is a link). |
+| rel | `string` | `undefined` | What the relationship of the linked URL is (this will not take effect unless the card is a link). |
+| aria | `object` | `undefined` | The ARIA labels used for various parts of the card. |
+
 
 In your markup or JSX, you can then use these to set the properties for the `pie-card-container` component:
 
 ```html
 <!-- Native HTML -->
-<pie-card-container></pie-card-container>
+<pie-card-container disabled href="/foo/bar" rel="noopener" target="_blank"></pie-card-container>
 
 <!-- JSX -->
-<PieCardContainer></PieCardContainer>
+<PieCardContainer disabled href="/foo/bar" rel="noopener" target="_blank"></PieCardContainer>
 ```
 
 ## Testing
--- a/packages/components/pie-card-container/src/card-container.scss
+++ b/packages/components/pie-card-container/src/card-container.scss
@@ -1,1 +1,95 @@
 @use '@justeattakeaway/pie-css/scss' as p;
+
+@mixin card-interactive-states($bg-color) {
+    &:hover:not([disabled]) {
+        --hover-modifier: calc(-1 * var(--dt-color-hover-01));
+        --card-bg-color: hsl(var(#{$bg-color}-h), var(#{$bg-color}-s), calc(var(#{$bg-color}-l) + var(--hover-modifier)));
+    }
+
+    &:active:not([disabled]) {
+        --active-modifier: calc(-1 * var(--dt-color-active-01));
+        --card-bg-color: hsl(var(#{$bg-color}-h), var(#{$bg-color}-s), calc(var(#{$bg-color}-l) + var(--active-modifier)));
+    }
+}
+
+// By default, the card is styled to the 'Default' variant designs
+.c-card-container {
+    --card-bg-color: var(--dt-color-container-default);
+    --card-radius: var(--dt-radius-rounded-c);
+    --card-shadow: var(--dt-elevation-01);
+
+    display: block;
+    color: currentcolor;
+    background-color: var(--card-bg-color);
+    border-radius: var(--card-radius);
+    text-decoration: none;
+    position: relative;
+
+    @include card-interactive-states('--dt-color-container-default');
+
+    &[disabled] {
+        --card-bg-color: var(--dt-color-disabled-01);
+        pointer-events: none;
+    }
+
+    // We should not apply the default elevation styling if the card is focused
+    // but we want to make sure that it is still applied when the card is clicked
+    &:not(:focus-within),
+    &:active:focus-within {
+        box-shadow: var(--card-shadow);
+    }
+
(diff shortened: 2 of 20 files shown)


packages/components/pie-card-container/src/card-container.scss
```diff
@@ -1 +1,95 @@
...
+        */
+        box-shadow: 0 0 0 2px var(--dt-color-focus-inner);
+        outline: none;
+
+        &:after {
+            content: '';
+            display: block;
+
+            /*
+            The pseudo element has to be larger than the element itself, but not just by
```
FayeCarter, line 75: Would `box-sizing: border-box;` on the `a` help with this?

FayeCarter: and could you use inset?

packages/components/pie-card-container/src/index.ts
```diff
@@ -1,16 +1,63 @@
...
+            variant,
+        } = this;
+
+        return html`
+                <div
+                    variant=${variant}
+                    class="c-card-container"
+                    data-test-id="pie-card-container"
+                    ?disabled=${disabled}>
+                    <a
```
raoufswe, line 49: should this be conditional only if the href is available?
maybe we can consider using a similar approach to the link component?

packages/components/pie-card-container/src/index.ts
```diff
@@ -1,16 +1,63 @@
...
+                        data-test-id="pie-card-container-link"
+                        href=${href || ''}
```
raoufswe, line 51: should not this be defaulted to nothing if not passed?
```suggestion
                        href=${href || nothing}
```

packages/components/pie-card-container/src/card-container.scss
```diff
@@ -1 +1,95 @@
...
+
+    // We should not apply the default elevation styling if the card is focused
+    // but we want to make sure that it is still applied when the card is clicked
+    &:not(:focus-within),
+    &:active:focus-within {
+        box-shadow: var(--card-shadow);
+    }
+
+    // We want to stretch the <a> tag over the entire card to make the entire component interactive
+    > a {
```
raoufswe, line 43: would these styles get applied if the card itself isn't a link, but part of its content has a link? like
<div>
fancy card, <a>Learn more about it</a>
</div>

packages/components/pie-card-container/src/card-container.scss
raoufswe: can you also please create a ticket and reference it in the comment? 🥇

packages/components/pie-card-container/src/card-container.scss
```diff
@@ -1 +1,95 @@
...
+            box-shadow: 0 0 0 2px var(--dt-color-focus-outer);
+        }
+    }
+}
+
+.c-card-container-slot {
+    // Ensures that any slot containers are also rounded
+    border-radius: inherit;
+    overflow: hidden;
+}
```
raoufswe, line 95: sorry I tried to understand what this does exactly by commenting on the styles / the div and I don't see a difference. could you please share more details?

packages/components/pie-card-container/src/card-container.scss
jamieomaguire: Nope :) because it is only selecting the immediate child anchor

packages/components/pie-card-container/src/card-container.scss
jamieomaguire: @FayeCarter unsure but I'm keeping it consistent with the other focus styles across the component library (this is adapted from pie-button) until we swap it out with a generic implementation in pie-css.

@raoufswe Ash and I are creating tickets for this very soon :)

packages/components/pie-card-container/src/card-container.scss
jamieomaguire: Sure! Try adding a background colour to your slot content and you'll see it doesn't respec the rounded corners of the card :)

packages/components/pie-card-container/src/index.ts
jamieomaguire: I think for a starting point this is ok - the original ticket had no mention of link behaviour (but it is required to add interactive states) so I would prefer not to add too much logic and instead refine the desired behaviours with Kevin and create more tickets.

This is just a very, very basic starting point and can be improved on as we scope out the required behaviours

packages/components/pie-card-container/src/index.ts
jamieomaguire: I wish it could but TS does not allow it!

Screenshot 2023-09-05 at 08 16 00

packages/components/pie-card-container/src/card-container.scss
jamieomaguire: Screenshot 2023-09-05 at 08 29 52

@raoufswe hopefully this example helps?

packages/components/pie-card-container/src/card-container.scss
jamieomaguire: @raoufswe that said, I do not believe the example you've created is one the card designs currently support, as they all require the entire card to be clickable, rather than nested content

packages/components/pie-card-container/src/card-container.scss
raoufswe: yes makes sense 💯

packages/components/pie-card-container/README.md
```diff
@@ -59,16 +59,22 @@ import { PieCardContainer } from '@justeattakeaway/pie-card-container/dist/react
...
+| rel | `string` | `undefined` | What the relationship of the linked URL is (this will not take effect unless the card is a link). |
+| aria | `object` | `undefined` | The ARIA labels used for various parts of the card. |
+
 
 In your markup or JSX, you can then use these to set the properties for the `pie-card-container` component:
 
 ```html
 <!-- Native HTML -->
-<pie-card-container></pie-card-container>
+<pie-card-container disabled href="/foo/bar" rel="noopener" target="_blank"></pie-card-container>
```
xander-marjoram, line 74: No need to change anything but I learned this the other day:
> Setting `target="_blank"` on `<a>` elements implicitly provides the same rel behavior as setting `rel="noopener"`

packages/components/pie-card-container/src/index.ts
raoufswe: aha I see, I thought the anchor was our component and was wondering why you can't. 
I believe we can replace it later when  we conditionally render the anchor / div

xander-marjoram (approved):

packages/components/pie-card-container/src/card-container.scss
jamieomaguire: @raoufswe I don't think those will be cards, probably worth catching up with Naz as I am just quoting the figma designs :)

jamieomaguire: @raoufswe I will create a ticket to look into this and send to @kevinrodrigues

kevinrodrigues (dismissed):

raoufswe (dismissed):

raoufswe (approved):

PR #791, opened 2023-09-11
feat(pie-card-container): DSW-1192 support link and button behavior

raoufswe:
[Added] Button and Link behavior to the card container

### Reviewer 1

### Reviewer 2

### DIFF
--- a/packages/components/pie-card-container/src/index.ts
+++ b/packages/components/pie-card-container/src/index.ts
@@ -1,10 +1,13 @@
 import {
-    html, LitElement, unsafeCSS, nothing,
+    LitElement, unsafeCSS, nothing, TemplateResult,
 } from 'lit';
+import { html } from 'lit/static-html.js';
 import { property } from 'lit/decorators.js';
 import { validPropertyValues } from '@justeattakeaway/pie-webc-core';
 import styles from './card-container.scss?inline';
-import { CardContainerProps, type AriaProps, variants } from './defs';
+import {
+    CardContainerProps, type AriaProps, variants, interactionTypes,
+} from './defs';
 
 // Valid values available to consumers
 export * from './defs';
@@ -12,6 +15,10 @@ export * from './defs';
 const componentSelector = 'pie-card-container';
 
 export class PieCardContainer extends LitElement implements CardContainerProps {
+    @property()
+    @validPropertyValues(componentSelector, interactionTypes, 'none')
+    public interactionType: CardContainerProps['interactionType'] = 'none';
+
     @property()
     @validPropertyValues(componentSelector, variants, 'default')
     public variant: CardContainerProps['variant'] = 'default';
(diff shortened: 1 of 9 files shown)


packages/components/pie-card-container/src/index.ts
```diff
@@ -33,31 +38,34 @@ export class PieCardContainer extends LitElement implements CardContainerProps {
...
-            disabled,
-            variant,
         } = this;
 
+        const isButton = tag === 'button';
+        const element = unsafeStatic(isButton ? 'div' : tag);
+
         return html`
-                <div
+                <${element}
```
jamieomaguire, line 53: @raouf I don't think we should use a button here. It goes against the HTML spec of how a button should be used.

Note here: https://html.spec.whatwg.org/multipage/form-elements.html#the-button-element

The button element should only contain _phrasing content_. Our slots will allow users to add (perhaps quite often) Flow content. 

Happy to discuss further offline :)

raoufswe: Hi @jamieomaguire 
good point. we aren't using a div with the needed attributes to make sure it behaves as a button as mentioned in the ticket. 

```
const element = unsafeStatic(isButton ? 'div' : tag);
```

jamieomaguire: 🤦 I need my morning coffee. Sorry @raoufswe!

I do have another comment - it seems that this kind of dynamic element interpolation breaks not only syntax highlighting but potentially Lit binding/ts validation. Please see this example. Note what value I have provided for href in the conditional.

Screenshot 2023-09-12 at 08 55 32

jamieomaguire: I know its long winded but please can we just render the two HTML variants separately in an if else? It will preserve highlighting etc even though its a bit long winded

raoufswe: updated to render different templates based on the interaction type 💯

jamieomaguire:
LGTM but please add a visual test for each interaction type so they don't regress? :)

jamieomaguire (approved):

FayeCarter (approved):

packages/components/pie-card-container/src/index.ts
```diff
@@ -31,30 +38,51 @@ export class PieCardContainer extends LitElement implements CardContainerProps {
     @property({ type: Object })
     public aria!: AriaProps;
 
+    /**
+     * Renders the card as an anchor element.
+     *
+     * @private
+     */
+    private renderAnchor (): TemplateResult {
```
kevinrodrigues, line 46: Sorry i was off on friday so might have missed this PR. I was thinking could we maybe having this function return a template based on the interaction type or "tag" instead? The markup looks quite similar to the button one. Something like `renderCardTemplate` this can be a generic method that handles the interaction type or tag setting all the props then inside the main render we can call either or based on user input.

PR #818, opened 2023-09-19
chore(pie-modal): DSW-1203 fix modal ssr issues

jamieomaguire:

Numerous changes to fix modal SSR issues:

---
---

[Changed] - Move any browser-based references to lifecycle hooks that are not run on the server. This is to fix SSR issues seen in NextJS integrations where calling things like `document.addEventListener` in places such as the constructor (which is called during SSR and CSR) causes errors to occur.

---
---

[Changed] - Defer importing the `dialog-polyfill` package to when the component is running in the browser. The package is not universal code, meaning it contains browser references and this code is run by simply importing the package. This breaks SSR in applications such as NextJS. Using a dynamic import (whilst not ideal) solves this issue by preventing the client-only code run running on the server.

---
---

[Changed] - Rewrite RTL mixin to remove the dir property. The LitElement base class is a subclass of HTMLElement, so a Lit component inherits all of the standard HTMLElement properties and methods. This means that the dir property is already available on the component and we don't need to add it again. [Reference](https://lit.dev/docs/components/defining/#a-lit-component-is-an-html-element). During SSR, if no dir is provided, it will be inferred from the document.documentElement once it's instantiated on the client.

---

---

[Changed] - Removed the DependantMap type and replaced all references with Lit's own PropertyValues helper. This provides exactly the same strongly typed map for a component's properties which makes our own type a little redundant. [Reference](https://lit.dev/docs/components/lifecycle/#typescript-types-for-changedproperties)

---
"pie-storybook": minor
---

[Changed] - Deleted the dir props and i18nargtypes helper as `dir` is no longer a Lit component property. It is simply a regular HTML attribute

---
"@justeattakeaway/pie-toggle-switch": minor
---

[Changed] - use `isRTL` attribute provided by the `RTLMixin` rather than `dir` for rtl scss styles


packages/components/pie-modal/src/index.ts
```diff
@@ -100,13 +98,9 @@ export class PieModal extends RtlMixin(LitElement) implements ModalProps {
...
     static styles = unsafeCSS(styles);
 
-    constructor () {
-        super();
-        this.addEventListener('click', (event) => this._handleDialogLightDismiss(event));
-    }
-
     connectedCallback () : void {
         super.connectedCallback();
+        this.addEventListener('click', (event) => this._handleDialogLightDismiss(event));
```
raoufswe, line 103: <3

apps/pie-storybook/stories/pie-modal.stories.ts
```diff
@@ -192,7 +191,6 @@ const BaseStoryTemplate = (props: ModalProps) : TemplateResult => {
         <pie-button @click=${toggleModal}>Toggle Modal</pie-button>
         <pie-modal
             .aria="${aria}"
-            dir="${dir || nothing}"
```
jamieomaguire, line 195: @raoufswe I plan to raise an immediate follow-up PR to update the component generator to no longer add the RTL component props stuff (as we won't need it if this change goes in).

I just want to keep this PR from getting out of hand 😄

packages/components/pie-modal/src/index.ts
```diff
@@ -119,8 +112,11 @@ export class PieModal extends RtlMixin(LitElement) implements ModalProps {
         super.disconnectedCallback();
     }
 
-    firstUpdated (changedProperties: DependentMap<ModalProps>) : void {
+    async firstUpdated (changedProperties: PropertyValues<this>) : Promise<void> {
+        super.firstUpdated(changedProperties);
```
raoufswe, line 116: just curious, why we had to call super.firstUpdated here?
same goes for         super.updated(changedProperties);

raoufswe:
looks really good 🔥

raoufswe (approved):

packages/components/pie-modal/src/index.ts
jamieomaguire: I am unsure if the order of the super call is important within the block. However, my understanding and experience is that not calling the base class super method leads to weird behaviours with lifecycle methods. It's the same reason you must always call the super constructor.

packages/components/pie-webc-core/src/mixins/rtl/rtlMixin.ts
```diff
@@ -1,55 +1,80 @@
...
+            get isRTL (): boolean {
+                if (this.dir) {
+                    return this.dir === 'rtl';
+                }
+
+                // This check is necessary as document is undefined during SSR
+                if (isServer) {
+                    // Inference from the document's root element is not available during SSR.
+                    return false;
+                }
```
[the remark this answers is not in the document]
jamieomaguire: great spot! updated :)

raoufswe (approved):

LTurns (approved):

kevinrodrigues (approved):

PR #820, opened 2023-09-20
feat(pie-form-label): DSW-1188 build form label

raoufswe:
[Added] - Build the Pie Form Label component

📓 as a result of [ this discussion](https://github.com/justeattakeaway/pie/discussions/782), we agreed to pass the optional / trailing labels as named slots to the component, however, that didn't seem to be doable during implementation as slots aren't moved in the DOM tree, but they're rendered as if they were children of the `<slot>` so when the parent element (label) is a flex container, optional and trailing slots won't be considered 'flex children' as they aren't a direct child. 

Besides, marking them as a slot will require end users to set the html element alongside the slot such as 
```
<pie-form-label>
Label
<span slot='trailing'> X out X </span>
</pie-form-label>
```
and that's going to be tricky to validate since we only want to accept text, so it's easier to lock it down as a prop and avoid the overhead of validating HTML elements, figuring out how to ensure the passed markup is positioned correctly in the component layout etc.

### Reviewer 1

### Reviewer 2

### DIFF
--- a/packages/components/pie-form-label/src/form-label.scss
+++ b/packages/components/pie-form-label/src/form-label.scss
@@ -1,10 +1,5 @@
 @use '@justeattakeaway/pie-css/scss' as p;
 
-*,
-*:before,
-*:after {
-    box-sizing: border-box;
-}
 
 .pie-form-label {
     --form-label-font-size: #{p.font-size(--dt-font-size-14)};
--- a/packages/components/pie-form-label/src/form-label.scss
+++ b/packages/components/pie-form-label/src/form-label.scss
@@ -1,1 +1,40 @@
 @use '@justeattakeaway/pie-css/scss' as p;
+
+*,
+*:before,
+*:after {
+    box-sizing: border-box;
+}
+
+.pie-form-label {
+    --form-label-font-size: #{p.font-size(--dt-font-size-14)};
+    --form-label-line-height: calc(var(--dt-font-size-14-line-height) * 1px);
+    --form-label-font-weight: var(--dt-font-weight-bold);
+    --form-label-color: var(--dt-color-content-default);
+    --form-label-offset: var(--dt-spacing-a);
+
+    display: flex;
+    justify-content: space-between;
+    gap: var(--dt-spacing-d);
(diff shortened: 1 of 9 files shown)


packages/components/pie-form-label/src/form-label.scss
```diff
@@ -1 +1,33 @@
...
+.pie-form-label {
+    --form-label-font-size: #{p.font-size(--dt-font-size-14)};
+    --form-label-line-height: calc(var(--dt-font-size-14-line-height) * 1px);
+    --form-label-font-weight: var(--dt-font-weight-bold);
+    --form-label-color: var(--dt-color-content-default);
+
+    display: flex;
+    justify-content: space-between;
+    gap: var(--dt-spacing-d);
+    width: 100%;
```
jamieomaguire, line 12: display flex should make this behave like a block element, so I believe width: 100% may be redundant

packages/components/pie-form-label/src/form-label.scss
```diff
@@ -1 +1,33 @@
 @use '@justeattakeaway/pie-css/scss' as p;
+
+.pie-form-label {
```
jamieomaguire, line 3: I would name this `c-formLabel` so we can use the hyphens to denote hierarchy.

We should stick to using `.c-` format for consistency with other components (see pie-button, modal, toggle-switch). The use of `pie` is probably unnecessary too;

jamieomaguire (dismissed):

jamieomaguire:
@raoufswe just curious, should there be any cursor styles applied to the label?

Looks great! 🔥

raoufswe:
> @raoufswe just curious, should there be any cursor styles applied to the label?
> 

I am not sure, maybe @tingting-lau can help with this?

tingting-lau:
Hey @raoufswe @jamieomaguire, it should be using default cursor over the text, and pointer over the tooltip icon.

kevinrodrigues (approved):

kevinrodrigues (approved):

jamieomaguire (approved):

PR #829, opened 2023-09-22
feat(pie-card-container): DSW-1157 Card container padding props settings

kevinrodrigues:
[Added] - Card container padding props

### Reviewer 1

### Reviewer 2


packages/components/pie-card-container/src/defs.ts
```diff
@@ -47,4 +49,14 @@ export interface CardContainerProps {
      * What the interaction intent of the card should be such as anchor, button or regular div (default).
      */
      interactionType?: typeof interactionTypes[number];
+
+    /**
+     * An array of padding values from a to g.
+     */
+    padding: PaddingValues | PaddingValues[];
```
raoufswe, line 56:
```suggestion
    padding: PaddingValue | PaddingValue[];
```

apps/pie-storybook/stories/pie-card-container.stories.ts
```diff
@@ -59,6 +60,26 @@ const cardContainerStoryMeta: CardContainerStoryMeta = {
                 summary: 'default',
             },
         },
+        padding: {
```
raoufswe, line 63: I would rename this to `paddingY` as it isn't clear if this will set the x and y or the four sides.

apps/pie-storybook/stories/pie-card-container.stories.ts
```diff
@@ -59,6 +60,26 @@ const cardContainerStoryMeta: CardContainerStoryMeta = {
...
+            control: {
+                type: 'text',
+            },
+            options: padding,
+            defaultValue: {
+                summary: 'none',
+            },
+        },
+        paddingX: {
+            description: 'Set the padding of the card.',
```
raoufswe, line 74: maybe we can elaborate here more such as 

```suggestion
            description: 'Set the left / right padding of the card.',
```

apps/pie-storybook/stories/pie-card-container.stories.ts
```diff
@@ -59,6 +60,26 @@ const cardContainerStoryMeta: CardContainerStoryMeta = {
...
             },
         },
+        padding: {
+            description: 'Set the padding of the card.',
+            control: {
+                type: 'text',
+            },
+            options: padding,
+            defaultValue: {
+                summary: 'none',
```
raoufswe, line 70: none isn't a valid value for your type definition, maybe we can default it to '' or remove the defaultValue  here at all since the prop is optional

packages/components/pie-card-container/src/index.ts
```diff
@@ -40,12 +40,20 @@ export class PieCardContainer extends LitElement implements CardContainerProps {
     @property({ type: Boolean })
     public isDraggable = false;
 
+    @property({ type: String, reflect: true })
+    public padding!: PaddingValues;
+
+    @property({ type: String, reflect: true })
+    public paddingX!: PaddingValues;
+
```
raoufswe, line 48: imo, these props should be marked as optional

```suggestion
    @property({ type: String, reflect: true })
    public padding?: PaddingValues;

    @property({ type: String, reflect: true })
    public paddingX?: PaddingValues;

```

the ! symbol or _definite assignment assertion operator_ indicates that the value must be provided beforehand by users and it isn't the case for padding.

packages/components/pie-card-container/src/index.ts
```diff
@@ -58,12 +66,42 @@ export class PieCardContainer extends LitElement implements CardContainerProps {
                 target=${this.target || nothing}
                 rel=${this.rel || nothing}
                 aria-label=${this.aria?.label || nothing}
-                aria-disabled=${this.disabled ? 'true' : 'false'}>
+                aria-disabled=${this.disabled ? 'true' : 'false'}
+                style=${paddingCSS ? `padding: ${paddingCSS}` : nothing}>
```
raoufswe, line 70: do you think we can return `padding: ${paddingCSS}` from `generatePaddingCSS()`; directly? so we control the whole generation from one place.  

```suggestion
                style=${paddingCSS ? paddingCSS  : nothing}>
```

apps/pie-storybook/stories/pie-card-container.stories.ts
kevinrodrigues: This value provides the padding for all the way around the component if it is the only value provided by the consumer I was initially going to call it Y but left it for that reason, thoughts?

packages/components/pie-card-container/src/index.ts
kevinrodrigues: Good catch!

packages/components/pie-card-container/src/index.ts
```diff
@@ -58,12 +66,42 @@ export class PieCardContainer extends LitElement implements CardContainerProps {
...
             </a>`;
     }
 
+    /**
+     * Generates padding for the component based on `padding` or `paddingX` values passed
+     * by the consumer.
+     *
+     * @private
+     */
+    private generatePaddingCSS (): string {
```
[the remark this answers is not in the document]
kevinrodrigues: I think potentially we could via a mixin / helper although i'm not sure how it would work here in this file, we might need to pass a class in and use that as a hook to apply those styles.

packages/components/pie-card-container/src/defs.ts
```diff
@@ -47,4 +49,14 @@ export interface CardContainerProps {
...
+
+    /**
+     * An array of padding values from a to g.
+     */
+    padding?: PaddingValue | PaddingValue[];
+
+    /**
+     * An array of padding values from a to g.
+     */
+    paddingX?: PaddingValue | PaddingValue[];
```
ashleynolan, line 61: What's the difference between `padding` and `paddingX`?

ashleynolan: Just wondering as well – could this be handled with a single padding prop? (similar to how CSS does with `padding: a b c d;`?

kevinrodrigues: So i did initially try that but the issue is that our valid prop function (the thing used to make sure we pass in valid values) wont allow a mixture of different values so i split it out (similar to what snacks does). The `padding` one is for a single value which can go all the way around the component i.e `padding: 5px` and the `paddingX` is one that if supplied will apply to left and right so if both are used together then it would output like `padding: 5px 10px` 10px being the paddingX value. If only `paddingX` is provided and not `padding` then no padding is rendered.

kevinrodrigues: Fixed.

packages/components/pie-card-container/README.md
```diff
@@ -57,16 +57,17 @@ import { PieCardContainer } from '@justeattakeaway/pie-card-container/dist/react
...
+| href            | `string`  | `undefined` | The URL that the card should point to (this will not take effect unless the card is a link).                                                       |
+| target          | `string`  | `undefined` | Where to display the linked URL such as _self, _blank, _parent or _top (this will not take effect unless the card is a link).                      |
+| rel             | `string`  | `undefined` | What the relationship of the linked URL is (this will not take effect unless the card is a link).                                                  |
+| aria            | `object`  | `undefined` | The ARIA labels used for various parts of the card.                                                                                                |
+| isDraggable     | `boolean` | `false`     | Sets a grab/grabbing cursor when set to true.                                                                                                      |
+| padding         | `String`  | `none`      | Sets a given padding of either a single or two values in string format separated by commas (if more than one padding value). e.g `'a'` or `'a, b'` |
```
ashleynolan, line 70: I would just add that setting single value adds padding to all sides and 2 values is for `leftRight, topBottom` padding

packages/components/pie-card-container/src/index.ts
```diff
@@ -58,12 +63,46 @@ export class PieCardContainer extends LitElement implements CardContainerProps {
...
                     <slot></slot>
                 </div>
             </a>`;
     }
 
+    /**
+     * Generates padding for the component based on `padding` values passed
+     * by the consumer.
+     *
+     * Example: 'a' or 'a, b'
```
ashleynolan, line 77: Same as in the docs, I would just put here that single applies it to all sides and `a, b` as `leftRight, topBottom`

packages/components/pie-card-container/test/component/pie-card-container.spec.ts
```diff
@@ -229,4 +229,47 @@ test.describe('PieCardContainer - Component tests', () => {
             });
         });
     });
+
+    test.describe('Prop: `padding` & `PaddingX`', () => {
```
ashleynolan, line 233: Guessing this can just be:

```suggestion
    test.describe('Prop: `padding`', () => {
```

packages/components/pie-card-container/test/visual/pie-card-container.spec.ts
```diff
@@ -14,11 +14,13 @@ import {
...
 import { percyWidths } from '@justeattakeaway/pie-webc-testing/src/percy/breakpoints.ts';
 import { PieCardContainer } from '@/index';
-import { interactionTypes, variants } from '@/defs';
+import { interactionTypes, variants, padding } from '@/defs';
 
 const props: PropObject = {
     interactionType: interactionTypes,
     variant: variants,
+    padding,
+    paddingX: padding,
```
ashleynolan, line 23:
```suggestion
    padding,
```

packages/components/pie-card-container/src/index.ts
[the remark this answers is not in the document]
kevinrodrigues: Shall we create a ticket for this?

packages/components/pie-card-container/README.md
```diff
@@ -57,16 +57,17 @@ import { PieCardContainer } from '@justeattakeaway/pie-card-container/dist/react
...
+| rel             | `string`  | `undefined` | What the relationship of the linked URL is (this will not take effect unless the card is a link).                                                                                                                                               |
+| aria            | `object`  | `undefined` | The ARIA labels used for various parts of the card.                                                                                                                                                                                             |
+| isDraggable     | `boolean` | `false`     | Sets a grab/grabbing cursor when set to true.                                                                                                                                                                                                   |
+| padding         | `String`  | `none`      | Sets a given padding of either a single value or two values in string format separated by commas. setting a single value adds padding to all sides of the component and 2 values is for top, bottom, left, right padding. e.g `'a'` or `'a, b'` |
```
ashleynolan, line 70:
```suggestion
| padding         | `String`  | `none`      | Sets a given padding of either a single value or two values in string format separated by commas. Setting a single value adds padding to all sides of the component and 2 values is for top, bottom, left, right padding. e.g `'a'` or `'a, b'` |
```

packages/components/pie-card-container/src/index.ts
```diff
@@ -58,12 +63,48 @@ export class PieCardContainer extends LitElement implements CardContainerProps {
...
             </a>`;
     }
 
+    /**
+     * Generates padding for the component based on `padding` values passed
+     * by the consumer.
+     *
+     *
+     * Example: 'a' or 'a, b'
+     * Single values i.e `'a'` applies to all sides and `'a, b'` applies to: top, right, bottom, left
```
ashleynolan, line 79:
```suggestion
     * Single values i.e `'a'` applies to all sides and `'a, b'` applies to: top & bottom, left & right
```

packages/components/pie-card-container/src/index.ts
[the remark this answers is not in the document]
ashleynolan: I'm ok with it living here for now and if we ever need it elsewhere, pulling it out at that stage?

packages/components/pie-card-container/README.md
```diff
@@ -57,16 +57,17 @@ import { PieCardContainer } from '@justeattakeaway/pie-card-container/dist/react
 
 ## Props
 
-| Property        | Type      | Default     | Description                                                                                                                  |
-|---|---|-------------|------------------------------------------------------------------------------------------------------------------------------|
+| Property        | Type      | Default     | Description                                                                                                                                                                                                                                     |
+|-----------------|-----------|-------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
```
xander-marjoram, line 61: Was this done manually or was it an automatic thing from your IDE? I'm not a big fan of making the markdown all aligned because it makes it harder to spot changes when you're reviewing

packages/components/pie-card-container/src/index.ts
```diff
@@ -58,12 +63,48 @@ export class PieCardContainer extends LitElement implements CardContainerProps {
...
+            return '';
+        }
+
+        const paddingArray = padding
+            .split(',')
+            .map((item) => item.trim())
+            .filter((value) => value !== '');
+
+        if (paddingArray.length > 0) {
+            paddingCSS += `var(--dt-spacing-${paddingArray[0]})`;
```
xander-marjoram, line 98: Do we want to validate that the values create valid tokens? Thinking in theory someone could try and pass something like `a); background-image: url(\"https://malicious.com/image.png\");` which would get through the processing we do above.

xander-marjoram: Hard to exploit in practice, I know, but I think we know that all of the valid values for this array are the letters `a-j`, so we could validate that with a regex, perhaps

packages/components/pie-card-container/src/index.ts
```diff
@@ -58,12 +63,48 @@ export class PieCardContainer extends LitElement implements CardContainerProps {
...
+
+            if (paddingArray.length > 1) {
```
xander-marjoram, line 100: Do we only support lengths of 1 or 2?

packages/components/pie-card-container/README.md
kevinrodrigues: Was IDE i think. Will fix it.

packages/components/pie-card-container/src/index.ts
kevinrodrigues: Yup only two values max. Min 1.

xander-marjoram: Okay great, could we have a unit test for >2 values?

packages/components/pie-card-container/src/index.ts
kevinrodrigues: Should be fixed now.

.changeset/pretty-pots-rush.md
```diff
@@ -0,0 +1,6 @@
+---
+"@justeattakeaway/pie-card-container": minor
+"pie-storybook": minor
+---
+
+[Added] - Card container padding props
```
jamieomaguire, line 6: please could you update the PR title to reflect these changes? :)

packages/components/pie-card-container/src/index.ts
```diff
@@ -40,12 +40,17 @@ export class PieCardContainer extends LitElement implements CardContainerProps {
     @property({ type: Boolean })
     public isDraggable = false;
 
+    @property({ type: String, reflect: true })
```
jamieomaguire, line 43: please add a TODO comment here to add the updated valid property values decorator and add the ticket number to the comment for reference :)

packages/components/pie-card-container/src/defs.ts
```diff
@@ -47,4 +49,9 @@ export interface CardContainerProps {
      * What the interaction intent of the card should be such as anchor, button or regular div (default).
      */
      interactionType?: typeof interactionTypes[number];
+
+    /**
+     * An array of padding values from a to g.
```
raoufswe, line 54: i think we need to update the description here, maybe just copy what we have in the props table

packages/components/pie-card-container/README.md
```diff
@@ -57,16 +57,17 @@ import { PieCardContainer } from '@justeattakeaway/pie-card-container/dist/react
...
+| rel             | `string`  | `undefined` | What the relationship of the linked URL is (this will not take effect unless the card is a link).                                                                                                                                               |
+| aria            | `object`  | `undefined` | The ARIA labels used for various parts of the card.                                                                                                                                                                                             |
+| isDraggable     | `boolean` | `false`     | Sets a grab/grabbing cursor when set to true.                                                                                                                                                                                                   |
+| padding         | `String`  | `none`      | Sets the padding of the card container. Can be either a single value or two values separated by commas. Setting a single value adds padding to all sides of the card, whereas setting two values will set the "topBottom, leftRight" padding. e.g `'a'` or `'a, b'` |
```
raoufswe, line 70: i think the default value is `undefined` not, `none`, as far as the prop type goes

apps/pie-storybook/stories/pie-card-container.stories.ts
```diff
@@ -130,6 +139,7 @@ const Template: TemplateFunction<CardContainerProps> = ({
             rel="${rel || nothing}"
             ?disabled="${disabled}"
             .aria="${aria}"
+            padding="${padding}"
```
raoufswe, line 142:
```suggestion
            padding="${padding || nothing}"
```

raoufswe (approved):
just a small nits and should be good to go!
good job Kev

packages/components/pie-card-container/src/index.ts
```diff
@@ -58,12 +63,52 @@ export class PieCardContainer extends LitElement implements CardContainerProps {
...
+        const paddingArray = padding
+            .split(',')
+            .map((item) => item.trim())
+            .filter((value) => /^[a-g]$/.test(value));
+
+        if (paddingArray.length > 0 && paddingArray.length <= 2) {
+            paddingCSS += `var(--dt-spacing-${paddingArray[0]})`;
+
+            if (paddingArray.length > 1) {
+                paddingCSS += ` var(--dt-spacing-${paddingArray[1]})`;
```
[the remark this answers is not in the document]
jamieomaguire: A more complete example (not tested):

```
private generatePaddingCSS(): string {
    const { padding } = this;
    let paddingCSS = '';
    
    type PaddingValues = 'a' | 'b' | 'c' | 'd' | 'e' | 'f' | 'g';
    
    const validPaddingTokenValues: Record<PaddingValues, string> = {
        a: 'var(--dt-spacing-a)',
        b: 'var(--dt-spacing-b)',
        c: 'var(--dt-spacing-c)',
        d: 'var(--dt-spacing-d)',
        e: 'var(--dt-spacing-e)',
        f: 'var(--dt-spacing-f)',
        g: 'var(--dt-spacing-g)'
    };
    
    if (!padding) return '';
    
    const paddingArray = padding
        .split(',')
        .map((item) => item.trim())
        .filter((value) => validPaddingTokenValues.hasOwnProperty(value));
    
    if (paddingArray.length > 0 && paddingArray.length <= 2) {
        paddingCSS += validPaddingTokenValues[paddingArray[0] as PaddingValues];
        
        if (paddingArray.length > 1) {
            paddingCSS += ` ${validPaddingTokenValues[paddingArray[1] as PaddingValues]}`;
        }
    }
    
    if (!paddingCSS) return '';
    
    return `padding: ${paddingCSS};`;
}
```

packages/components/pie-card-container/src/index.ts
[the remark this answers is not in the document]
kevinrodrigues: Will remove, earlier SB wasn't showing the changes when you change it but now it looks fine.

packages/components/pie-card-container/src/defs.ts
kevinrodrigues: Fixed

packages/components/pie-card-container/README.md
kevinrodrigues: Fixed.

apps/pie-storybook/stories/pie-card-container.stories.ts
kevinrodrigues: Fixed.

packages/components/pie-card-container/src/index.ts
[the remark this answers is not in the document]
kevinrodrigues: I think if we update the valid properties method we wont need to do this as we will validate the input via that method? For now i think the regex should be fine until we have that method which we can then swap out once complete.

packages/components/pie-card-container/src/index.ts
kevinrodrigues: Fixed.

packages/components/pie-card-container/test/component/pie-card-container.spec.ts
```diff
@@ -229,4 +229,126 @@ test.describe('PieCardContainer - Component tests', () => {
...
+                });
+
+                const component = page.locator(componentSelector);
+                const componentAttribute = await component.getAttribute('style');
+
+                // Assert
+                expect(componentAttribute).toBe('padding: var(--dt-spacing-a)');
+            });
+
+            padding.forEach((padding) => {
```
xander-marjoram, line 253: Variable name clash here, surprised it's not causing any issues?

raoufswe (approved):

packages/components/pie-card-container/test/component/pie-card-container.spec.ts
kevinrodrigues: Fixed.

raoufswe (approved):

packages/components/pie-card-container/src/index.ts
```diff
@@ -58,12 +68,52 @@ export class PieCardContainer extends LitElement implements CardContainerProps {
...
+
+        // Check if padding is empty (null, undefined, or an empty string)
+        if (!padding) {
+            return '';
+        }
+
+        const paddingArray = padding
+            .split(',')
+            .map((item) => item.trim())
+            .filter((value) => /^[a-g]$/.test(value));
```
xander-marjoram, line 100: Does `a, z` result in just a value of `a`?

kevinrodrigues: Yup that's correct.

xander-marjoram (approved):

packages/components/pie-card-container/src/index.ts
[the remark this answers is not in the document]
jamieomaguire: @kevinrodrigues for what it's worth, even with validation I would still suggest this comment.

I think as a rule of thumb we should always avoid injecting user provided data into our code. It's much safer. The validation helps shape error messages and graceful fallbacks. Providing a layer between our component and user data is the security aspect.  

I don't think we should just rely on the validation for this.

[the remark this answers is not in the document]
kevinrodrigues: I'm struggling to work out how this could be exploited. (Sorry not trying to be difficult here) If there is a valid scenario where i've missed something then happy to add this change in but i dont want to add it in for the sake of it. I think Xander's example was a good catch where someone could add in a string and render a background image (which is now fixed) but with this one i think i need an example to show what else can be done. We use string inputs in other areas in our components and unless im mistaken wont those be the same problem you're highlighting here too?

PR #849, opened 2023-10-02
fix(pie-toggle-switch): DSW-1123 address design feedback

raoufswe:
[Fixed] - The label should have the same cursor as the control itself

[Fixed]- The cursor should be disabled, following the same pattern as the Button

### Reviewer 1

### Reviewer 2

### DIFF
--- a/packages/components/pie-toggle-switch/src/index.ts
+++ b/packages/components/pie-toggle-switch/src/index.ts
@@ -58,14 +58,13 @@ export class PieToggleSwitch extends RtlMixin(LitElement) implements ToggleSwitc
      * @private
      */
     private renderToggleSwitchLabel (): TemplateResult {
-        const { label, labelPlacement } = this;
+        const { label } = this;
 
-        if (label && labelPlacement && (labelPlacements.includes(labelPlacement))) {
+        if (label) {
             return html`
                 <label
                     for="toggle-switch"
-                    data-test-id="toggle-switch-label"
-                    class="c-toggleSwitch-label c-toggleSwitch--${labelPlacement}">
+                    data-test-id="toggle-switch-label">
                     ${label}
                 </label>`;
         }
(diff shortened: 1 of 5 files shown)


packages/components/pie-toggle-switch/src/index.ts
```diff
@@ -58,14 +58,13 @@ export class PieToggleSwitch extends RtlMixin(LitElement) implements ToggleSwitc
      * @private
      */
     private renderToggleSwitchLabel (): TemplateResult {
-        const { label, labelPlacement } = this;
+        const { label } = this;
 
-        if (label && labelPlacement && (labelPlacements.includes(labelPlacement))) {
```
raoufswe, line 63: this check is unnecessary as the validation layer is doing that and defaulting to the default value if an invalid value is passed.

packages/components/pie-toggle-switch/test/component/pie-toggle-switch.spec.ts
```diff
@@ -182,7 +182,7 @@ test.describe('Component: `Pie toggle switch`', () => {
 
     test.describe('Props: `LabelProps`', () => {
         test.describe('when a position is passed as `leading`', () => {
-            test('should render a leading label with class `c-toggleSwitch--leading`', async ({ mount }) => {
+            test('should render a leading label', async ({ mount }) => {
```
xander-marjoram, line 185: To me it looks like these tests check for the label but not that it's in the correct position?

raoufswe: Initially, the label element had a class name that indicated its position, allowing us to adjust its margin either to the right or left. but I updated the implementation to use the "gap" property instead. thus, the class name + test case isn't feasible anymore but we still ensured that the label container is defined if you pass the prop. 

Besides, I believe this type of test should be done in visual tests rather than component tests.

xander-marjoram: If the DOM order changes based on whether it's leading or trailing, then I think it would be best done by inspecting the DOM as part of a unit/component test. If it's purely a styling change then I can see the argument for using a visual regression test (although those are less reliable and are slower and more costly to run).

kevinrodrigues (approved):

packages/components/pie-toggle-switch/test/component/pie-toggle-switch.spec.ts
raoufswe: i added the unit tests as requested :)

kevinrodrigues (approved):

PR #874, opened 2023-10-10
ci(pie-monorepo): DSW-1193 ensure all GH checks are passing

fernandofranca:

- Add Github action settings to ensure the required CI jobs are passing

A new check will appear under the name "Build / check-all-jobs".

image

In the [workflow view](https://github.com/justeattakeaway/pie/actions/runs/6507309654) you can see all the jobs "check-all-jobs" relies on:

image

In case some job fails, it will represented as usual:

image

Also, as an example the workflow of failed pipeline might look like this:

image

image

This is only an additional check and there's no need to change how we work.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -310,3 +310,29 @@ jobs:
       os: ubuntu-latest
       node-version: 18
     secrets: inherit
+
+  check-all-jobs:
+    if: always()
+
+    needs:
+    - lint-styles
+    - lint-js
+    - build-ubuntu-node-16
+    - build-windows-node-16
+    - build-ubuntu-node-18
+    - build-windows-node-18
+    - unit-tests
+    - browser-tests-components
+    - deploy-docs
+    - deploy-storybook
+    - browser-tests-docs
+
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -310,3 +310,29 @@ jobs:
...
+    needs:
+    - lint-styles
+    - lint-js
+    - build-ubuntu-node-16
+    - build-windows-node-16
+    - build-ubuntu-node-18
+    - build-windows-node-18
+    - unit-tests
+    - browser-tests-components
+    - deploy-docs
```
jamieomaguire, line 326: will this be ignored if the PR doesn't require a docs/storybook deployment?

fernandofranca: That's the idea Jamie. See line 337.

jamieomaguire: perfect 👍

jamieomaguire (approved):

siggerzz (approved):

ashleynolan (approved):

PR #872, opened 2023-10-10
feat(pie-webc-core): DSW-1306 extend validProperty decorator to support comma separated values

raoufswe:
## Describe your changes (can list changeset entries if preferable)

- I tried to wrap the validator so it gets used as an extension within the pie-card-container component but the actual value isn't passed to the decorator, instead, it's referenced as an instance field using the `this` keyword. 
- I extended the validator to accept comma-separated values, but I'm wondering if it's easier to just update our `paddingValues` array in `defs.ts` of the card to include every possible padding combination instead of modifying the decorator such as [here](https://stackoverflow.com/questions/43241174/javascript-generating-all-combinations-of-elements-in-a-single-array-in-pairs). this way we don't need to touch the decorator and be robust/clear about the valid values. 

## Author Checklist (complete before requesting a review)
- [ ] I have performed a self-review of my code
- [ ] If it is a core feature, I have added thorough tests
- [ ] If it is a `PIE Docs` change, I have reviewed the Docs site preview
- [ ] If it is a component change, I have reviewed the Storybook preview
- [ ] If there are visual test updates, I have reviewed them properly before approving

### Reviewer 1
- [ ] If it is a `PIE Docs` change, I have reviewed the PR preview
- [ ] If there are visual test updates, I have reviewed them

### Reviewer 2
- [ ] If it is a `PIE Docs` change, I have reviewed the PR preview
- [ ] If there are visual test updates, I have reviewed them

PR #875, opened 2023-10-10
feat(pie-card-container): DSW-1306 validate the padding prop

raoufswe:

- after raising [this PR ](https://github.com/justeattakeaway/pie/pull/872), I realized we could validate the padding prop in a simpler way than introducing a specific case to the validator. 

### Reviewer 1

### Reviewer 2


xander-marjoram (approved):

packages/components/pie-card-container/src/defs.ts
```diff
@@ -1,9 +1,14 @@
 export const variants = ['default', 'outline', 'inverse', 'outline-inverse'] as const;
 export const interactionTypes = ['anchor', 'button', 'none'] as const;
-export const padding = ['a', 'b', 'c', 'd', 'e', 'f', 'g'] as const;
-
-export type Variant = typeof variants[number];
-export type PaddingValue = typeof padding[number];
+export const paddingValues = [
```
kevinrodrigues, line 3: More of a suggestion so it might be easier to maintain in future (incase we add further values). You could define a set of the initial padding values i.e `'a-g'` in a defined const `paddingValues` then push the combination into it or just export the original like you have here but use a bit of functionality inside. Something like below:
Happy to keep it like you have it though so just a suggestion.

```javascript
export const paddingValues = (() => {
  const variants = ['a', 'b', 'c', 'd', 'e', 'f', 'g'];
  const result: (string)[] = [];

  variants.forEach((first) => {
    result.push(first);
    variants.forEach((second) => {
      result.push(`${first},${second}`);
    });
  });

  return result as const;
})();
```

kevinrodrigues (approved):

kevinrodrigues:
Nice one @raoufswe 🥇

packages/components/pie-card-container/src/defs.ts
raoufswe: good idea, i refactored to that but with a slightly different approach 🤝

kevinrodrigues (approved):

ashleynolan (commented):
Just one comment from me – looks great otherwise though!

xander-marjoram (approved):

apps/pie-storybook/stories/pie-card-container.stories.ts
```diff
@@ -63,9 +62,9 @@ const cardContainerStoryMeta: CardContainerStoryMeta = {
         padding: {
             description: 'Set the padding of the card.',
             control: {
-                type: 'text',
+                type: 'select',
```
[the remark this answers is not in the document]
raoufswe: updated to use a text input with a more detailed description

ashleynolan (approved):

kevinrodrigues (approved):

ashleynolan (approved):

PR #877, opened 2023-10-11
feat(pie-button): DSW-1272 form submits

jamieomaguire:
---
---

[Added] - Native form submit and reset handling logic to pie-button

---
"pie-storybook": minor
---

[Added] - Form integration test to pie-button

New form integration story:
Screenshot 2023-10-12 at 16 59 56

iOS Safari 14 verfied on IPhone 12 via Browserstack
Screenshot 2023-10-11 at 13 54 46


packages/components/pie-button/src/index.ts
```diff
@@ -43,6 +58,96 @@ export class PieButton extends LitElement implements ButtonProps {
...
+    }
+
+    private _handleClick () {
+        if (!this.isLoading && this.form) {
+            if (this.type === 'submit') {
+                if (!this.formNoValidate && this.form.reportValidity()) {
+                    this._simulateNativeButtonClick('submit');
+                } else {
+                    this._simulateNativeButtonClick('submit');
+                }
```
maledr5, line 142: Why do we need this check (if/else) if we call the same method?

jamieomaguire: Hey I will try to tidy this up.

If `formnovalidate` is set, we just submit. Otherwise we need to run a report Validity check first to trigger native form validation :) 

This code could be improved - will look now

packages/components/pie-button/README.md
```diff
@@ -146,6 +155,48 @@ For example, to add a click handler in various templates:
...
+The `pie-button` web component is designed to integrate with standard HTML forms just like a native HTML button. When positioned inside a form, the component will automatically associate itself, enabling it to directly interact with the form context.
+
+### Button Attributes
+
+The `pie-button` provides a set of attributes to customize its behavior within forms:
+
+- `type`: Determines the button's function. Set to `submit` for form submissions or `reset` to clear form fields.
+- `formaction`: Designates an alternative URL for form data submission when this specific button is clicked.
+- `formenctype`: Specifies the form data encoding type during submission via this button.
+- `formmethod``: Sets the HTTP method (e.g., GET or POST) for form data when initiated by this button.
```
xander-marjoram, line 169:
```suggestion
- `formmethod`: Sets the HTTP method (e.g., GET or POST) for form data when initiated by this button.
```

packages/components/pie-button/src/index.ts
```diff
@@ -43,6 +58,97 @@ export class PieButton extends LitElement implements ButtonProps {
...
     public isFullWidth = false;
 
+    @property({ type: String })
+    public name?: string;
+
+    @property({ type: String })
+    public value?: string;
+
+    @property({ attribute: 'formaction' })
+    public formAction: ButtonProps['formAction'];
```
xander-marjoram, line 68: Just curious if there was a reason why you went for camelCase prop names instead of aligning with the lowercase attribute? If I understand correctly we could either do what you've done, or go with the lowercase name and use `reflect: true`?

jamieomaguire: I thought this would be a bit more readable to reference in our code but I'm happy to reflect if you'd prefer? I am unsure if either approach has a performance benefit

packages/components/pie-button/src/index.ts
```diff
@@ -43,6 +58,95 @@ export class PieButton extends LitElement implements ButtonProps {
...
+        this.form.append(btn);
+        btn.click();
+        btn.remove();
+    }
+
+    private _handleClick () {
+        if (!this.isLoading && this.form) {
+            if (this.type === 'submit') {
+                // only submit the form if either formnovalidate is set, or the form passes validation checks (triggers native form validation)
+                if (this.formNoValidate || (!this.formNoValidate && this.form.reportValidity())) {
```
xander-marjoram, line 139: I think this is equivalent?
```suggestion
                if (this.formNoValidate || this.form.reportValidity()) {
```

jamieomaguire: lovely spot - pushing now

xander-marjoram (approved):

packages/components/pie-button/src/index.ts
```diff
@@ -43,6 +58,95 @@ export class PieButton extends LitElement implements ButtonProps {
...
+
+        // Visually hidden styles
+        btn.style.position = 'absolute';
+        btn.style.width = '1px';
+        btn.style.height = '1px';
+        btn.style.padding = '0';
+        btn.style.margin = '-1px';
+        btn.style.overflow = 'hidden';
+        btn.style.border = '0';
+        btn.style.whiteSpace = 'nowrap';
```
raoufswe, line 104: nit: do you think we should create a utility class somewhere in `pie-css` called `visually-hidden` and add to the element instead? 

something like: `btn.classList.add('visually-hidden');`

```
  position: absolute;
  width: 1px;
  height: 1px;
  padding: 0;
  margin: -1px;
  overflow: hidden;
  border: 0;
  white-space: nowrap;
```

also in case it helps, we have a mixin called `hidden-pseudo` in snacks that are used across multiple places that provides a similar thing.

jamieomaguire: @raoufswe I agree - I am going to add a helper class to `pie-css` in an immediate follow up PR :) just want to keep this one minimal 👍

packages/components/pie-button/src/index.ts
```diff
@@ -43,6 +58,95 @@ export class PieButton extends LitElement implements ButtonProps {
...
+            }
+            if (this.formmethod) {
+                btn.setAttribute('formmethod', this.formmethod);
+            }
+            if (this.formnovalidate) {
+                btn.setAttribute('formnovalidate', 'formnovalidate');
+            }
+            if (this.formtarget) {
+                btn.setAttribute('formtarget', this.formtarget);
+            }
```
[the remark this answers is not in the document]
raoufswe: yep sure 💯

maledr5 (approved):

packages/components/pie-button/README.md
```diff
@@ -96,15 +97,23 @@ yarn dev --filter=pie-storybook
...
-| isFullWidth | `Boolean` | `false`   | If `true`, sets the button width to 100% of its container.                                                       |
-| isLoading   | `Boolean` | `false`   | If `true`, displays a loading indicator inside the button.                                                        |
-| iconPlacement | `String`  | `leading`   | Icon placements of the icon slot, if provided, one of `iconPlacements` - `leading`, `trailing` |
+| Property       | Type      | Default    | Description                                                                                                          |
+|----------------|-----------|------------|----------------------------------------------------------------------------------------------------------------------|
+| size           | `String`  | `medium`   | Size of the button, one of `sizes` – `xsmall`, `small-expressive`, `small-productive`, `medium`, `large`             |
+| type           | `String`  | `submit`   | Type of the button, one of `types` – `submit`, `button`, `reset`. [Read further on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-type)|
```
raoufswe, line 103:
```suggestion
| type           | `String`  | `button`   | Type of the button, one of `types` – `submit`, `button`, `reset`. [Read further on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-type)|
```

raoufswe (dismissed):

maledr5 (approved):

siggerzz (dismissed):

apps/pie-storybook/stories/pie-button.stories.ts
```diff
@@ -40,7 +40,7 @@ const buttonStoryMeta: ButtonStoryMeta = {
             control: 'select',
             options: types,
             defaultValue: {
-                summary: 'submit',
+                summary: 'button',
```
ashleynolan, line 43: QQ – The default for this type in the actual component is `submit` as that matches what most browsers default to when rendering a button.

Should we keep the default here to be `submit` to match the component default?

jamieomaguire: the default has been changed in this PR to button. I can change it back in a follow up if you prefer

siggerzz (approved):

ashleynolan (approved):

packages/components/pie-button/src/index.ts
```diff
@@ -43,6 +58,95 @@ export class PieButton extends LitElement implements ButtonProps {
...
+                btn.setAttribute('formaction', this.formaction);
+            }
+            if (this.formenctype) {
+                btn.setAttribute('formenctype', this.formenctype);
+            }
+            if (this.formmethod) {
+                btn.setAttribute('formmethod', this.formmethod);
+            }
+            if (this.formnovalidate) {
+                btn.setAttribute('formnovalidate', 'formnovalidate');
```
xander-marjoram, line 123: Should this have been this instead?
```suggestion
                btn.setAttribute('formnovalidate', this.formnovalidate);
```

PR #910, opened 2023-10-20
docs(pie-css): DSW-000 fix readme navigation and improve structure

jamieomaguire:
- Fix the table of contents navigation in `pie-css` readme
- Slightly amend the readme copy to be a bit more structured for consumers

### DIFF
--- a/packages/tools/pie-css/README.md
+++ b/packages/tools/pie-css/README.md
@@ -12,15 +12,15 @@
 
 # Table of Contents
 
-1. [Introduction](#pie-css)
-2. [Installation](#installation)
+1. [Introduction](#introduction)
+2. [Using pie-css for building web applications](#using-pie-css-for-building-web-applications)
     1. JS or Framework import (via bundler)
     2. NuxtJS
     3. Sass /SCSS
     4. Native HTML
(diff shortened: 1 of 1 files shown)


raoufswe (approved):

maledr5 (approved):

packages/tools/pie-css/README.md
```diff
@@ -121,15 +122,17 @@ Of course, you could include the styles straight inside your HTML document – m
 
 
 
-## Using the PIE CSS – SCSS helpers
+## Using pie-css for building web components
+> [!NOTE]
+> The following section is only relevant if you are building web components as a part of our component library within the PIE monorepo.
```
ashleynolan, line 127: QQ – I don't think this is necessarily true. Other applications could choose to use these SCSS helpers if they wanted to in their own applications.

Chatting to teams that use the CSS variables for things like font-sizes for instance, they have also had to create functions/mixins to convert the unitless values to have px suffixes.

I'd probably update this to just indicate that these helpers are used in the PIE web components and can optionally be used by consuming applications.

ashleynolan: I'll pick this up and raise a PR for this update

PR #927, opened 2023-10-24
feat(pie-cookie-banner): DSW-1397 add hasPrimaryActionsOnly prop

xander-marjoram:

### Reviewer 1

### Reviewer 2

### DIFF
--- a/apps/pie-storybook/stories/pie-cookie-banner.stories.ts
+++ b/apps/pie-storybook/stories/pie-cookie-banner.stories.ts
@@ -47,8 +47,10 @@ const cookieBannerStoryMeta: CookieBannerStoryMeta = {
 
 export default cookieBannerStoryMeta;
 
-const necessaryOnlyAction = action('Necessary only');
-const acceptAllAction = action('Accept all');
+const necessaryOnlyAction = action('necessary-only');
+const acceptAllAction = action('accept-all');
+const managePrefsAction = action('manage-prefs');
+const prefsSavedAction = action('prefs-saved');
 
 const BaseStoryTemplate = (props: CookieBannerProps) : TemplateResult => {
     const { hasPrimaryActionsOnly } = props;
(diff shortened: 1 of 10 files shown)


raoufswe (approved):

xander-marjoram:
@fernandofranca Not sure really, the linter seems to pass either way...

apps/pie-storybook/stories/pie-cookie-banner.stories.ts
```diff
@@ -5,6 +5,7 @@ import { PieLink } from '@justeattakeaway/pie-link';
 import { PieModal } from '@justeattakeaway/pie-modal';
 import { PieIconButton } from '@justeattakeaway/pie-icon-button';
 import { PieToggleSwitch } from '@justeattakeaway/pie-toggle-switch';
+import { action } from '@storybook/addon-actions';
```
ashleynolan, line 8: QQ – in the storybook docs, it says that `addon-actions` is installed by default as part of the `addon-essentials` package. If that's the case, can e get away with not installing it separately as well?

packages/components/pie-cookie-banner/README.md
```diff
@@ -86,7 +86,7 @@ yarn dev --filter=pie-storybook
 
 | Property | Type | Default | Description |
 |---|---|---|---|
-| - | - | - | - |
+| hasPrimaryActionsOnly | `Boolean` | `false` | When true, sets the "Necessary only" button variant to "primary". |
```
ashleynolan, line 89: I'd probably avoid using copy in the description here in case it changes. Could be:

`When true, sets all action buttons in the cookie banner to display as variant "primary".`

apps/pie-storybook/stories/pie-cookie-banner.stories.ts
```diff
@@ -23,14 +24,17 @@ const keptReferences = [
...
+    hasPrimaryActionsOnly: false,
 };
 
 const cookieBannerStoryMeta: CookieBannerStoryMeta = {
     title: 'Cookie Banner',
     component: 'pie-cookie-banner',
     argTypes: {
-
+        hasPrimaryActionsOnly: {
+            description: 'When true, sets the "Necessary only" button variant to "primary".',
```
ashleynolan, line 35: Same as below:

`When true, sets all action buttons in the cookie banner to display as variant "primary".`

xander-marjoram: I actually think what I have is clearer, because "Manage preferences" is a button but that remains as an inverse button rather than changing to primary.

packages/components/pie-cookie-banner/README.md
ashleynolan: Yeh, I think we'll definitely need to add some docs somewhere about this when we start integrating.

Totally get that some events etc are fired as necessary only right now, just figured as this one could be uncoupled may be bette me to do that potentially.

apps/pie-storybook/stories/pie-cookie-banner.stories.ts
ashleynolan: Good point - I don't mind too much either way, just was wondering if it could be decoupled from the translation text

PR #935, opened 2023-10-27
fix(pie-button): DSW-1421 submit forms on enter press

jamieomaguire:
---
---

[Added] - Functionality to ensure pressing enter when in a form using a pie-button will trigger a form submit


packages/components/pie-button/src/index.ts
```diff
@@ -34,6 +34,18 @@ export class PieButton extends LitElement implements ButtonProps {
...
+
+        this.form?.addEventListener('keydown', this._handleFormKeyDown);
+    }
+
+    disconnectedCallback () {
+        super.disconnectedCallback();
+
+        this.form?.removeEventListener('keydown', this._handleFormKeyDown);
+    }
+
```
raoufswe, line 48: do you think we can pass `@keydown` to the button directly instead of this way? such as 

`<button @keydown={this. _handleFormKeyDown}`

jamieomaguire: I believe we want to capture `keydown` on anything inside the form. I.e. pressing enter when on a text input to submit the form. Would attaching the listener to the button enable this still?

raoufswe: sorry i didn't see this.form :,(

packages/components/pie-button/test/component/pie-button.spec.ts
```diff
@@ -266,6 +266,84 @@ test.describe('Form Actions', () => {
             expect(headers['content-type']).toMatch(/^multipart\/form-data;/);
             expect(method).toBe('POST');
         });
+
```
siggerzz, line 269: 🔥

siggerzz (approved):

fernandofranca (dismissed):

fernandofranca:
Out of curiosity, is the fact that our button is not exactly native the main reason that prevents the form from being submitted?

jamieomaguire:
> Out of curiosity, is the fact that our button is not exactly native the main reason that prevents the form from being submitted?

- Hey, yes it would seem that using web components as form controls has a number of issues. Generally speaking you need to add extra logic to get them to trigger native form behaviours. The element internals API is supposed to assist with setting up form association but it seems we still need to listen to events and such.

JoshuaNg2332:
> > Out of curiosity, is the fact that our button is not exactly native the main reason that prevents the form from being submitted?
> 
> Hey, yes it would seem that using web components as form controls has a number of issues. Generally speaking you need to add extra logic to get them to trigger native form behaviours. The element internals API is supposed to assist with setting up form association but it seems we still need to listen to events and such.

Just been debugging issues around this on your branch and found that `.focus()` doesn't work on the pie button component regardless of its' type. For that 1 failing test, you can get the correct behaviour of testing the button does not submit, via emulating tab keyboard presses, then emulating the enter keyboard press. Happy to chat about this issue whenever you are free.

JoshuaNg2332 (approved):

raoufswe (approved):

PR #949, opened 2023-10-31
fix(pie-monorepo): DSW-1434 add missing component imports

xander-marjoram:
- Add missing component imports
- Update READMEs with new instructions
- Add/update component imports in stories
- Update generator to include README and story changes
- Move pie component dependencies from `devDependencies` to `dependencies` for modal and cookie-banner
- Set `@justeattakeaway` packages as externals (exclude from bundle) for modal and cookie banner

### Reviewer 1

### Reviewer 2


xander-marjoram:
I'm not the biggest fan of having to disable eslint in lots of the files, but I imagine that most consumers of the components will only have to import the module to register the component.

xander-marjoram:
Not yet ready for review, I just want to check the visual tests still work

jamieomaguire (approved):

packages/components/pie-cookie-banner/src/index.ts
```diff
@@ -1,10 +1,20 @@
 import {
     LitElement, html, unsafeCSS, TemplateResult, nothing,
 } from 'lit';
-import { defineCustomElement } from '@justeattakeaway/pie-webc-core';
 import { property, state, queryAll } from 'lit/decorators.js';
 import { repeat } from 'lit/directives/repeat.js';
+
+import '@justeattakeaway/pie-button';
```
ashleynolan, line 7: Out of interest – what's the bundle size difference between the before/after of this change? (would be useful to know for when we get back onto the bundling in the next couple of weeks)

ashleynolan (approved):
🎉

jamieomaguire (approved):

packages/components/pie-modal/vite.config.js
```diff
@@ -1,3 +1,20 @@
...
+                index: 'src/index.ts',
+                react: 'src/react.ts',
+            },
+            formats: ['es'],
+        },
+        rollupOptions: {
+            external: [
+                'react',
+                /^lit/,
+                /@justeattakeaway/
```
ashleynolan, line 16: I'm guessing it'd just make sense to add this into the shared vite config potentially as a follow up (as all components should exclude these?)

xander-marjoram: Yes, that is the logical progression but I didn't want to do too much at once in this PR

fernandofranca:
With moving deps from devDependencies to dependencies aren't we risking to bundle code twice if a consumer imports both cookie banner and button in an app?

xander-marjoram:
@fernandofranca I tried it in the vanilla Aperture app and it actually behaves extremely nicely. Adding `@justeattakeaway/` as `external` for modal and cookie banner means the button (and others) won't be bundled inside them. We also don't see a duplicate registration error, presumably because the browser is clever enough to only load that bundle once! Looking at the Network tab seems to confirm this, as the button bundle is only downloaded once even though I import both components.

```javascript
import '@justeattakeaway/pie-button';
import '@justeattakeaway/pie-modal';
```

ashleynolan (approved):

packages/components/pie-cookie-banner/src/index.ts
xander-marjoram: | Component | Before | After |
| --- | --- | --- |
| `pie-modal` | 30.9kB | 22.7kB |
| `pie-cookie-banner` | 64.7kB |15.7kB |

It's also worth noting that before this PR, the cookie banner bundle also included the modal's dialog polyfill. Now it is only bundled as part of the modal.

fernandofranca (approved):

PR #958, opened 2023-11-02
feat(pie-spinner): DSW-1430 use pie-spinner in pie-button

raoufswe:
[Changed] - reuse pie-spinner in pie-button

### Reviewer 1

### Reviewer 2

### DIFF
--- a/packages/components/pie-button/src/index.ts
+++ b/packages/components/pie-button/src/index.ts
@@ -204,7 +204,7 @@ export class PieButton extends LitElement implements ButtonProps {
      */
     private renderSpinner (): TemplateResult {
         const spinnerSize = this.size.includes('small') ? 's' : 'm';
-        const inverseVariants = ['primary', 'destructive', 'outline-inverse', 'ghost-inverse'];
+        const inverseVariants: Array<ButtonProps['variant']> = ['primary', 'destructive', 'outline-inverse', 'ghost-inverse'];
         const spinnerVariant = inverseVariants.includes(this.variant) ? 'inverse' : 'secondary';
 
         return html`
--- a/packages/components/pie-button/src/index.ts
+++ b/packages/components/pie-button/src/index.ts
@@ -1,5 +1,5 @@
 import {
-    LitElement, html, unsafeCSS, nothing, PropertyValues,
+    LitElement, html, unsafeCSS, nothing, PropertyValues, TemplateResult,
 } from 'lit';
 import { property } from 'lit/decorators.js';
 import { validPropertyValues, defineCustomElement } from '@justeattakeaway/pie-webc-core';
@@ -8,6 +8,7 @@ import {
 } from './defs';
 import styles from './button.scss?inline';
 import 'element-internals-polyfill';
+import '@justeattakeaway/pie-spinner';
 
 // Valid values available to consumers
 export * from './defs';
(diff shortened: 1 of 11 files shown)


packages/components/pie-button/src/index.ts
```diff
@@ -196,6 +197,23 @@ export class PieButton extends LitElement implements ButtonProps {
         this._handleClick();
     };
 
+    /**
+     * Template for the loading state
+     *
+     * @private
+     */
+    private renderSpinner (): TemplateResult {
+        const spinnerSize = this.size.includes('small') ? 's' : 'm';
```
fernandofranca, line 206: Why not `===`?

packages/components/pie-button/src/index.ts
```diff
@@ -196,6 +197,23 @@ export class PieButton extends LitElement implements ButtonProps {
...
+        const inverseVariants = ['primary', 'destructive', 'outline-inverse', 'ghost-inverse'];
```
fernandofranca, line 207: Shouldn't we rely on references instead of strings for this? Once a variant name changes (hopefully never 😆) this will become outdated.

packages/components/pie-button/src/index.ts
raoufswe: the button sizes can be `'xsmall', 'small-productive', 'small-expressive',` and based on the design specs, we have to use the `s` size of the spinner for those button sizes. 

so instead of having something like -> 

```
this.size === 'xsmall' || this.size === "small-productive" |||  this.size === 'small-expressive' ? 's' : 'm';
```

i use the `inlcude` syntax as it should cover all cases.

packages/components/pie-button/src/index.ts
raoufswe: I might be missing your point, but these `strings` are a design contract only applicable to the spinner inside a button, if the button variant is one of the above, we need to use the inverse variant of the spinner and if ever needs updating, the logic is all located in the same code block. could you please elaborate more?

packages/components/pie-button/src/index.ts
fernandofranca: That's fair! I'd add a comment to remind others of this implicit behaviour.

packages/components/pie-button/src/index.ts
fernandofranca: My point is that strings are weak as references in the sense of which variants exist. 

`'primary', 'destructive', 'outline-inverse', 'ghost-inverse'` are a subset of the `variants` set in `defs.ts` correct?

The fact that the logic is in that block doesn't prevent us from making a mistake. If one of those variant names changes, let's say `outline-inverse`, since the `renderSpinner` has no connection with the `variants` from `defs.ts` it will not match the new name.

raoufswe: could you please share what you think is the best approach here instead? is moving `inverseVariants` to `defs `would be enough?

fernandofranca: I'm not suggesting to use enums, but anything that relies on references would be good.

raoufswe: @fernandofranca do you think that would fix the issue?

raoufswe: i think the typing trick is neat and will ensure we don't use invalid values. 
updated 💯

packages/components/pie-button/src/index.ts
```diff
@@ -196,6 +197,23 @@ export class PieButton extends LitElement implements ButtonProps {
         this._handleClick();
     };
 
+    /**
+     * Template for the loading state
+     *
+     * @private
+     */
+    private renderSpinner (): TemplateResult {
```
kevinrodrigues, line 205: Do we need tests?

packages/components/pie-button/src/index.ts
```diff
@@ -217,6 +235,7 @@ export class PieButton extends LitElement implements ButtonProps {
                 ?disabled=${disabled}
                 ?isFullWidth=${isFullWidth}
                 ?isLoading=${isLoading}>
+                    ${isLoading ? this.renderSpinner() : nothing}
```
kevinrodrigues, line 238: Similar to above.

packages/components/pie-button/src/index.ts
raoufswe: Similar to the other comment, we already have visual and component tests in place for this, this PR just refactored the implementation of the logic.

xander-marjoram:
Can you add `/@justeattakeaway/` as an external in the vite config for button so that pie-spinner isn't bundled?

xander-marjoram (approved):

kevinrodrigues (approved):

PR #986, opened 2023-11-09
feat(pie-button): DSW-1293 implement responsive mode

fernandofranca:

[Added] - `isResponsive` and `responsiveSize` prop

### Reviewer 1

### Reviewer 2

### DIFF
--- a/apps/pie-storybook/stories/pie-button.stories.tsx
+++ b/apps/pie-storybook/stories/pie-button.stories.tsx
@@ -80,7 +80,7 @@ const buttonStoryMeta: ButtonStoryMeta = {
             },
         },
         isResponsive: {
-            description: 'If `true`, enables responsive behaviour.',
+            description: 'If `true`, uses the next larger size on wide viewports',
             control: 'boolean',
             defaultValue: {
                 summary: false,
--- a/apps/pie-storybook/stories/pie-button.stories.tsx
+++ b/apps/pie-storybook/stories/pie-button.stories.tsx
@@ -3,7 +3,7 @@ import { html, nothing } from 'lit';
 /* eslint-disable import/no-duplicates */
 import '@justeattakeaway/pie-button';
 import {
-    ButtonProps as ButtonPropsBase, iconPlacements, sizes, types, variants,
+    ButtonProps as ButtonPropsBase, iconPlacements, sizes, types, variants, responsiveSizes,
 } from '@justeattakeaway/pie-button';
 /* eslint-enable import/no-duplicates */
 import '@justeattakeaway/pie-icons-webc/IconPlusCircle';
@@ -21,6 +21,7 @@ const defaultArgs: ButtonProps = {
     disabled: false,
     isFullWidth: false,
     isLoading: false,
+    isResponsive: false,
     slot: 'Label',
 };
 
@@ -73,6 +74,13 @@ const buttonStoryMeta: ButtonStoryMeta = {
         },
         isLoading: {
             description: 'If `true`, displays a loading indicator inside the button.',
+            control: 'boolean',
+            defaultValue: {
+                summary: false,
+            },
+        },
+        isResponsive: {
+            description: 'If `true`, enables responsive behaviour.',
             control: 'boolean',
             defaultValue: {
                 summary: false,
@@ -141,6 +149,14 @@ const buttonStoryMeta: ButtonStoryMeta = {
             },
             if: { arg: 'type', eq: 'submit' },
         },
+        responsiveSize: {
+            description: 'Set the size of the button when set as responsive for wider viewports.',
+            control: 'select',
+            options: ['', ...responsiveSizes],
+            defaultValue: {
+                summary: 'productive',
+            },
+        },
     },
     args: defaultArgs,
     parameters: {
@@ -160,6 +176,7 @@ const Template: TemplateFunction<ButtonProps> = ({
     disabled,
     isFullWidth,
     isLoading,
+    isResponsive,
     slot,
     iconPlacement,
     name,
@@ -169,6 +186,7 @@ const Template: TemplateFunction<ButtonProps> = ({
     formmethod,
     formnovalidate,
     formtarget,
+    responsiveSize,
 }) => html`
 <pie-button
     size="${size}"
@@ -178,8 +196,10 @@ const Template: TemplateFunction<ButtonProps> = ({
     ?disabled="${disabled}"
     ?isLoading="${isLoading}"
     ?isFullWidth="${isFullWidth}"
+    ?isResponsive="${isResponsive}"
     name=${name || nothing}
     value=${value || nothing}
+    responsiveSize="${responsiveSize || nothing}"
     formaction=${formaction || nothing}
     formenctype=${formenctype || nothing}
     formmethod=${formmethod || nothing}
--- a/packages/components/pie-button/README.md
+++ b/packages/components/pie-button/README.md
@@ -95,7 +95,7 @@ import { PieButton } from '@justeattakeaway/pie-button/dist/react';
 | formmethod     | `String`  | `undefined`| Sets the HTTP method for form data. [Read further on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-formmethod) |
 | formnovalidate | `Boolean` | `undefined`    | Ensures the form is submitted without native HTML validation. [Read further on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-formnovalidate) |
 | formtarget     | `String`  | `undefined`| Dictates where to display the response after form submission. [Read further on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-formtarget) |
-| responsiveSize | `String`  | `undefined`| Sets a specific size for xsmall button when rendered as responsive. It can be `productive` or `expressive`. When no prop is attributed it defaults to `productive` settings |
+| responsiveSize | `String`  | `productive`| Sets a specific size for xsmall button when rendered as responsive. It can be `productive` or `expressive`.         |
 
 
 In your markup or JSX, you can then use these to set the properties for the `pie-button` component:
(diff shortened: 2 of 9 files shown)


packages/components/pie-button/README.md
```diff
@@ -94,6 +95,7 @@ import { PieButton } from '@justeattakeaway/pie-button/dist/react';
 | formmethod     | `String`  | `undefined`| Sets the HTTP method for form data. [Read further on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-formmethod) |
 | formnovalidate | `Boolean` | `undefined`    | Ensures the form is submitted without native HTML validation. [Read further on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-formnovalidate) |
 | formtarget     | `String`  | `undefined`| Dictates where to display the response after form submission. [Read further on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-formtarget) |
+| responsiveSize | `String`  | `undefined`| Sets a specific size for xsmall button when rendered as responsive. It can be `productive` or `expressive`. When no prop is attributed it defaults to `productive` settings |
```
xander-marjoram, line 98: I wonder if it might be slightly clearer to have `productive` as the default value rather than `undefined`, what do you think?

packages/components/pie-button/test/component/pie-button.spec.ts
```diff
@@ -1,11 +1,56 @@
...
+import type { Locator } from '@playwright/test';
 import { PieButton, ButtonProps } from '@/index';
 
 const props: Partial<ButtonProps> = {
     size: 'large',
     variant: 'primary',
 };
 
+type SizeResponsiveSize = {
+    sizeName: 'xsmall' | 'small-productive' | 'small-expressive' | 'medium' | 'large';
```
xander-marjoram, line 11: I think you could do something like `Omit<ButtonSize, 'xlarge'>`?

xander-marjoram: Ah actually it's `Exclude` not `Omit`

packages/components/pie-button/test/component/pie-button.spec.ts
```diff
@@ -525,3 +570,110 @@ test.describe('Form Actions', () => {
...
+                    {
+                        props,
+                        slots: {
+                            default: 'Click me!',
+                        },
+                    },
+                );
+
+                await expect(component.locator('button'))
+                    .not.toHaveAttribute('isResponsive', '');
```
xander-marjoram, line 589: I suspect this test will pass if the attribute has a different value, as it is only checking that the attribute is not `''`. I think you could just have `.not.toHaveAttribute('isResponsive');` and that would be fine

packages/components/pie-button/test/component/pie-button.spec.ts
```diff
@@ -525,3 +570,110 @@ test.describe('Form Actions', () => {
...
+                    {
+                        props: testProps,
+                        slots: {
+                            default: 'Click me!',
+                        },
+                    },
+                );
+
+                await expect(component.locator('button'))
+                    .toHaveAttribute('isResponsive', '');
```
xander-marjoram, line 611: I think this should also work without the second parameter

packages/components/pie-button/test/component/pie-button.spec.ts
fernandofranca: I also was expecting it to work like that, but will only be available once we upgrade Playwright.

image

packages/components/pie-button/test/component/pie-button.spec.ts
fernandofranca: Ditto

packages/components/pie-button/test/component/pie-button.spec.ts
fernandofranca: I suspect we don't have that.

packages/components/pie-button/test/component/pie-button.spec.ts
fernandofranca: I have added comments to remind us to clean it up after the upgrade.

apps/pie-storybook/stories/pie-button.stories.tsx
```diff
@@ -78,6 +79,13 @@ const buttonStoryMeta: ButtonStoryMeta = {
                 summary: false,
             },
         },
+        isResponsive: {
+            description: 'If `true`, enables responsive behaviour.',
```
ashleynolan, line 83: I'd use the same description as you have in the `README` here:

```suggestion
            description: 'If `true`, uses the next larger size on wide viewports',
```

ashleynolan (commented):
One super small comment, otherwise looks great 👍

packages/components/pie-button/test/component/pie-button.spec.ts
```diff
@@ -1,11 +1,56 @@
...
+ * @returns The values of the given style properties
+ */
+async function getShadowElementStylePropValues (element:Locator, selector:string, props:Array<string>):Promise<Array<string>> {
+    const data = { selector, props };
+
+    const evaluated = await element.evaluate((el, data) => {
+        const { selector, props } = data;
+
+        if (!el || !el.shadowRoot) {
+            throw new Error('getShadowElementStylePropValues: evaluate didnt returned an element');
```
xander-marjoram, line 29:
```suggestion
            throw new Error('getShadowElementStylePropValues: evaluate didn't return an element');
```

packages/components/pie-button/test/component/pie-button.spec.ts
```diff
@@ -1,11 +1,56 @@
...
+        }
+
+        const shadowEl = el.shadowRoot.querySelector(selector);
+
+        if (!shadowEl) {
+            throw new Error('getShadowElementStylePropValues: no shadow element were found');
```
xander-marjoram, line 35:
```suggestion
            throw new Error('getShadowElementStylePropValues: no shadow element was found');
```

xander-marjoram (dismissed):

ashleynolan (dismissed):

packages/components/pie-button/test/component/pie-button.spec.ts
siggerzz: No problems from my side! 🙌🏼

xander-marjoram (dismissed):

ashleynolan (approved):

packages/components/pie-button/test/component/pie-button.spec.ts
```diff
@@ -525,3 +570,110 @@ test.describe('Form Actions', () => {
...
+            });
+        });
+
+        test.describe('when set to a valid value', () => {
+            test('the button should have the attribute', async ({ mount }) => {
+                const testProps:Partial<ButtonProps> = {
+                    ...props,
+                    size: 'xsmall',
+                    isResponsive: true,
+                    responsiveSize: 'expressive',
```
kevinrodrigues, line 661: Do we need a test for `productive` too?

kevinrodrigues (dismissed):

fernandofranca:
Extracted the new visual tests for responsive mode into a [new snapshot](https://percy.io/4bc223d1/pie-button/builds/31179049/changed/1718366757?browser=edge&browser_ids=47%2C48%2C49%2C50&subcategories=approved&viewLayout=side-by-side&viewMode=new&width=375&widths=375).

The addition of the new props led to a huge increase in tests, as well as the generated snapshot was extrapolating Percy's size limits as @siggerzz noted.

siggerzz (approved):
Awesome work @fernandofranca! 🚀

apps/pie-storybook/stories/pie-button.stories.tsx
```diff
@@ -141,6 +149,14 @@ const buttonStoryMeta: ButtonStoryMeta = {
             },
             if: { arg: 'type', eq: 'submit' },
         },
+        responsiveSize: {
+            description: 'Set the size of the button when set as responsive for wider viewports.',
+            control: 'select',
+            options: ['', ...responsiveSizes],
+            defaultValue: {
+                summary: 'productive',
+            },
```
raoufswe, line 158: nit: can be addressed in a follow-up PR.

I think we can use an if statement to only render this if isResponsive is true and avoid adding an empty string to the options.

```suggestion
            if: { arg: 'isResponsive', eq: true },

```

packages/components/pie-button/README.md
```diff
@@ -86,6 +86,7 @@ import { PieButton } from '@justeattakeaway/pie-button/dist/react';
 | disabled       | `Boolean` | `false`    | If `true`, disables the button.                                                                                      |
 | isFullWidth    | `Boolean` | `false`    | If `true`, sets the button width to 100% of its container.                                                          |
 | isLoading      | `Boolean` | `false`    | If `true`, displays a loading indicator inside the button.                                                           |
+| isResponsive   | `Boolean` | `false`    | If `true`, uses the next larger size on wide viewports.                                                              |
```
raoufswe, line 89: I know the button docs PR went in before this PR, but we need to update the `props.json` file in the docs to reflect this new prop until we entirely remove the props table from the readme

packages/components/pie-button/README.md
```diff
@@ -94,6 +95,7 @@ import { PieButton } from '@justeattakeaway/pie-button/dist/react';
 | formmethod     | `String`  | `undefined`| Sets the HTTP method for form data. [Read further on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-formmethod) |
 | formnovalidate | `Boolean` | `undefined`    | Ensures the form is submitted without native HTML validation. [Read further on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-formnovalidate) |
 | formtarget     | `String`  | `undefined`| Dictates where to display the response after form submission. [Read further on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-formtarget) |
+| responsiveSize | `String`  | `productive`| Sets a specific size for xsmall button when rendered as responsive. It can be `productive` or `expressive`.         |
```
raoufswe, line 98: same here, this needs to be added to the props.json file in the button docs

raoufswe (approved):
🥇 
my comments are non blocking and can be addressed in a follow up PR

PR #1056, opened 2023-11-24
feat(pie-icons-webc): DSW-1281 react entry point

jamieomaguire:
---
---

[Added] - React entry point to the library that allows consumers to import individual or all icons either from /react/icon path or entire /react folder


.changeset/short-colts-buy.md
```diff
@@ -0,0 +1,5 @@
+---
+"@justeattakeaway/pie-icons-webc": minor
+---
+
+[Added] - React entry point to the library that allows consumers to import individual or all icons either from /react/icon path or entire /react folder
```
ashleynolan, line 5: Could we add this to the package `README` so it's documented?

ashleynolan (commented):
Looks good – would just be good to make sure the README is up-to-date with current expected usage.

packages/tools/pie-icons-webc/tests/icons-react.spec.js
```diff
@@ -0,0 +1,14 @@
+import { test, expect } from '@playwright/experimental-ct-react';
+import { IconAlcohol } from '../icons/react/IconAlcohol';
+
+test.use({ viewport: { width: 500, height: 500 } });
+
+test('should render', async ({ page, mount }) => {
+    await mount(<IconAlcohol />);
+    expect(await page.screenshot()).toMatchSnapshot('IconAlcohol.png');
```
jamieomaguire, line 8: myself and @siggerzz agreed for now that we will add two snapshot tests that verify the component renders with and without props. We will create a larger ticket to develop a more robust suite of tests 👍

PR #1067, opened 2023-11-28
ci(pie-monorepo): DSW-000 remove only deployed applications after merge & associated GH Environment

siggerzz:
This PR fixes an issue where Amplify Teardown action would fail in instances where only one application was deployed.

New logic has been added to:
- Check which applications were deployed, and delete them
- Remove the associated Github Environment


.github/workflows/closed.yml
```diff
@@ -11,6 +11,45 @@ env:
   STORYBOOK_AMPLIFY_ID: d17ja0ul7nrdy0
 
 jobs:
+
+  check-change-type:
```
siggerzz, line 15: This job has been copied from `ci.yml`

ashleynolan: Is there a way of pulling it out into a shared file (if so, can create a ticket to tidy this up?)

ashleynolan (approved):

.github/workflows/closed.yml
siggerzz: Yeah, we should be able to make use of reusable actions. I'll get a ticket created for this, as there's probably a few places we can make use of this!

jamieomaguire (approved):

PR #1101, opened 2023-12-12
feat(pie-wrapper-react): DSW-1314 All component props are marked as optional

fernandofranca:

This PR updates:
- Updates the React wrapper script to use [the workaround](https://github.com/lit/lit/issues/4219#issuecomment-1740033448) for the bug that sets all props as optional in React
- Set non-required props as optional
- Added safeguards for optional props
- Added temporary `ReactBaseType`

After updating the components React wrapper with the solution suggested by the Lit team, when testing the components in React it became clear that we were making most of the props required, and that can be rather inconvenient for consumers.

image

IMO the best approach is to set accurately what should be required and what is optional from consumers' perspective, and runtime safeguards should be implemented where necessary since that's not a TypeScript responsibility.

This PR doesn't include all the necessary changes to make our types completely accurate, and DSW-1577 was created to tackle them.

### Reviewer 1

### Reviewer 2

### DIFF
--- a/.changeset/dry-beds-love.md
+++ b/.changeset/dry-beds-love.md
@@ -2,4 +2,4 @@
 "@justeattakeaway/pie-wrapper-react": minor
 ---
 
-[Changed] React Wrapper script
+[Changed] - React Wrapper script adding the workaround for the bug that sets all props as optional in React and TS
--- a/.changeset/spicy-guests-add.md
+++ b/.changeset/spicy-guests-add.md
@@ -12,4 +12,4 @@
 "@justeattakeaway/pie-link": patch
 ---
 
-[Changed] React Wrapper script
+[Changed] - Use the updated React Wrapper script


kevinrodrigues (approved):

packages/tools/pie-wrapper-react/scripts/add-react-wrapper.js
```diff
@@ -102,21 +102,27 @@ export function addReactWrapper (customElementsObject, folderName = process.argv
...
-    elementClass: ${component.class.name}React,
+    elementClass: ${component.class.name}Lit,
     react: React,
     tagName: '${component.class.tagName}',
     events: ${eventsObject},
 });
+
+type ReactBaseType = Pick<React.ButtonHTMLAttributes<HTMLButtonElement>, 'children' | 'onClick'>
+
+export const ${component.class.name} = ${component.class.name}React as React.ForwardRefExoticComponent<React.PropsWithoutRef<${componentPropsExportName}> & React.RefAttributes<${component.class.name}Lit> & ReactBaseType>;
```
kevinrodrigues, line 125: Do we have a ticket in the backlog to relook at this workaround when Lit fix their bug? Might be worth making on and referencing the issue we raised?

fernandofranca: I believe it's a good idea.

LTurns (approved):

packages/tools/pie-wrapper-react/scripts/add-react-wrapper.js
```diff
@@ -102,21 +102,27 @@ export function addReactWrapper (customElementsObject, folderName = process.argv
...
+            componentSrc = `import * as React from 'react';
 import { createComponent${component.class.events?.length > 0 ? ', EventName' : ''} } from '@lit/react';
-import { ${component.class.name} as ${component.class.name}React } from './index';
+import { ${component.class.name} as ${component.class.name}Lit } from './index';
+import { ${componentPropsExportName} } from './defs';
 
 export * from './defs';
 
-export const ${component.class.name} = createComponent({
+const ${component.class.name}React = createComponent({
```
LTurns, line 115: LGTM! small comment - you may need to update the example in the README so it matches this format 😄

LTurns (dismissed):

packages/components/pie-button/src/defs.ts
```diff
@@ -17,36 +17,36 @@ export interface ButtonProps {
     /**
      * What size the button should be.
      */
-    size: typeof sizes[number];
+    size?: typeof sizes[number];
```
ashleynolan, line 20: Will we need to make similar changes across all other components?

ashleynolan (commented):
This change looks good to me – I would personally run it past one of the Skip engineers (such as @davidpn11 or @thejfreitas), mainly as I consider them to have extremely good opinions on TS + React usage.

packages/components/pie-button/src/defs.ts
fernandofranca: Yes, and a ticket for that is ready for refinement (DSW-1577)

packages/components/pie-button/src/index.ts
```diff
@@ -244,9 +244,9 @@ export class PieButton extends LitElement implements ButtonProps {
             <button
                 @click=${this._handleClick}
                 class="o-btn"
-                type=${type}
-                variant=${variant}
-                size=${size}
+                type=${ifDefined(type)}
```
xander-marjoram, line 247: Is there a way we can tell typescript that there will definitely be a value at this point? If there is no value by the time it gets to the template then something has gone wrong

packages/tools/pie-wrapper-react/README.md
```diff
@@ -9,20 +9,27 @@ With this package, the below code (example of `pie-button`) is automatically gen
...
+This package references a `custom-elements.json` file, which is generated via a `yarn cem analyze` command from the npm package `@custom-elements-manifest/analyzer`. This package searches through the repo for any custom elements and condenses its information, such as events and attributes, into a large JSON object - making it easier to use for wrappers. Therefore, any changes made to a web component will automatically be reflected in the `custom-elements.json` file during the build process.
 
-If you notice changes in this file, please commit this as part of your PR. These are necessary for the react-wrapper to be generated with the most up to date version of the component so should be committed along with your other file changes.
+If you notice changes in this file, please commit this as part of your PR. These are necessary for the react-wrapper to be generated with the most up-to-date version of the component so should be committed along with your other file changes.
```
xander-marjoram, line 32: ⭐

packages/tools/pie-wrapper-react/scripts/add-react-wrapper.js
xander-marjoram, line 125: I'm curious to know how something like this would behave, rather than setting them all to optional inside defs.ts?
```suggestion
export const ${component.class.name} = ${component.class.name}React as React.ForwardRefExoticComponent<React.PropsWithoutRef<Partial<${componentPropsExportName}>> & React.RefAttributes<${component.class.name}Lit> & ReactBaseType>;
```

PR #1103, opened 2023-12-12
feat(pie-switch): DSW-1381 forms integration

jamieomaguire:
---
"@justeattakeaway/pie-switch": minor
---

[Added] - Form submission and validation logic

---
"pie-docs": minor
---

[Added] - Form example usage for pie-switch component to code page

---
"@justeattakeaway/pie-cookie-banner": minor
---

[Changed] - Replace isChecked property references to pie-switch with checked

---
"pie-storybook": minor
---

[Added] - New form integration story for pie-switch


apps/pie-storybook/stories/pie-switch.stories.ts
```diff
@@ -82,23 +110,89 @@ const changeAction = (event: Event) => action('change')({
...
+        // Display a success message to the user when they submit the form
+        const form = document.querySelector('#testForm');
+        const formLog = document.querySelector('#formLog');
+
+        const el = document.querySelector('pie-switch');
+        const s = document.querySelector('pie-switch');
+        setTimeout(() => {
+        console.log(s);
+        // console.log('checking the report v value');
+        // s.setCustomValidity('wang it in, son');
```
xander-marjoram, line 160: 👀

apps/pie-storybook/stories/pie-switch.stories.ts
```diff
@@ -82,23 +110,89 @@ const changeAction = (event: Event) => action('change')({
...
+
+            for (const entry of formData.entries()) {
+                console.table(entry);
+            }
+
+            // Reset the success message after roughly 8 seconds
+            setTimeout(() => {
+                formLog.innerHTML = '';
+                formLog.style.display = 'none';
+            }, 8000);
```
xander-marjoram, line 188: Not important at all but I noticed if you submit the form twice within those 8 seconds the reset from the first click will apply to the second

apps/pie-docs/src/components/switch/code/props.json
```diff
@@ -2,7 +2,7 @@
     "headings": ["Prop", "Options", "Description", "Default"],
     "rows": [
         [
-            "isChecked",
+            "checked",
```
ashleynolan, line 5: I don't have an issue with changing this to `checked`, but if we're going to do that, likely need to also change `isDisabled` at the same time so that all HTML attribute style props are treated consistently

packages/components/pie-switch/src/index.ts
```diff
@@ -31,37 +61,103 @@ export class PieSwitch extends RtlMixin(LitElement) implements SwitchProps {
...
 
     static styles = unsafeCSS(styles);
 
+    /**
+     * Ensures that the form value and validation state are in sync with the component.
+     */
+    private handleFormAssociation () : void {
+        const isFormAssociated = !!this._internals.form && !!this.name && !!this.value;
+        if (isFormAssociated) {
+            // If checked and disabled, should the value still be submitted?
```
xander-marjoram, line 86: This comment can be removed now that we worked out the answer is no 😄

packages/components/pie-switch/src/switch.scss
```diff
@@ -85,8 +85,17 @@
 }
 
 .c-switch-input {
-    appearance: none;
-    margin: 0;
+    position: absolute;
```
xander-marjoram, line 88: Did we check this still looks okay in RTL?

apps/pie-storybook/stories/pie-switch.stories.ts
jamieomaguire: this is debug code I didn't mean to commit!

apps/pie-storybook/stories/pie-switch.stories.ts
jamieomaguire: Oh no

apps/pie-storybook/stories/pie-switch.stories.ts
jamieomaguire: regarding the double submit, I don't think it matters for the sake of this story. It's just minimal code to add some extra visibility of integration

packages/components/pie-switch/test/component/pie-switch.spec.ts
```diff
@@ -278,4 +278,107 @@ test.describe('Component: `Pie switch`', () => {
...
+                    e.preventDefault(); // Prevent the actual submission
+
+                    // Serialize form data
+                    const formData = new FormData(form);
+                    const formDataObj = {};
+
+                    // This is JS - we don't need TS inside these calls
+                    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+                    // @ts-ignore
+                    formData.forEach((value, key) => { formDataObj[key] = value; });
```
xander-marjoram, line 312: This probably is correct but it caught my eye and I thought I'd ask, are value and key the right way round?
```suggestion
                    formData.forEach((value, key) => { formDataObj[key] = value; });
```

apps/pie-docs/src/components/switch/code/props.json
jamieomaguire: we use `disabled` everywhere else, this is the component that breaks the pattern. Xander and I have agreed to use `disabled` instead but not as part of this work. We can create a new ticket for it :)

packages/components/pie-switch/test/component/pie-switch.spec.ts
```diff
@@ -278,4 +278,107 @@ test.describe('Component: `Pie switch`', () => {
...
+            // Assert
+            const formDataJson = await page.$eval('#formDataJson', (el) => el.textContent);
+            const formDataObj = JSON.parse(formDataJson || '{}');
+
+            // Check if the switch value is in the form data
+            expect(switchName in formDataObj).toBe(true);
+            expect(formDataObj[switchName]).toBe(switchValue);
+        });
+
+        test('form should be invalid and not submit if the switch is required but not set', async ({ page }) => {
```
xander-marjoram, line 338: These tests look good! Another case I can think of:

When disabled, the value is not included in form submission (both checked and unchecked)

apps/pie-docs/src/components/switch/code/props.json
xander-marjoram: DSW-1571 😄

packages/components/pie-switch/src/switch.scss
jamieomaguire: Nope - will do now

jamieomaguire: Seems fine locally testing but I believe Percy will also catch any regression here

packages/components/pie-switch/test/component/pie-switch.spec.ts
jamieomaguire: It does appear to be correct, strangely!

xander-marjoram (approved):

ashleynolan (approved):

PR #1112, opened 2023-12-14
fix(pie-monorepo): DSW-000 fixed stylelint-config-pie folder name

ashleynolan:
## Describe your changes

- [Fixed] - stylelint-config-pie folder name


.github/project-labeler.yml
```diff
@@ -30,7 +30,7 @@ eslint-config-pie:
 - 'packages/tools/eslint-config-pie/**/*'
 
 stylelint-config-pie:
-- 'packages/tools/pie-stylelint-config/**/*'
+- 'packages/tools/stylelint-config-pie/**/*'
```
jamieomaguire, line 33: lgtm, do we need to update any references in the monorepo?

sebastienrocks (approved):

.github/project-labeler.yml
ashleynolan: Nope – I did a find and replace and because the package name itself was changed before, all those links have been updated previously through that link. It's only this one left from what I could see

jamieomaguire (approved):

.github/project-labeler.yml
jamieomaguire: Nice!

PR #1117, opened 2023-12-15
DNM Dsw 0 form label switch

jamieomaguire:
DO NOT MERGE - this is a demo of how to get `pie-form-label` component working with any input components we create, such as `pie-switch`

### DIFF
--- a/apps/pie-storybook/stories/pie-switch.stories.ts
+++ b/apps/pie-storybook/stories/pie-switch.stories.ts
@@ -119,10 +119,11 @@ const Template : TemplateFunction<SwitchProps> = (props) => {
         required,
     } = props;
 
+    // can put the label adjacent to the switch OR put directly in the switch component(check the switch index.ts file to see example)
+    // <pie-form-label for="pie-switch">
+    //     label slot content
+    // </pie-form-label>
     return html`
-        <pie-form-label for="pie-switch">
-            label slot content
-        </pie-form-label>
         <pie-switch
             id="pie-switch"
             name="${name || nothing}"
--- a/apps/pie-storybook/stories/pie-switch.stories.ts
+++ b/apps/pie-storybook/stories/pie-switch.stories.ts
@@ -1,6 +1,6 @@
 import { html, nothing } from 'lit';
 import { action } from '@storybook/addon-actions';
-
+import '@justeattakeaway/pie-form-label';
 /* eslint-disable import/no-duplicates */
 import '@justeattakeaway/pie-switch';
 import { SwitchProps, labelPlacements } from '@justeattakeaway/pie-switch';
@@ -120,6 +120,9 @@ const Template : TemplateFunction<SwitchProps> = (props) => {
     } = props;
 
     return html`
+        <pie-form-label for="pie-switch">
+            label slot content
+        </pie-form-label>
         <pie-switch
             id="pie-switch"
             name="${name || nothing}"
--- a/apps/pie-storybook/stories/pie-switch.stories.ts
+++ b/apps/pie-storybook/stories/pie-switch.stories.ts
@@ -119,11 +119,10 @@ const Template : TemplateFunction<SwitchProps> = (props) => {
         required,
     } = props;
 
-    // can put the label adjacent to the switch OR put directly in the switch component(check the switch index.ts file to see example)
-    // <pie-form-label for="pie-switch">
-    //     label slot content
-    // </pie-form-label>
     return html`
+        <pie-form-label for="pie-switch">
+            label slot content
+        </pie-form-label>
         <pie-switch
             id="pie-switch"
             name="${name || nothing}"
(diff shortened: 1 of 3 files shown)


packages/components/pie-form-label/src/index.ts
```diff
@@ -6,6 +6,11 @@ import { property } from 'lit/decorators.js';
 import styles from './form-label.scss?inline';
 import { FormLabelProps } from './defs';
 
+type PIEInputElement = Partial<HTMLInputElement> & {
```
jamieomaguire, line 9: this is simply a PoC interface that could represent what kinds of methods and properties our own input components expose to consumers

packages/components/pie-form-label/src/index.ts
```diff
@@ -30,6 +35,20 @@ export class PieFormLabel extends RtlMixin(LitElement) implements FormLabelProps
         return optional ? html`<span class="c-formLabel-optional">${optional}</span>` : nothing;
     }
 
+    private handleClick () {
```
jamieomaguire, line 38: in a nutshell, if the input component this targets exists, either just focus it or if it can be checked (a check box or switch for example) then check it as well

packages/components/pie-switch/src/index.ts
```diff
@@ -50,6 +51,8 @@ export class PieSwitch extends RtlMixin(LitElement) implements SwitchProps {
         this.handleFormAssociation();
     }
 
+    static shadowRootOptions = { ...LitElement.shadowRootOptions, delegatesFocus: true };
```
jamieomaguire, line 54: This basically tells the component that if it receives a focus call, to pass that focus on to whatever the first focusable element in the template is. In this case, it is the input element. This means consumers can call focus on the input without having to know of the internal component template.

packages/components/pie-switch/src/index.ts
```diff
@@ -99,9 +102,14 @@ export class PieSwitch extends RtlMixin(LitElement) implements SwitchProps {
     /**
      * The onChange function updates the checkbox state and emits an event for consumers.
      * @param {Event} event - This should be the change event that was listened for on an input element with `type="checkbox"`.
+     * @param {boolean | null} checkedOverride - This is an optional parameter that allows consumers to override the checked state of the switch if they want to manually trigger a change event
      */
-    onChange (event: Event) {
-        const { checked } = event?.currentTarget as HTMLInputElement;
+    onChange (event: Event, checkedOverride: boolean | null = null) {
```
jamieomaguire, line 107: I have extended this method to accept a boolean override for the checked state. This allows our label component to explicitly set the checked state if it needs to but still allows this component to be responsible for dispatching it's own change event

packages/components/pie-switch/src/index.ts
```diff
@@ -197,7 +205,9 @@ export class PieSwitch extends RtlMixin(LitElement) implements SwitchProps {
                 class="c-switch-wrapper"
                 ?isRTL=${isRTL}
                 ?isDisabled=${isDisabled}>
-                ${labelPlacement === 'leading' ? this.renderSwitchLabel() : nothing}
+                <pie-form-label for="pie-switch">
```
jamieomaguire, line 208: I only placed the form-label component in this template so the styling looked nicer. It might be preferable to place it next to this component in the consumer template. Either approach works the same so it is a matter of preference

packages/components/pie-form-label/src/index.ts
jamieomaguire: I will need to confirm that the `document` reference doesn't break SSR on consumers

apps/pie-storybook/stories/pie-switch.stories.ts
```diff
@@ -119,6 +119,10 @@ const Template : TemplateFunction<SwitchProps> = (props) => {
         required,
     } = props;
 
+    // can put the label adjacent to the switch OR put directly in the switch component(check the switch index.ts file to see example)
+    // <pie-form-label for="pie-switch">
+    //     label slot content
+    // </pie-form-label>
```
jamieomaguire, line 125: This is the approach I would suggest we go with for usage. It is more flexible as the alternative means passing a selector down to use with the label

packages/components/pie-switch/src/index.ts
jamieomaguire: reference: https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot/delegatesFocus

packages/components/pie-form-label/src/index.ts
```diff
@@ -38,6 +67,7 @@ export class PieFormLabel extends RtlMixin(LitElement) implements FormLabelProps
 
         return html`
             <label
+                @click=${this.handleClick}
```
jamieomaguire, line 70: we would still need to ensure that this works like a native label from an A11y perspective - screenreaders should treat this like a native HTML label

raoufswe:
this has already been implemented in pie-form-label 
https://github.com/justeattakeaway/pie/pull/1257

PR #1122, opened 2023-12-19
test(pie-switch): DSW-1556 add test to prevent accessibility regression

xander-marjoram:

There was an issue where the switch component wasn't being announced properly by NVDA. I discovered that the styling changes that @jamieomaguire and I made as part of #1103 fixed the issue, so this PR adds a test which covers the key part of those changes to prevent a regression.

### Reviewer 1

### Reviewer 2

### DIFF
--- a/packages/components/pie-switch/test/component/pie-switch.spec.ts
+++ b/packages/components/pie-switch/test/component/pie-switch.spec.ts
@@ -28,6 +28,17 @@ test.describe('Component: `Pie switch`', () => {
         await expect(pieSwitch).toBeVisible();
     });
 
+    test('should have a "visible" input to help with accessibility', async ({ mount, page }) => {
+        // Arrange
+        await mount(PieSwitch);
+
+        // Act
+        const input = page.locator(inputSelector);
+
+        // Assert
+        await expect(input).toBeVisible();
+    });
+
     test('should set `checked` to `false` by default', async ({ mount }) => {
         // Arrange
         const component = await mount(PieSwitch);


packages/components/pie-switch/test/component/pie-switch.spec.ts
```diff
@@ -28,6 +28,17 @@ test.describe('Component: `Pie switch`', () => {
         await expect(pieSwitch).toBeVisible();
     });
 
+    test('should have a "visible" input to help with accessibility', async ({ mount, page }) => {
```
kevinrodrigues, line 31: Is this to account for the off screen input i.e the way we hide it? If so could the description be updated to something like below?

`should have an input element that is a readable and appropriately associated input element for screen readers.`

kevinrodrigues (approved):

packages/components/pie-switch/test/component/pie-switch.spec.ts
xander-marjoram: Previously we were hiding it with `appearance: none;` which NVDA didn't like, because it didn't want to describe an input which wasn't shown. Now we just make it only a single pixel in size; this way it's still technically displayed and the screen reader likes it.

fernandofranca (approved):

packages/components/pie-switch/test/component/pie-switch.spec.ts
kevinrodrigues: Yeah so i think the test description might read better to explain that. More of a NIT. At first i thought the input was visible on screen but saw the style changes and realised it was hidden with the css using the 1px method.

PR #1107, opened 2023-12-13
fix(pie-icons-webc): DSW-1539 Remove :host-context usage for icon sizing

kevinrodrigues:
## Describe your changes (can list changeset entries if preferable)

Before:
Screenshot 2023-12-14 at 11 52 07

After:

Screenshot 2023-12-14 at 11 49 32

## Author Checklist (complete before requesting a review)
- [ ] I have performed a self-review of my code
- [ ] If it is a core feature, I have added thorough tests
- [ ] If it is a `PIE Docs` change, I have reviewed the Docs site preview
- [ ] If it is a component change, I have reviewed the Storybook preview
- [ ] If there are visual test updates, I have reviewed them properly before approving

### Reviewer 1
- [ ] If it is a `PIE Docs` change, I have reviewed the PR preview
- [ ] If there are visual test updates, I have reviewed them

### Reviewer 2
- [ ] If it is a `PIE Docs` change, I have reviewed the PR preview
- [ ] If there are visual test updates, I have reviewed them

PR #1123, opened 2023-12-19
feat(pie-tag): DSW-1520 tag component functionality

dandel10n:

---
"@justeattakeaway/pie-tag": minor
---

[Added] Tag component functionality

🎟️  [Jira ticket with the requirements](https://justeattakeaway.atlassian.net/browse/DSW-1520)

🎨  [Figma designs](https://www.figma.com/file/OOgnT2oNMdGFytj5AanYvt/branch/QGEtmJqZM3OL9QG33L4053/%E2%9D%8C-%5BBETA%5D-%5BCore%5D-Component-Documentation-%5BPIE-3%5D-%E2%9D%8C?type=design&node-id=419-62146&mode=design&t=xWB9LgRzGqZnIf3g-0)

**Note 1:** Jira ticket specified `disabled` prop that at the end we decided not to add as the component is not interactive. Instead we added a `--tag-opacity` css variable that can be overridden on consumer side.

**Note 2:** Jira ticket specified that aria-label is needed for the tag. According to the latest Tag design Tag always has text. I read [aria-label MDN docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label), [aria-roledescription MDN-docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-roledescription) and checked some other popular UI libraries and came to the conclusion that aria object is not needed for the tag as its purpose is to give a clear concise category information and that is done by tag text itself.

---
---

[Added] :host-context(pie-tag) svg styles

---
---

[Removed] getShadowElementStylePropValues function from button component test to pie-webc-testing helpers

---
"@justeattakeaway/pie-webc-testing": minor
---

[Added] getShadowElementStylePropValues helper function

### Reviewer 1

### Reviewer 2


packages/components/pie-tag/src/index.ts
```diff
@@ -11,10 +13,38 @@ const componentSelector = 'pie-tag';
...
+        } = this;
+        return html`
+            <div
+                class="c-tag"
+                variant=${variant}
+                size=${size}
+                ?isStrong=${isStrong}
+                data-test-id="pie-tag"
+            >
+                ${size === 'large' ? html`<slot name="icon"></slot>` : nothing}
```
kevinrodrigues, line 45: Do you need a check to make sure that the icon is enabled? I'd assume you can show a large tag element without an icon too so here i think if size was large but we didn't want an icon it would render something here?

packages/tools/pie-icons-webc/build.js
```diff
@@ -44,6 +44,12 @@ export class ${name} extends LitElement implements IconProps {
             width: var(--btn-icon-size);
             height: var(--btn-icon-size);
         }
+
+        :host-context(pie-tag) svg {
```
kevinrodrigues, line 48: You might need to change this a bit as host-context doesn't work in firefox and safari. Unless you got this working somehow? Here's a pr i made for ref: https://github.com/justeattakeaway/pie/pull/1107/files#diff-beb4382d1fd71394d71c0f0f07f2e5111af3e72f86d94ce723a1a03b48251af9R41

packages/components/pie-tag/test/component/pie-tag.spec.ts
```diff
@@ -17,4 +46,68 @@ test.describe('PieTag - Component tests', () => {
         // Assert
         expect(tag).toBeVisible();
     });
+
+    test.describe('icon slot', () => {
+        test.describe('when passed', () => {
```
kevinrodrigues, line 51: You could remove one of these describe blocks and just do the follow:

```javascript
test.describe('when passed in with a size `large`', () => {
```

packages/components/pie-tag/test/component/pie-tag.spec.ts
```diff
@@ -17,4 +46,68 @@ test.describe('PieTag - Component tests', () => {
...
+                    // Act
+                    const tagIcon = page.locator(tagIconSelector);
+
+                    // Assert
+                    expect(tagIcon).toBeVisible();
+                });
+            });
+        });
+
+        test.describe('if the size is small', () => {
```
kevinrodrigues, line 72:
```javascript
test.describe('when passed in with a size of `small`', () => {
```

packages/components/pie-tag/test/component/pie-tag.spec.ts
```diff
@@ -17,4 +46,68 @@ test.describe('PieTag - Component tests', () => {
...
+                    isStrong: true,
+                },
+                slots: {
+                    default: 'Label',
+                },
+            });
+
+            const [currentBgStyle, expectedBgStyle] = await getShadowElementStylePropValues(component, componentSelector, ['--tag-bg-color', bgStyle]);
+
+            await expect(currentBgStyle).toBe(expectedBgStyle);
```
kevinrodrigues, line 110: Do you need the further await here?

packages/components/pie-tag/src/index.ts
xander-marjoram: @kevinrodrigues As it's a slot, if they don't want an icon then they just don't pass one in 😄

kevinrodrigues: 🤦 Yes you're right.

packages/tools/pie-icons-webc/build.js
dandel10n: Hmm, does your change `:host svg {` mean that all the svgs will have `var(--btn-icon-size, 24px)`? It doesn't feel right. I think we need to find a way to dictate the size prop value to the icon slot as different components that use icons will need different icon sizes

dandel10n: I checked link component with an icon in Firefox after your change and not it looks off 😞 :

Screenshot 2023-12-21 at 10 49 26

packages/components/pie-tag/test/component/pie-tag.spec.ts
dandel10n: I updated the test to include "when NOT passed" scenario as well so now it makes more sense as is

packages/components/pie-tag/src/tag.scss
```diff
@@ -1 +1,152 @@
...
+    --icon-display-override: block;
+    --icon-size-override: 16px;
+
+    display: inline-flex;
+    vertical-align: middle;
+    align-items: center;
+    justify-content: center;
+    gap: var(--dt-spacing-a);
+    height: var(--tag-height);
+    padding: var(--tag-padding);
```
ashleynolan, line 38: I'd setup the default value of this above in your variable declarations (set to whatever the default size/variation is) – just makes it a bit easier to see whats being set as variables altogether.

Same for height/font-size etc.

packages/components/pie-tag/src/tag.scss
```diff
@@ -1 +1,152 @@
 @use '@justeattakeaway/pie-css/scss' as p;
+
+*,
+*:before,
+*:after {
+    box-sizing: border-box;
```
ashleynolan, line 6: You shouldn't have to set box-sizing in the component styles – this gets set through `pie-css` at an application level (which is an expected dependency).

raoufswe (approved):
🥇

kevinrodrigues (approved):

apps/pie-storybook/stories/pie-tag.stories.ts
```diff
@@ -1,36 +1,118 @@
...
+        },
+        isStrong: {
+            description: 'If `true`, displays strong tag styles for some of the variant',
+            control: 'boolean',
+            defaultValue: {
+                summary: false,
+            },
+        },
+        showIcon: {
+            description: 'Enable to see the example of Tag with icon. Available only for large tag size.',
```
raoufswe, line 53: nit, can be addressed in a follow up PR:
we probably need to mention `showIcon` is only used to control the story and not an actual prop.

packages/components/pie-tag/README.md
```diff
@@ -74,16 +74,40 @@ import { PieTag } from '@justeattakeaway/pie-tag/dist/react';
 
 | Property | Type | Default | Description |
 |---|---|---|---|
-| - | - | - | - |
+| size | `String` | `large` | Size of the tag. Can be either `large` or `small` |
+| variant | `String` | `neutral` | Variant of the tag, one of `variants` - `neutral-alternative`, `neutral`, `outline`, `ghost`, `blue`, `green`, `yellow`, `red`, `brand` |
+| isStrong | `Boolean` | `false` | If `true`, displays strong tag styles for `green`, `yellow`, `red`, `blue` and `neutral` variants'|
+
+Since the component is not interactive it doesn't have a disabled property. To give the Tag a disabled look please use `--tag-opacity` css variable. Recommended opacity level for disabled tag is 0.5.
```
raoufswe, line 81: wouldn't this approach get applied to all tags on the page since the variable isn't unique? for example, i have a Tag used in a header, and then another one in the footer, setting the variable to 0.5 will make both tags dimmed? if not, I feel this might be an extra load to our end users, maybe we can introduce a prop called isDimmed or something instead of isDisabled?

dandel10n: It shouldn't. For example if you have a disabled card on the page and you want a tag inside the card to look disabled you can override `--tag-opacity` only for a tag inside a disabled card smth like `.card[disabled] .tag {}`. Does this make sense?

packages/components/pie-tag/test/visual/pie-tag.spec.ts
```diff
@@ -1,14 +1,69 @@
...
+import { sizes, variants } from '@/defs';
 
-test.describe('PieTag - Visual tests`', () => {
-    test('should display the PieTag component successfully', async ({ page, mount }) => {
-        await mount(PieTag, {
-            props: {} as TagProps,
-        });
+// TODO: Currently setting the slot to use a straight up SVG
+//       This should be updated to use pie-icons-webc, but after some investigation, we think that we'll
+//       need to convert the webc icons to use Lit, as currently the components don't work well in a Node env like Playwright
```
jamieomaguire, line 21: @dandel10n I appreciate this has already been merged but this comment isn't correct. The webc Icons now use Lit. So we likely should not be passing raw SVGs. 

Please could you take a look at using a `pie-icon-webc` component in this test if you have another PR to do?

dandel10n: Sure, I will check and create a follow up PR 👍

packages/components/pie-tag/README.md
raoufswe: yea that makes a lot of sense, tho, I think having the behaviour applied as a prop might reduce the overhead of end users and will provide a simpler API imo.

PR #1128, opened 2023-12-22
feat(pie-button): DSW-1539 wip assigning size prop to an icon slot

dandel10n:
### The idea:

Currently to make sure that when we provide an icon slot in our web components we use css variables to ensure the correct icon size. This approach works but goes against size property used on icon component. The goal of this change is to assign size icon prop to the icon slot to make sure that the size prop is correct. We can still use a css var for the option when consumers pass custom svg instead of pie icon.

### Problems:
 - SSR:
    - there is an issue with webc icons package to be fixed first: https://justeattakeaway.atlassian.net/browse/DSW-1686
    - Tested with Tag component inside the button as a POC while icons don't render in SSR:
        - the logic works in vanilla js app when js is enabled;
        - Then I checked Next app and while with `yarn dev` it worked the same way as in vanilla js, it didn't work as I would expect even without ssr after `yarn build && yarn preview`. While the attr was attached to the top wrapper it was not passed down to shadow dom. And then with disabled js the attr was not attached even to the top level element :sadness:

#### Vanilla JS:

Screenshot 2024-04-12 at 16 06 05

#### NextJS prod build served locally when js is enabled:

Screenshot 2024-04-12 at 16 21 07

#### NextJS when js is disabled:

Screenshot 2024-04-12 at 16 02 06

### DIFF
--- a/.changeset/chilled-snails-cry.md
+++ b/.changeset/chilled-snails-cry.md
@@ -1,0 +1,5 @@
+---
+"@justeattakeaway/pie-icon-button": patch
+---
+
+[Changed] - `--btn-icon-display`, `--btn-icon-size` and `--btn-icon-size-default` to `--icon-display-override` and `--icon-size-override` variables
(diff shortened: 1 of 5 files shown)


siggerzz:
Closing this as it's quite stale. Feel free to re-open if this is still needed!

PR #1142, opened 2024-01-03
feat(pie-webc-core): DSW-1582 form Control Mixin

jamieomaguire:
---
---

[Added] - Form Control Mixin for adding native form association behaviour to web components

---
---

[Changed] - Re organised test structure and added playwright testing

---
---

[Changed] - Use the new formControlMixin to apply pre-existing form association behaviour


packages/components/pie-webc-core/src/test/functions/defineCustomElement.spec.ts
```diff
@@ -3,7 +3,7 @@ import { LitElement } from 'lit';
 import {
     vi, expect, it,
 } from 'vitest';
-import { defineCustomElement } from '../defineCustomElement';
+import { defineCustomElement } from '../../functions/defineCustomElement';
```
raoufswe, line 6: nit, can be addressed in a follow-up PR:

do you think the folder name should be `helpers` to stay consistent with the naming convention in pie-docs and other places?

jamieomaguire: Possibly! Happy to discuss but I think both are quite ambiguous 😂 

Will chat with you offline and fix forward if that's cool 👍

packages/components/pie-webc-core/README.md
```diff
@@ -56,3 +57,18 @@ Other dependencies may be added in future to make it easier for people to instal
 Check out our [contributing guide](https://github.com/justeattakeaway/pie/wiki/Contributing-Guide) for more information on [local development](https://github.com/justeattakeaway/pie/wiki/Contributing-Guide#local-development).
 
 To run the unit tests, simply run `yarn test --filter=pie-webc-core` from the root of the monorepo.
+
+## Testing
```
raoufswe, line 61: this is so useful, thank you for adding it 🥇 

nit: I think this applies to other components and packages as well. do you think we can add this perhaps to a more global readme?

jamieomaguire: Thanks :) I am chatting with @siggerzz and @JoshuaNg2332  about more globally applicable documentation. This is very specific to the webc core package as I've had to write some custom naming conventions to get playwright and vitest to work together nicely - I think we should figure out a global approach and then unify the documentation as you suggest 😃

raoufswe (dismissed):

raoufswe:
good stuff!
i assume we are going to update the button to use this mixin right?

jamieomaguire:
> good stuff! i assume we are going to update the button to use this mixin right?

yeah man! Originally I included that in this PR but I decided I would raise separate PRs to add it to pie-button, pie-switch and the new pie-input components. That way it reduces the risk and makes it easier to test and review 😃

raoufswe (approved):

packages/components/pie-webc-core/playwright-lit.config.ts
```diff
@@ -0,0 +1,18 @@
+import { defineConfig, devices } from '@sand4rt/experimental-ct-web';
+import { getPlaywrightConfig as getPlaywrightConfigOG } from '@justeattakeaway/pie-components-config';
+import type { PlaywrightTestConfig } from '@playwright/test';
+
+const getPlaywrightConfig = () : PlaywrightTestConfig => ({
+    ...getPlaywrightConfigOG(),
```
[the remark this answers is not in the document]
jamieomaguire: ah I missed that, yes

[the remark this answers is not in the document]
jamieomaguire: I will fix in the next PR if that's ok 👍

packages/components/pie-webc-core/src/mixins/formControl/formControlMixin.ts
```diff
@@ -0,0 +1,60 @@
+/* eslint-disable max-classes-per-file */
+import { LitElement } from 'lit';
+import type { GenericConstructor } from '../types/GenericConstructor';
+
+/**
+ * Interface for FormControl behavior.
+ */
+declare class _FormControlInterface {
```
xander-marjoram, line 8: Something for another PR, perhaps, but I've done a bit of reading and it seems like `declare class` is roughly the same as `interface`, so I wonder if you could use that here instead? It may also remove the need to disable the `max-classes-per-file` rule.

packages/components/pie-webc-core/src/test/mixins/formControlMixin/formControlMixin.browser.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+
+test.describe('FormControlMixin', () => {
+    // IMPORTANT: Mounting and Unmounting the component before each test ensures that any tests that do not explicitly
+    // mount the component will still have it available in Playwright's cache (loaded and registered in the test browser)
+    test.beforeEach(async ({ mount }) => {
+        await (await mount(MockComponent)).unmount();
+    });
+
+    test.describe('form property', () => {
+        test('should return null when not inside of a form', async ({ mount }) => {
```
xander-marjoram, line 12: I don't think this test name matches its contents

packages/components/pie-webc-core/src/test/mixins/formControlMixin/formControlMixin.browser.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+                <form-control-mixin-mock></form-control-mixin-mock>
+                `;
+                document.body.innerHTML = formHTML;
+            });
+
+            // Get the component from the page and add a queryable element
+            await page.evaluate(() => {
+                const component = document.querySelector('form-control-mixin-mock');
+                // Assuming your component has a 'form' property that references its associated form
+                const form = component?.form;
```
xander-marjoram, line 81: I'm curious to know if we could just assert on the value of this variable instead of having all the debug stuff?

raoufswe (approved):

packages/components/pie-webc-core/src/test/mixins/formControlMixin/formControlMixin.browser.spec.ts
jamieomaguire: updated!

packages/components/pie-webc-core/src/mixins/formControl/formControlMixin.ts
jamieomaguire: fully agreed - updated to use an interface!

packages/components/pie-webc-core/src/mixins/rtl/rtlMixin.ts
```diff
@@ -24,8 +17,8 @@ declare class _RTLInterface {
  * allowing you to easily determine the text direction within your component.
  *
  * @function
- * @param {Constructor<LitElement>} superClass - The LitElement class to extend with RTL functionality.
- * @returns {Constructor<_RTLInterface> & T} - A new class extending both LitElement and _RTLInterface.
+ * @param {GenericConstructor<LitElement>} superClass - The LitElement class to extend with RTL functionality.
```
ashleynolan, line 20: Just to keep up my noob TS questions. Why change this? (and what's the difference between the two)

ashleynolan (dismissed):

packages/components/pie-webc-core/src/test/mixins/formControlMixin/formControlMixin.browser.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+            // Arrange
+            const component = await mount(
+                MockComponent,
+                {
+                    props: {},
+                },
+            );
+
+            // Assert
+            expect(component).not.toBeNull();
```
kevinrodrigues, line 17: If it is not null it might be worth adding the expected output here.

packages/components/pie-webc-core/src/test/mixins/formControlMixin/formControlMixin.browser.spec.ts
```diff
@@ -0,0 +1,99 @@
+import { test, expect } from '@sand4rt/experimental-ct-web';
+import { MockComponent } from './MockComponent.ts';
+
+test.describe('FormControlMixin', () => {
+    // IMPORTANT: Mounting and Unmounting the component before each test ensures that any tests that do not explicitly
+    // mount the component will still have it available in Playwright's cache (loaded and registered in the test browser)
+    test.beforeEach(async ({ mount }) => {
+        await (await mount(MockComponent)).unmount();
```
kevinrodrigues, line 8: Do you need the brace around the mount?  Should be enough here?`await mount(MockComponent).unmount()`

packages/components/pie-webc-core/src/mixins/rtl/rtlMixin.ts
jamieomaguire: The latter - it's just been moved into a separate file to reuse between Mixins we create now that we have more than one. This is an example of a [constrained mixin](https://www.typescriptlang.org/docs/handbook/mixins.html), annoyingly we have to write a type for a constructor to achieve this. The constraint here is that we only want to apply these to classes that extend `LitElement`.

You'll see in that docs link that TS do a similar thing in their example and name it `GConstructor` 😄

packages/components/pie-webc-core/src/test/mixins/formControlMixin/formControlMixin.browser.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+    });
+
+    test.describe('form property', () => {
+        test('should not have an associated form when not inside of a form', async ({ mount }) => {
```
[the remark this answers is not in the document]
jamieomaguire: Yeah happy to make this change 👍

packages/components/pie-webc-core/src/test/mixins/formControlMixin/formControlMixin.browser.spec.ts
jamieomaguire: yes, the brace is needed to access properties on the return type of `mount` other you get the error `Property 'unmount' does not exist on type 'Promise<MountResult<MockComponent>>'.ts(2339)`

packages/components/pie-webc-core/src/test/mixins/formControlMixin/formControlMixin.browser.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+            // Arrange
+            await page.evaluate(() => {
+                const formHTML = `
+                <form id="testForm" action="/foo" method="POST">
+                    <input type="text" id="username" name="username" required>
+                    <input type="password" id="password" name="password" required>
+                    <form-control-mixin-mock></form-control-mixin-mock>
+                </form>
+                `;
+                document.body.innerHTML = formHTML;
```
kevinrodrigues, line 37: You could probably make use of `await page.setContent(formHTML)`

packages/components/pie-webc-core/src/test/mixins/formControlMixin/formControlMixin.browser.spec.ts
jamieomaguire: This was a mistake - the test has been revamped :)

packages/components/pie-webc-core/src/test/mixins/formControlMixin/formControlMixin.browser.spec.ts
jamieomaguire: excellent idea! Will update

packages/components/pie-webc-core/src/test/mixins/formControlMixin/formControlMixin.browser.spec.ts
kevinrodrigues: I see, I think in that case because it is returning the promise it is looking for the unmounted value you can destructure it and then await that. I think it is a little more readable that way.

```javascript
const { unmount } = await mount( MockComponent )
await unmount()
```

jamieomaguire: I personally find that to be less readable, so if it's ok I will leave for now - destructuring the class method seems a bit strange to me

xander-marjoram: I think if we are going to break it into two lines it should probably be:
```javascript
const component = await mount(MockComponent);
await component.unmount();
```

jamieomaguire: Yeah, I am happy with Xanders approach if you both prefer

kevinrodrigues: LGTM 👍

kevinrodrigues (approved):

xander-marjoram (approved):

PR #1145, opened 2024-01-04
refactor(pie-switch): DSW-1582 use form control mixin

jamieomaguire:
Tested on PIE Aperture NextJS (SSR) & Vanilla App

---
"@justeattakeaway/pie-switch": minor
---

[Changed] - Use new FormControlMixin for form association logic in `pie-switch` component

---
---

[Changed] - Use proper interface keyword for RTLmixin interface. Was using a class before which is incorrect.

---
---

[Removed] - unused LitElementMixin type as we do not need it anywhere


packages/components/pie-webc-core/src/mixins/rtl/rtlMixin.ts
```diff
@@ -6,7 +6,7 @@ import type { GenericConstructor } from '../types/GenericConstructor';
  * An interface representing the structure of RTL related class.
  * @interface
  */
-declare class _RTLInterface {
+export interface RTLInterface {
```
raoufswe, line 9: 🔥

raoufswe (approved):

dandel10n (approved):

PR #1159, opened 2024-01-09
fix(pie-components-config): DSW-1622 fix issue where TS / Vite config weren't published as part of the package

siggerzz:
- Tested the generator in another repo and can successfully generate and build a component (Once pie-css / vite-dts have been installed, but I think this is expected)

- Modified existing `extends` to reference published package tsconfig

- Able to build the monorepo

- I've checked in node_modules and can see the configs are now present (they're not currently in main):

Screen Shot 2024-01-09 at 16 36 42

- When this package is yalced into another repo, the configs are also present:

Screen Shot 2024-01-09 at 16 37 44

I've also tested in PIE Aperture NextJS that if we add the following property to the tsconfig.json:

`"extends": "@justeattakeaway/pie-components-config/tsconfig.json`

It fails as expected:

Screen Shot 2024-01-09 at 16 43 49

Once yalced, we see NextJS uses our shared config:

Screen Shot 2024-01-09 at 16 46 32

### Reviewer 1

### Reviewer 2

### DIFF
--- a/configs/pie-components-config/tsconfig.json
+++ b/configs/pie-components-config/tsconfig.json
@@ -8,11 +8,6 @@
     "sourceMap": true,
     "inlineSources": true,
     "outDir": "./compiled",
-    "baseUrl": ".",
-    "paths": {
-      "@/*": ["./src/*"]
-    },
-    "rootDir": ".",
     "strict": true,
     "noUnusedLocals": true,
     "noUnusedParameters": true,
--- a/packages/components/pie-button/tsconfig.json
+++ b/packages/components/pie-button/tsconfig.json
@@ -1,5 +1,5 @@
 {
-  "extends": "../../../configs/pie-components-config/tsconfig.json",
+  "extends": "@justeattakeaway/pie-components-config/tsconfig.json",
   "compilerOptions": {
     "baseUrl": ".",
     "rootDir": ".",
--- a/packages/components/pie-button/tsconfig.json
+++ b/packages/components/pie-button/tsconfig.json
@@ -3,9 +3,6 @@
   "compilerOptions": {
     "baseUrl": ".",
     "rootDir": ".",
-    "paths": {
-      "@/*": ["./src/*"]
-    },
   },
   "include": ["src/**/*.ts","./declaration.d.ts", "test/**/*.ts", "playwright-lit-visual.config.ts", "playwright-lit.config.ts"],
 }
--- a/packages/components/pie-button/tsconfig.json
+++ b/packages/components/pie-button/tsconfig.json
@@ -3,6 +3,9 @@
   "compilerOptions": {
     "baseUrl": ".",
     "rootDir": ".",
+    "paths": {
+      "@/*": ["./src/*"]
+    },
   },
   "include": ["src/**/*.ts","./declaration.d.ts", "test/**/*.ts", "playwright-lit-visual.config.ts", "playwright-lit.config.ts"],
 }
--- a/packages/components/pie-modal/test/visual/pie-modal.spec.ts
+++ b/packages/components/pie-modal/test/visual/pie-modal.spec.ts
@@ -1,10 +1,8 @@
-import { test, expect } from '@sand4rt/experimental-ct-web';
+import { test } from '@sand4rt/experimental-ct-web';
 import { percyWidths } from '@justeattakeaway/pie-webc-testing/src/percy/breakpoints.ts';
 import percySnapshot from '@percy/playwright';
-import { PieModal } from '@/index';
-import { ModalProps, sizes, positions } from '../../src/defs';
-
-const componentSelector = '[data-test-id="pie-modal"]';
+import { PieModal } from '../../src/index.ts';
+import { ModalProps, sizes, positions } from '../../src/defs.ts';
 
 sizes.forEach((size) => {
     test(`should render correctly with size = ${size}`, async ({ page, mount }) => {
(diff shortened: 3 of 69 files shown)


configs/pie-components-config/tsconfig.json
```diff
@@ -8,11 +8,6 @@
     "sourceMap": true,
     "inlineSources": true,
     "outDir": "./compiled",
-    "baseUrl": ".",
```
siggerzz, line 11: I removed these as they were causing issues when tests were referencing `@/foo.ts` due to relative path defined in the root config.

I'm not sure why the `paths` wasn't causing problems before, but I believe this is why we override the `baseUrl` / `rootDir` in all our configs.

packages/components/pie-button/tsconfig.json
```diff
@@ -1,8 +1,11 @@
 {
-  "extends": "../../../configs/pie-components-config/tsconfig.json",
```
jamieomaguire, line 2: Its smaller PRs but any reason we can't change each component (or smaller groups) in separate PRs?

configs/pie-components-config/tsconfig.json
jamieomaguire: I'm unsure if I'm a fan of removing the `paths` property here. Can you elaborate on the issues found? I looked at the linked failure but I noticed that's for a rather new component and I don't think we've seen this issue anywhere else

siggerzz: From the TS Documentation - https://www.typescriptlang.org/tsconfig#extends

>All relative paths found in the configuration file will be resolved relative to the configuration file they originated in.

Because of this, combined with the fact `pie-tag` didn't declare it's own `paths` property, the tests were essentially looking for the `/defs` directory from `node_modules/@justeattakeaway/pie-components-config/src`, as that's the config we're inheriting from.

packages/components/pie-button/tsconfig.json
siggerzz: No reason, but I do think it's unnecessary personally. I'm happy to do that if others don't have the same risk tolerance, but this change won't affect consumers, only PIE

jamieomaguire: It's less about consumer risk and more about tracking regressions. I am happy to go with the majority opinion

siggerzz: Sure thing.

We're not even using the shared config in the aperture apps, so there's no risk there for now.

I did do a quick test locally to extend the aperture apps tsconfigs and didn't observe and TS / IDE issues.

Also can't spot anything within PIE 👍🏼.

packages/components/pie-button/tsconfig.json
```diff
@@ -1,8 +1,11 @@
...
+  "extends": "@justeattakeaway/pie-components-config/tsconfig.json",
   "compilerOptions": {
     "baseUrl": ".",
     "rootDir": ".",
+    "paths": {
```
ashleynolan, line 6: If we're defining this `paths` setting in most of our components/tools already, why not leave it in and override it when wee don't want it set to this?

siggerzz: See this comment :) - https://github.com/justeattakeaway/pie/pull/1159#discussion_r1447161107

siggerzz: Caught up with @ashleynolan about this, and we'll be leaving the `paths` in the component TSConfig 👍🏼

siggerzz:
Documentation has now been added around where to provide any path aliases:

https://github.com/justeattakeaway/pie/wiki/PIE-Shared-Configuration

cc @ashleynolan @jamieomaguire

configs/pie-components-config/tsconfig.json
xander-marjoram: Is it only in tests that we're using `@/`? If so, I don't think we should adapt our tsconfig just for those. Especially as the docs say that [published libraries should not set up `paths`](https://www.typescriptlang.org/docs/handbook/modules/reference.html#paths-does-not-affect-emit).

siggerzz: Nice find! Yeah they are. Shall I remove the aliases entirely and refactor the tests to use relative path @xander-marjoram?

xander-marjoram: That sounds great to me, thanks!

fernandofranca (dismissed):

packages/components/pie-modal/test/visual/pie-modal.spec.ts
```diff
@@ -262,8 +259,8 @@ test.describe('Prop: `leadingAction`', () => {
                 } as ModalProps,
             });
 
-            const modal = page.locator(componentSelector);
-            await expect.soft(modal).toBeVisible();
+            // Follow up to remove in Jan
```
ashleynolan, line 262: Looks like this is overriding Josh's latest changes that went into main?

siggerzz: 🤦🏼

packages/tools/generator-pie-component/src/app/templates/test/accessibility/pie-placeholder.__spec__.ts
```diff
@@ -1,6 +1,6 @@
 
 import { test, expect } from '@justeattakeaway/pie-webc-testing/src/playwright/fixtures.ts';
-import { Pie<%= componentName %>, <%= componentName %>Props } from '@/index';
+import { Pie<%= componentName %>, <%= componentName %>Props } from '../../src/index.ts';
```
xander-marjoram, line 3: ⭐

xander-marjoram (approved):

ashleynolan (approved):

PR #1164, opened 2024-01-10
feat(pie-monorepo): DSW-1577 Evaluate components props and provide additional React props

fernandofranca:
### pie-wrapper-react
[Changed] - Updates the React wrapper script to use the [workaround suggested by the Lit team](https://github.com/lit/lit/issues/4219#issuecomment-1740033448) for the bug that sets all props as optional in React
[Changed] - Added workaround for missing callback props in React and TS
[Changed] - Updated the React Wrapper to read and include the React base type declaration
[Changed] - updated unit tests snapshots to reflect the changes

### pie-cookie-banner, pie-notification, pie-icon-button, pie-form-label, pie-divider, pie-spinner, pie-button, pie-switch, pie-input, pie-modal, pie-card, pie-link, pie-tag

[Added] - React base TypeScript types
[Changed] - set optional props as optional

### generator-pie-component:
[Added] - template file for React base type

### Reviewer 1

### Reviewer 2


packages/components/pie-card/src/defs-react.ts
```diff
@@ -0,0 +1,3 @@
+import React from 'react';
+
+export type ReactBaseType = React.AnchorHTMLAttributes<HTMLAnchorElement> & React.ButtonHTMLAttributes<HTMLButtonElement>
```
fernandofranca, line 3: This is a polymorphic component, can be either `a` or `button`

packages/components/pie-switch/src/defs-react.ts
```diff
@@ -0,0 +1,3 @@
+import React from 'react';
+
+export type ReactBaseType = React.InputHTMLAttributes<HTMLInputElement>
```
fernandofranca, line 3: I'm not entirely sure if it should use Input as base type as it is wrapped in a div and a label

packages/components/pie-notification/src/defs-react.ts
```diff
@@ -0,0 +1,3 @@
+import React from 'react';
+
+export type ReactBaseType = React.HTMLAttributes<HTMLDivElement>
```
fernandofranca, line 3: As the component is WIP, this definition should be revisited later

packages/components/pie-modal/src/index.ts
```diff
@@ -387,8 +387,8 @@ export class PieModal extends RtlMixin(LitElement) implements ModalProps {
         <dialog
             id="dialog"
             class="c-modal"
-            size="${size}"
-            position="${position}"
+            size="${size || 'medium'}"
+            position="${position || 'center'}"
```
fernandofranca, line 391: Since we want to make sure `size` and `position` are always set, the default values are passed again, instead of using `ifDefined` which would omit the attribute in case the user passes something like "falsey" value.

What bothers me here is repeating the default value, since we already use them in the class fields. I believe this can be error-prone since we define the same values in different places using plain strings.

I wonder if it wouldn't be better if we had a single immutable object that stores the component default values, so we could use its reference wherever needed, saving us from potential regressions.

xander-marjoram: If repeating is indeed unavoidable then I prefer this approach

packages/components/pie-notification/src/defs-react.ts
xander-marjoram: Might be worth a to-do comment and a note in the notification epic? (Seems too small for its own ticket)

packages/components/pie-switch/src/defs-react.ts
xander-marjoram: I think this is okay, we do try to have it behave as a native checkbox by exposing some properties of the inner input element

jamieomaguire: @fernandofranca does this tell React TS users that they can apply any HTML Input attributes to the component?

packages/components/pie-button/src/index.ts
```diff
@@ -230,9 +230,9 @@ export class PieButton extends FormControlMixin(LitElement) implements ButtonPro
             <button
                 @click=${this._handleClick}
                 class="o-btn"
-                type=${type}
-                variant=${variant}
-                size=${size}
+                type=${type || 'submit'}
```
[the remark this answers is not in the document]
jamieomaguire: This is my concern as well. It seems we now set defaults in two places which I don't think is ideal

packages/components/pie-button/src/index.ts
```diff
@@ -230,9 +230,9 @@ export class PieButton extends FormControlMixin(LitElement) implements ButtonPro
...
+                variant=${variant || 'variant'}
```
jamieomaguire, line 234: I don't think 'variant' is a valid variant

packages/components/pie-modal/src/index.ts
jamieomaguire: Agreed - the default object approach sounds like a great idea @fernandofranca!

packages/tools/generator-pie-component/src/app/templates/src/defs-react.ts
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+/**
+ * TODO: Verify if ReactBaseType can be set as a more specific React interface
```
jamieomaguire, line 3: Please can we create a ticket to investigate/do this?

fernandofranca: In fact, that is a reminder, not a leftover. With that, we can easily identify if the React base type wasn't taken care of for any generated  component.

packages/components/pie-switch/src/defs-react.ts
fernandofranca: @jamieomaguire 👍 Exactly!
Important caveat, it doesn't mean such attributes will be forwarded to the implementation.

fernandofranca:
I updated `pie-wrapper-react` to add the types for the events callbacks as the workaround for "props marked as optional" got rid of them.

Now it shouldn't be an issue for React+TS consumers.

image

image

If you want to check how the wrapped React components work on `pie-aperture` check out the branch `dsw-1577-test-snapshot-release`.

packages/components/pie-notification/src/defs-react.ts
fernandofranca: Added a `TODO` comment

jamieomaguire (dismissed):

packages/components/pie-link/src/index.ts
```diff
@@ -120,14 +121,14 @@ export class PieLink extends LitElement implements LinkProps {
...
-                underline=${this.underline}
+                tag=${this.tag || 'a'}
+                variant=${this.variant || 'default'}
+                size=${this.size || 'medium'}
+                underline=${this.underline || 'default'}
                 ?isBold=${this.isBold}
                 ?isStandalone=${this.isStandalone}
                 ?hasVisited=${this.hasVisited}
-                href=${this.href || nothing}
+                href=${ifDefined(this.href)}
```
kevinrodrigues, line 131: Just checking. Will this line remove the href attribute if it is not defined? Should it default to empty instead if it does remove it?

kevinrodrigues (dismissed):
LGTM just one minor question.

xander-marjoram (approved):

kevinrodrigues (approved):

xander-marjoram (approved):

PR #1170, opened 2024-01-11
feat(pie-input): DSW-1582 add vanilla form submission

jamieomaguire:
---
"pie-storybook": minor
---

[Added] - name property to `pie-input` component

---
---

[Added] - `updated` lifecycle method to ensure we update the form value when the value prop is changed externally

### DIFF
--- a/packages/components/pie-input/src/index.ts
+++ b/packages/components/pie-input/src/index.ts
@@ -1,4 +1,6 @@
-import { LitElement, html, unsafeCSS } from 'lit';
+import {
+    LitElement, html, unsafeCSS, PropertyValues,
+} from 'lit';
 import { property } from 'lit/decorators.js';
 import { ifDefined } from 'lit/directives/if-defined.js';
 import { live } from 'lit/directives/live.js';
@@ -29,20 +31,30 @@ export class PieInput extends FormControlMixin(RtlMixin(LitElement)) implements 
     @property({ type: String })
     public value = '';
 
+    @property({ type: String })
+    public name = '';
+
+    protected firstUpdated (_changedProperties: PropertyValues<this>): void {
+        super.firstUpdated(_changedProperties);
+        this._internals.setFormValue(this.value);
+    }
+
     /**
      * Handles data processing in response to the input event. The native input event is left to bubble up.
      * @param event - The input event.
      */
     private handleInput = (event: InputEvent) => {
         this.value = (event.target as HTMLInputElement).value;
+        this._internals.setFormValue(this.value);
     };
 
     render () {
-        const { type, value } = this;
+        const { type, value, name } = this;
 
         return html`<input
             type=${ifDefined(type)}
             .value=${live(value)}
+            name=${ifDefined(name)}
             @input=${this.handleInput}
             data-test-id="pie-input">`;
     }
--- a/packages/components/pie-input/src/index.ts
+++ b/packages/components/pie-input/src/index.ts
@@ -8,7 +8,7 @@ import {
 } from '@justeattakeaway/pie-webc-core';
 
 import styles from './input.scss?inline';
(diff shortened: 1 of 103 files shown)


packages/components/pie-input/test/component/pie-input.spec.ts
```diff
@@ -1,10 +1,72 @@
 
 import { test, expect } from '@sand4rt/experimental-ct-web';
+import type { Page } from '@playwright/test';
 import { PieInput, InputProps } from '../../src/index.ts';
 
 const componentSelector = '[data-test-id="pie-input"]';
 
+/**
+ * Sets up form data extraction for testing form submissions. This function expects a form element
```
kevinrodrigues, line 9: Could we move this to a test/helpers directory please?

jamieomaguire: I was planning on doing that in a follow up PR if that's ok @kevinrodrigues as I'd like to get people's thoughts on how that project structure should look. Is that alright?

packages/components/pie-input/test/component/pie-input.spec.ts
```diff
@@ -171,4 +261,56 @@ test.describe('PieInput - Component tests', () => {
             });
         });
     });
+
+    test.describe('Form integration', () => {
+        test('should appear within the form data object when the form is submitted', async ({ page }) => {
```
kevinrodrigues, line 266: I think this test description needs updating slightly. Something like below:

`should set the "formDataObject" value of username to the correctly when submitted`

packages/components/pie-input/test/component/pie-input.spec.ts
```diff
@@ -74,6 +136,34 @@ test.describe('PieInput - Component tests', () => {
                 expect((await input.inputValue())).toBe('test');
             });
         });
+
+        test.describe('name', () => {
+            test('should default to an empty string if no name prop provided', async ({ mount }) => {
```
kevinrodrigues, line 141: Does this actually default to an empty string? I thought that `ifDefined` will not set an attribute if the value does not exist?

packages/components/pie-input/test/component/pie-input.spec.ts
kevinrodrigues: Yeah sure 👍

packages/components/pie-input/test/component/pie-input.spec.ts
kevinrodrigues: Actually i think i see now. I think it is removing it but the test description may be a little misleading. Probably needs to be read: `should not render a name attribute if no name prop is provided`?

jamieomaguire: in my testing, even with that approach if you programmatically query an `input` element for the attribute (name and value at least) then it will return an empty string in the browser. So I think Lit is not adding anything, and browser defaults to empty string but not as an attribute you can see in the inspected html

packages/components/pie-input/src/index.ts
```diff
@@ -29,20 +31,38 @@ export class PieInput extends FormControlMixin(RtlMixin(LitElement)) implements
     @property({ type: String })
     public value = '';
 
+    @property({ type: String })
+    public name = '';
+
+    protected firstUpdated (_changedProperties: PropertyValues<this>): void {
```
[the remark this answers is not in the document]
jamieomaguire: I think its entirely possible. I think we should start by waiting to see what parts get repeated each time we add a new form control. Otherwise we risk creating a mixin that is too rigid in its assumptions about a component's implementation.

Once we see need for duplication, we can bring the logic into the mixin 😃

kevinrodrigues (approved):

raoufswe (approved):

PR #1176, opened 2024-01-12
feat(pie-docs): DSW-1565 added show more button

LTurns:

[Added] - show more button for component status page

Tested on branch `dsw-1565-status-page` - see deployment here: https://pr1168-docs.pie.design/components/component-status/
System test added on the above branch but merging shortcodes and page separately 👍 

https://github.com/justeattakeaway/pie/assets/49618712/2954cafc-6e6e-4c89-b291-e444d47bc51d

### Reviewer 1

### Reviewer 2


raoufswe:
looks like the first interaction of the show more doesn't work. i had to click two times for it to work 

https://github.com/justeattakeaway/pie/assets/28605803/74149af5-a4e4-4cf5-b9fd-944f0095ac52

apps/pie-docs/src/assets/styles/components/_showMore.scss
```diff
@@ -0,0 +1,22 @@
+@use '../fozzie-config' as f;
+
+.c-showMore-backdrop {
```
ashleynolan, line 3: To fit with the naming scheme, it would make more sense to call this class the `c-showMore` and then the button class below `c-showMore-btn`

apps/pie-docs/src/assets/styles/components/_showMore.scss
```diff
@@ -0,0 +1,22 @@
...
+    display: flex;
+    align-items: center;
+    justify-content: center;
+}
+
+.c-showMore {
+    cursor: pointer;
+    color: f.$color-content-link-distinct;
```
ashleynolan, line 11: You could use CSS variables here now 😄

ashleynolan (commented):
Couple of general UX-y comments (don't have to stop this PR, but would be good to look at):

- Is the plan to keep the component status page in amongst all of the other components in the sideNav? It's pretty hard to find atm (to me it would make more sense at the top of that section, above all the other components listed?)
- When opening the "Show more" the column widths change, which looks a bit jarring. Would be good to fix a couple of the column widths potentially to stop that happening.
- When clicking "Show less", would be good to move the user back up the page to the web section when closed. Otherwise it leaves you at the bottom of the page and you have to scroll back up to get back to that section.

Thanks for these! Yep the component status page will be at the top of that section, I haven't done this yet but plan on updating the nav in the component-status-page PR once the shortcode PRs are merged 🎉 
Thanks for spotting the change of widths! I'll investigate this some more
ah yes 100% agree there should be an automatic scroll up to the top of the table when the 'show less' is closed! I'll add that in now 👍

apps/pie-docs/src/assets/js/show-more.js
```diff
@@ -1,29 +1,27 @@
...
     const componentStatusTables = document.querySelectorAll('[class="c-componentStatusTable "]');
 
-    if (!showMoreButtons) return;
-
     // ensures status tables initially only show 6 columns each
-    if (componentStatusTables) {
-        componentStatusTables.forEach((el) => {
-            el.querySelectorAll('tbody tr').forEach((tr, index) => {
-                const rows = index >= 6 ? tr.style.display = 'none' : '';
+    componentStatusTables.forEach((el) => {
```
raoufswe: nit: I would add an early return if the length of showMoreButtons or componentStatusTables before the `forEach` just to make sure the script doesn't necessarily run for other pages

apps/pie-docs/src/assets/js/show-more.js
```diff
@@ -0,0 +1,59 @@
+let shown = false;
+let showMoreButtons;
+
+window.addEventListener('DOMContentLoaded', () => {
+    showMoreButtons = document.querySelectorAll('[data-js="show-more"]');
+    const componentStatusTables = document.querySelectorAll('[class="c-componentStatusTable "]');
+
+    if (!showMoreButtons) return;
+    if (!componentStatusTables) return;
```
raoufswe, line 9: both values are `nodeList` so we need to check the length here cause empty list is a truthy value. 

```suggestion
    if (!showMoreButtons.length || componentStatusTables.length ) return;
```

LTurns: 🤦‍♀️ thank youu!

raoufswe (approved):

apps/pie-docs/src/assets/js/show-more.js
```diff
@@ -0,0 +1,58 @@
...
+    const componentStatusTableSection = document.querySelector(`[id="${elm.id.split('-')[0]}"]`);
+
+    const hide = () => {
+        trs.forEach((tr, index) => {
+            const rows = index >= 6 ? tr.style.display = 'none' : '';
+
+            return rows;
+        });
+
+        elm.innerHTML = 'Show More <svg xmlns="http://www.w3.org/2000/svg" role="presentation" focusable="false" fill="#125fca" viewBox="0 0 16 16" height="16" width="16" class="c-pieIcon c-pieIcon--chevronDown u-iconFilled" aria-hidden="true"><path d="M2.82 5.044 8 10.399 13.197 5l.963.875-5.364 5.565a1.164 1.164 0 0 1-1.636 0L1.875 5.945l.945-.901Z"></path></svg>';
```
ashleynolan, line 38: Again, just a small thing, but rather than have hardcoded SVGs, you could use the filter that is used to display the pieIcons svgs from the library:

```
const pieIcons = require('../filters/pieIconsSvg')();
const chevronDownIcon = pieIcons.find('chevronDown');
```

apps/pie-docs/src/assets/js/show-more.js
```diff
@@ -0,0 +1,58 @@
...
+    };
+
+    const show = () => {
+        trs.forEach((tr) => {
+            tr.style.display = 'table-row';
+
+            return tr;
+        });
+
+        elm.innerHTML = 'Show Less <svg xmlns="http://www.w3.org/2000/svg" role="presentation" focusable="false" fill="#125fca" viewBox="0 0 16 16" height="16" width="16" class="c-pieIcon c-pieIcon--chevronUp u-iconFilled" aria-hidden="true"><path d="M13.18 10.97 8 5.615l-5.18 5.399-.962-.875 5.346-5.565a1.164 1.164 0 0 1 1.671 0l5.25 5.495-.945.901Z"></path></svg>';
```
ashleynolan, line 51: Same here as above

apps/pie-docs/src/assets/js/show-more.js
LTurns: Hey! I did intend on using this approach originally, however, I'm not sure it's possible due to how the show-more button is rendered - because I'm loading this js module through a script tag, I get the error `ReferenceError: require is not defined` because the browser does not support the require() function 😞 I'm not sure how best to tackle it, so went for the hardcoded values for now - happy to look into this separately though as the search.js module could also benefit from this being able to be imported 👍

ashleynolan (approved):

PR #1180, opened 2024-01-15
fix(pie-docs): DSW-1604 update all spacing styles to use pie-css

raoufswe:

- updates all references of fozzie spacing tokens to use pie-css

### Reviewer 1

### Reviewer 2


ashleynolan (dismissed):

apps/pie-docs/src/assets/styles/components/_contentPage.scss
```diff
@@ -7,17 +7,17 @@
     min-height: 100%;
 
     > * {
-        padding-left: f.spacing(d);
-        padding-right: f.spacing(d);
+        padding-inline-start: var(--dt-spacing-d);
```
LTurns, line 10: nit but could these be shortened to `padding-inline: var(--dt-spacing-d) var(--dt-spacing-d)` or maybe just `padding-inline: var(--dt-spacing-d)`? I think If one value is given, it is used as the value for both inline start and end 😄

xander-marjoram: It's a great shout @LTurns but I think we have to avoid this specific shorthand because it's not supported by Safari 14 😞

raoufswe: good point! i will address this in the next PR.

raoufswe: @xander-marjoram I see we already use them several times in the codebase, so we probably need to clean that up at some point

apps/pie-docs/src/assets/styles/components/_card.scss
```diff
@@ -3,20 +3,20 @@
 
 $card-min-width: 270px;
 $card-image-height: 180px;
-$card-gap: f.$spacing-e;
-$card-container-max-width: calc(($card-min-width * 3) + ($card-gap * 2));
+$card-gap: var(--dt-spacing-e);
+$card-container-max-width: calc(($card-min-width * 3) + 48px); // 48px is card-gap * 2
```
fernandofranca, line 7: Doesnt calc work for `$card-gap`?

webbertakken (approved):
LGTM

apps/pie-docs/src/assets/styles/components/_contentPage.scss
LTurns: oh I see, thank you! I've seen some `padding-inline` and `padding-block` shortcodes dotted around (just spotted one [here](https://github.com/justeattakeaway/pie/pull/1180/files#diff-a1a7b39ce88d57776b4d4176d9777d9a27fad605af9debe8c30e911e41290b1fL37)) - can defo be removed in a different PR though 🎉

LTurns: oh sorry @raoufswe just seen you beat me to it! they do seem to be in lots of different areas - shall we make a ticket for this? I'm happy to write something up 👍

fernandofranca (approved):
🔥🔥🔥

apps/pie-docs/src/assets/styles/components/_card.scss
ashleynolan: I think we'd need to convert the variable to be a CSS variable (but we should do this imo and phase out the SCSS variables where possible)

apps/pie-docs/src/assets/styles/components/_contentPage.scss
raoufswe: @LTurns yea that would be great! thanks

apps/pie-docs/src/assets/styles/components/_card.scss
raoufswe: @fernandofranca when I changed the value of `$card-gap` to use a css variable, the calculation of `card-container-max-width` didn't work anymore 👀 i assume it's related to the unit conversions but I'm not 100% sure. maybe @ashleynolan would know?

i will also address scss variables cleanup in my follow up PR if you don't mind.

ashleynolan: Let me know if you have issues with this in the followup – can help you make the change 😄

PR #1197, opened 2024-01-19
fix(pie-docs): DSW-1604 replace fozzie typography

raoufswe:

[Changed] - Replaced fozzie typography tokens with pie-css tokens

### Reviewer 1

### Reviewer 2


apps/pie-docs/src/assets/styles/base/_layout.scss
```diff
@@ -99,20 +100,24 @@
     .c-layout-item,
     .c-content-container {
         h2 {
-            @include f.font-size('heading-xl', $scale: 'narrow');
+            @include p.font-size(--dt-font-heading-xl-size--narrow);
+            @include p.line-height(--dt-font-heading-xl-line-height--narrow);
```
kevinrodrigues, line 104: Question around this. Would it make sense to update the `to-px` function or have a font-size function that calculates the `px` and then returns a font-size and `line-height` together (setting a line-height to a default value if not passed in)? I think this may save on having to remember to include the line-height call everytime we include the font-size call what do you think?

raoufswe: yea, i agree and I already shared it in the engineering channel but i decided to keep it out of the scope of this ticket as its gonna require more work that i initially anticipated.

kevinrodrigues (approved):

apps/pie-docs/src/assets/styles/components/_centredPage.scss
```diff
@@ -18,10 +19,10 @@
 
     .c-centredPage-heading {
         margin-block-end: 0;
-        font-size: f.$font-heading-xxl-size--narrow * 1px;
+        font-size: #{p.font-size(--dt-font-heading-xxl-size--narrow)};
```
xander-marjoram, line 22: Is there a difference between these two syntaxes?
```suggestion
        @include p.font-size(--dt-font-heading-xxl-size--narrow);
```

raoufswe: nope, both should do the same thing. pie-css ships two ways of setting font-sizes. scss functions and mixins.

xander-marjoram (approved):

PR #1209, opened 2024-01-23
feat(pie-input): DSW-1582 add text input props

jamieomaguire:
---
"pie-storybook": minor
---

[Added] - pattern, minlength and maxlength attributes to pie-input component


packages/components/pie-input/src/index.ts
```diff
@@ -27,24 +27,44 @@ export class PieInput extends FormControlMixin(RtlMixin(LitElement)) implements
...
+     * https://developer.mozilla.org/en-US/docs/Web/API/HTMLObjectElement/validity
+     */
+    public get validity (): ValidityState {
+        return (this.input as HTMLInputElement).validity;
+    }
 
     protected firstUpdated (_changedProperties: PropertyValues<this>): void {
         super.firstUpdated(_changedProperties);
-        this._internals.setFormValue(this.value);
+        this._internals.setFormValue(this.value as string);
```
raoufswe, line 60: are we sitting the assertion here cause value can be undefined?

raoufswe (approved):

packages/components/pie-input/src/index.ts
xander-marjoram, line 60: Does this handle `undefined` nicely?

jamieomaguire: Not currently simply because it either defaults to an empty string or it is driven off of the value typed in. I can write some tests in a follow up PR and add some more defensive coding to handle this situation. I am unsure how you'd ever get undefined unless we can pass it from a SPA

jamieomaguire: Tbh I'm not quite sure what we'd want to occur in that situation. I think passing undefined up and letting it error is likely the preferred behaviour rather than the input trying to self-correct

raoufswe: maybe we can add an early return if it is intended?

xander-marjoram (approved):

PR #1217, opened 2024-01-25
fix(pie-input): DSW-0 required input value

jamieomaguire:
---
---

[Added] - Make the `value` property required and default to an empty string

### DIFF
--- a/.changeset/unlucky-oranges-divide.md
+++ b/.changeset/unlucky-oranges-divide.md
@@ -2,4 +2,4 @@
 "@justeattakeaway/pie-input": minor
 ---
 
-[Added] - Guard against `undefined` being passed as a value to the `value` prop and write a supporting browser test
+[Added] - Make the `value` property required and default to an empty string
--- a/.changeset/unlucky-oranges-divide.md
+++ b/.changeset/unlucky-oranges-divide.md
@@ -1,0 +1,5 @@
+---
+"@justeattakeaway/pie-input": minor
+---
+
+[Added] - Guard against `undefined` being passed as a value to the `value` prop and write a supporting browser test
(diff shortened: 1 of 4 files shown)


packages/components/pie-input/test/component/pie-input.spec.ts
```diff
@@ -121,6 +121,21 @@ test.describe('PieInput - Component tests', () => {
...
+                    props: {
+                        value: undefined,
+                    } as InputProps,
+                });
+
+                // Act
+                const input = component.locator('input');
+
+                // Assert
+                expect((await input.inputValue())).toBe('');
```
xander-marjoram, line 136: Out of interest would something along the lines of `page.inputValue('pie-input')` work, or do you need to find the inner input?

xander-marjoram (dismissed):

packages/components/pie-input/test/component/pie-input.spec.ts
jamieomaguire: It's a nice suggestion. Unfortunately, Playwright throws on any element that isn't an input :

"Throws for non-input elements. However, if the element is inside the <label> element that has an associated control, returns the value of the control."

I would prefer to use an approach that doesn't check the underlying input though. I am unsure how we can do that. I considered something like:

`const inputValue = await page.evaluate(() => document.querySelector('pie-input')?.value);`

But it's cumbersome and didn't quite work either

fernandofranca (dismissed):

ashleynolan (approved):

.changeset/unlucky-oranges-divide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@justeattakeaway/pie-input": minor
+---
+
+[Added] - Make the `value` property required and default to an empty string
```
xander-marjoram, line 5: If it's required, do we need a default?

jamieomaguire: Screenshot 2024-01-25 at 17 25 29

fernandofranca (approved):

PR #1221, opened 2024-01-26
refactor(pie-webc-testing): DSW-1676 added exportable RTL function

JoshuaNg2332:

- Added exportable RTL function
- Updated `pie-switch`'s visual test to use new function

### Reviewer 1

### Reviewer 2

### DIFF
--- a/packages/components/pie-switch/test/visual/pie-switch.spec.ts
+++ b/packages/components/pie-switch/test/visual/pie-switch.spec.ts
@@ -1,6 +1,7 @@
 import { test } from '@sand4rt/experimental-ct-web';
 import percySnapshot from '@percy/playwright';
 import { percyWidths } from '@justeattakeaway/pie-webc-testing/src/percy/breakpoints.ts';
+import { setRTL } from '@justeattakeaway/pie-webc-testing/src/helpers/set-rtl-direction.ts';
 import { PieSwitch } from '../../src/index.ts';
 import { SwitchProps, labelPlacements } from '../../src/defs.ts';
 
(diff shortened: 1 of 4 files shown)


packages/components/pie-switch/test/visual/pie-switch.spec.ts
```diff
@@ -41,10 +42,8 @@ test.describe('Prop: `Label`', () => {
 
 test.describe('when in RTL settings', () => {
     [true, false].forEach(async (checkedState) => {
-        test(`should render in rtl correctly when (checked: ${checkedState})`, async ({ page, mount }) => {
-            await page.evaluate(() => {
-                document.documentElement.setAttribute('dir', 'rtl');
-            });
+        test.only(`should render in rtl correctly when (checked: ${checkedState})`, async ({ page, mount }) => {
```
xander-marjoram, line 45: Did you mean to leave the `.only` in?

packages/components/pie-webc-testing/src/helpers/set-rtl-direction.ts
```diff
@@ -0,0 +1,7 @@
+import { Page } from '@playwright/test';
```
xander-marjoram, line 1: I don't think it makes a difference in test code, but just as a good habit you should be able to import this as a type rather than a value:
```suggestion
import { type Page } from '@playwright/test';
```

packages/components/pie-webc-testing/src/helpers/set-rtl-direction.ts
```diff
@@ -0,0 +1,7 @@
...
+
+export const setRTL = async (page: Page) => {
+    await page.evaluate(() => {
+        document.documentElement.setAttribute('dir', 'rtl');
```
xander-marjoram, line 5: I'm sure you've checked this, but just checking that any subsequent tests will be back to LTR by default?

packages/components/pie-switch/test/visual/pie-switch.spec.ts
JoshuaNg2332: Ooops. Yeah I'll remove that. Forgot about that.

packages/components/pie-webc-testing/src/helpers/set-rtl-direction.ts
JoshuaNg2332: It should tear down after the end of the test. I've just tested this and that's what happens so it does go back to LTR.

ashleynolan (approved):

raoufswe (approved):

kevinrodrigues (approved):
🥇

PR #1239, opened 2024-02-01
feat(pie-cookie-banner): DSW-1702 follow up pr

kevinrodrigues:
- [x] have performed a self-review of my code


packages/components/pie-cookie-banner/src/defs.ts
```diff
@@ -63,7 +63,13 @@ export interface CookieBannerProps {
...
     cookieTechnologiesLink: string;
 
-    defaultPreferences: Record<PreferenceIds, boolean> | object
+    /**
+     * Allows consumers to pass in specific preference(s) to the component which will toggle
+     * the switch to be on by default (if set to `true`).
+     *
+     * e.g. { 'functional': true }
+     */
+    defaultPreferences: Record<PreferenceIds, boolean> | NonNullable<unknown>;
```
jamieomaguire, line 72: Can you please test in Aperture if Typescript correctly picks up the valid strings we can pass in? I noticed when testing yesterday it seemed to let me put anything in there but I think it should be telling me I'm not allowed if I did something like `{ foo: 'bar' }`

jamieomaguire: Also out of interest what is `NonNullable<unknown>;` doing?

packages/components/pie-cookie-banner/test/component/pie-cookie-banner.spec.ts
```diff
@@ -485,4 +485,79 @@ test.describe('PieCookieBanner - Component tests', () => {
...
+                        },
+                    } as CookieBannerProps,
+                });
+
+                // Act
+                await page.click(managePreferencesSelector);
+
+                const functional = await page.locator(getPreferenceItemSelector('functional')).isChecked();
+                const personalized = await page.locator(getPreferenceItemSelector('personalized')).isChecked();
+                const analytical = await page.locator(getPreferenceItemSelector('analytical')).isChecked();
```
jamieomaguire, line 507:
```suggestion
                const functionalEnabled = await page.locator(getPreferenceItemSelector('functional')).isChecked();
                const personalizedEnabled = await page.locator(getPreferenceItemSelector('personalized')).isChecked();
                const analyticalEnabled = await page.locator(getPreferenceItemSelector('analytical')).isChecked();
```

jamieomaguire: (or something similar)

packages/components/pie-cookie-banner/test/component/pie-cookie-banner.spec.ts
```diff
@@ -485,4 +485,79 @@ test.describe('PieCookieBanner - Component tests', () => {
...
+        test.describe('when defaultPreferences are partially provided', () => {
+            test('should toggle the position to `on` for property', async ({ mount, page }) => {
+                await mount(PieCookieBanner, { props: { defaultPreferences: { functional: true } } as CookieBannerProps });
+
+                // Act
+                await page.click(managePreferencesSelector);
+                const output = await page.locator(getPreferenceItemSelector('functional')).isChecked();
+
+                // Assert
+                expect(output).toBe(true);
```
jamieomaguire, line 544: could you please assert that the other 2 are not checked?

packages/components/pie-cookie-banner/src/defs.ts
jamieomaguire, line 72: I belive this prop (along with a few others) needs to be optional as well. Otherwise there are issues with Typescript React consumers

jamieomaguire:
I noticed the following TS errors in VSCode (unsure why they aren't found at build) which might be worth trying to fix:

Screenshot 2024-02-01 at 17 39 43

Screenshot 2024-02-01 at 17 40 58

packages/components/pie-cookie-banner/src/defs.ts
jamieomaguire: Just wondering if something like `Partial<Record<PreferenceIds, boolean>>;` might be cleaner/achieve the same thing. Let me know what you think!

.changeset/fresh-chairs-behave.md
```diff
@@ -0,0 +1,5 @@
+---
+"@justeattakeaway/pie-cookie-banner": minor
+---
+
+[Added] - Follow up change requests from previous pr
```
jamieomaguire, line 5: I would suggest writing something that describes the actual changes being made rather than alluding to PR change requests. Otherwise it makes it very difficult to understand what was changed when we are reviewing the changelog (especially for consumers of our packages)

jamieomaguire: (and to nitpick, I think it could be `Changed` rather than `Added`)

kevinrodrigues: I think it should be added as i have added tests, a new prop to the readme etc. No actual functional changes: https://keepachangelog.com/en/1.1.0/

packages/components/pie-cookie-banner/src/defs.ts
kevinrodrigues: The nonNullable removes null and undefined. Ill try the other suggestion tho. For the point about {foo:bar} are you suggesting we validate it? Our validation helper wont work in this situation. I wanted to split this prop back out into 3 values that are passed as individual props instead of a single prop then that way we can validate it properly so either we do that or we leave this as it is and add individual validation for this.

apps/pie-storybook/stories/pie-cookie-banner.stories.ts
```diff
@@ -60,13 +60,15 @@ const BaseStoryTemplate = (props: CookieBannerProps) : TemplateResult => {
         cookieTechnologiesLink,
     } = props;
 
+    const defaultPreferences = { functional: true, personalized: true, analytical: true };
```
xander-marjoram, line 63: Is there a TS type to describe the preferences?

packages/components/pie-cookie-banner/test/component/pie-cookie-banner.spec.ts
siggerzz: You could probably remove the `isChecked` here, and instead, modify your assertion like so:

https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-be-checked

.changeset/fresh-chairs-behave.md
jamieomaguire: Perhaps no changeset file should be added in that case? Because the current message doesn't indicate no functional changes were added

packages/components/pie-cookie-banner/src/defs.ts
jamieomaguire: I do not think we should validate it. My understanding is these Typings do not prevent incorrect values being passed at run time, just help during development. 

When you are working with this component in an IDE it should prevent you from adding { foo: 'bar' } to this prop, which is what I was asking about :)

I don't _think_ nonnullable is needed here but I think something like `Partial<Record<PreferenceIds, boolean>>;` could be nice

kevinrodrigues: @xander-marjoram so the prop is optional. Not forced to use. They can also pass in partial values so instead of the full set 3 keys they could pass in 2 or 1 key instead.

xander-marjoram: @kevinrodrigues Optional makes sense to me, as long as we document the _default_ defaults well!

I think it could get confusing when a partial object is given, as then there could be values, default values and default default values all at the same time.

fernandofranca (dismissed):

packages/components/pie-cookie-banner/test/component/pie-cookie-banner.spec.ts
```diff
@@ -213,10 +212,10 @@ test.describe('PieCookieBanner - Component tests', () => {
                 await page.click(getPreferenceItemSelector(preference.id));
             }
         }
-        const isToggleAllChecked = await page.locator(getPreferenceItemSelector('all')).isChecked();
+        const isToggleAllChecked = await page.locator(getPreferenceItemSelector('all'));
```
jamieomaguire, line 215: this name suggests a boolean value but it's a locator result which is a little misleading - could you rename it to be clearer please?

packages/components/pie-cookie-banner/test/component/pie-cookie-banner.spec.ts
```diff
@@ -247,10 +246,10 @@ test.describe('PieCookieBanner - Component tests', () => {
         await page.click(managePreferencesSelector);
         await page.click(getPreferenceItemSelector('all')); // turn on all nodes
         await page.click(getPreferenceItemSelector('functional')); // turn off one of the preferences
-        const isToggleAllChecked = await page.locator(getPreferenceItemSelector('all')).isChecked();
+        const isToggleAllChecked = await page.locator(getPreferenceItemSelector('all'));
```
jamieomaguire, line 249: same issue here regarding the variable name sounding like it's a boolean

packages/components/pie-cookie-banner/test/component/pie-cookie-banner.spec.ts
```diff
@@ -485,4 +484,93 @@ test.describe('PieCookieBanner - Component tests', () => {
             });
         });
     });
+
+    test.describe('`defaultPreferences` : prop', () => {
+        test.describe('when defaultPreferences are set to three values', () => {
+            test('should toggle the position to `on` for properties', async ({ mount, page }) => {
```
jamieomaguire, line 490:
```suggestion
            test('should toggle the position to `on` for all 3 properties', async ({ mount, page }) => {
```

packages/components/pie-cookie-banner/src/defs.ts
```diff
@@ -63,7 +63,15 @@ export interface CookieBannerProps {
...
     cookieTechnologiesLink: string;
 
-    defaultPreferences: Record<PreferenceIds, boolean> | object
+    /**
+     * Allows consumers to pass in specific preference(s) to the component which will toggle
+     * the switch to be on by default (if set to `true`).
+     *
+     * e.g. { 'functional': true }
+     */
+    defaultPreferences?: Partial<Record<PreferenceIds, boolean>> & {
```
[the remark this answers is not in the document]
jamieomaguire: I am sure it is fine, do you mind explaining how this solves that for my own understanding please?

packages/components/pie-cookie-banner/test/component/pie-cookie-banner.spec.ts
kevinrodrigues: Fixed.

packages/components/pie-cookie-banner/test/component/pie-cookie-banner.spec.ts
kevinrodrigues: Fixed.

packages/components/pie-cookie-banner/src/defs.ts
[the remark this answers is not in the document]
kevinrodrigues: Are you able to suggest an alternative approach please as i tried quite a few things and this was the only one that actually worked. I'd be happy to update it if there is a better way tbh.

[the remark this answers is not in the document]
kevinrodrigues: Fixed. Thanks for pairing on that one!

jamieomaguire (approved):

raoufswe (approved):

PR #1246, opened 2024-02-05
feat(pie-input): DSW-1583 add new disabled prop to pie-input

xander-marjoram:

### Reviewer 1
No docs or visual regression changes.

### Reviewer 2
No docs or visual regression changes.


packages/components/pie-input/src/index.ts
```diff
@@ -73,6 +76,16 @@ export class PieInput extends FormControlMixin(RtlMixin(LitElement)) implements
         return (this.input as HTMLInputElement).validity;
     }
 
+    /**
+     * Called after the disabled state of the element changes,
+     * either because the disabled attribute of this element was added or removed;
+     * or because the disabled state changed on a <fieldset> that's an ancestor of this element.
+     * @param disabled - The latest disabled state of the input.
+     */
+    public formDisabledCallback (disabled: boolean): void {
```
jamieomaguire, line 85: 🔥

packages/components/pie-input/src/index.ts
```diff
@@ -35,6 +35,9 @@ export class PieInput extends FormControlMixin(RtlMixin(LitElement)) implements
     @property({ type: String })
     public name?: InputProps['name'];
 
+    @property({ type: Boolean })
```
jamieomaguire, line 38: It may be worth reflecting this property. I believe we have done for the same property in other components and I think especially because it can be changed indirectly via fieldsets etc it would be worth doing. Let me know what you think!

packages/components/pie-input/test/component/pie-input.spec.ts
```diff
@@ -689,7 +742,7 @@ test.describe('PieInput - Component tests', () => {
             await page.setContent(`
                 <form id="testForm" action="/foo" method="POST">
                     <pie-input type="text" name="username"></pie-input>
-                    <button type="reset">Submit</button>
+                    <button type="reset">Reset</button>
```
jamieomaguire, line 745: 🙇

packages/components/pie-input/test/component/pie-input.spec.ts
```diff
@@ -700,5 +753,27 @@ test.describe('PieInput - Component tests', () => {
...
+
+            await setupFormDataExtraction(page, '#testForm', '#formDataJson');
+
+            // Act
+            await page.click('button[type="submit"]');
+
+            const formDataObj = await getFormDataObject(page, '#formDataJson');
+
+            // Assert
+            expect(formDataObj).toStrictEqual({ email: '' });
```
jamieomaguire, line 776: entirely up to you, but I wonder if it is worth adding a value to the disabled input to verify that does not submit either

.changeset/late-feet-smash.md
```diff
@@ -0,0 +1,6 @@
+---
+"@justeattakeaway/pie-input": minor
+"pie-storybook": patch
+---
+
+[Added] - disabled prop
```
jamieomaguire, line 6: killer changeset name 😂

jamieomaguire (approved):

jamieomaguire (approved):

.changeset/late-feet-smash.md
siggerzz: 🦶🏼

siggerzz (approved):

fernandofranca (dismissed):

packages/components/pie-input/test/component/pie-input.spec.ts
```diff
@@ -490,6 +490,59 @@ test.describe('PieInput - Component tests', () => {
...
             });
         });
+
+        test.describe('disabled', () => {
+            test('should not disable the component if the prop is not provided', async ({ mount }) => {
+                // Arrange
+                const component = await mount(PieInput, {});
+
+                // Act
+                const input = await component.locator('input');
```
kevinrodrigues, line 500: I'm not sure you need the awaits here. You may need it on the expect lines but for the locator itself i dont think it returns a promise.

packages/components/pie-input/test/component/pie-input.spec.ts
```diff
@@ -760,5 +813,52 @@ test.describe('PieInput - Component tests', () => {
...
+            `);
+
+            await setupFormDataExtraction(page, '#testForm', '#formDataJson');
+
+            // Act
+            await page.click('button[type="submit"]');
+
+            const formDataObj = await getFormDataObject(page, '#formDataJson');
+
+            expect(formDataObj).toStrictEqual({
```
kevinrodrigues, line 859: Minor, `// Assert`

kevinrodrigues (approved):

packages/components/pie-input/test/component/pie-input.spec.ts
```diff
@@ -490,6 +490,59 @@ test.describe('PieInput - Component tests', () => {
...
+
+                // Act
+                const input = await page.locator('pie-input');
+                await input.focus();
+
+                // Assert
+                expect(input).not.toBeFocused();
+            });
+
+            test('should still be able to focus the component when not disabled', async ({ page }) => {
```
kevinrodrigues, line 534: Should this test be here because it falls under the describe of `disabled`?

xander-marjoram: I see it as a variation of the value for the `disabled` prop, so I think it does belong here 👍🏻

packages/components/pie-input/test/component/pie-input.spec.ts
xander-marjoram: Good shout. There is indeed a (very subtle) IDE warning about this which I completely missed. I'll resolve all instances of that in this file.

packages/components/pie-input/test/component/pie-input.spec.ts
kevinrodrigues: OK. How about the below (appreciate it might be a nit picky so can keep as is, i just think it would read better/clearer imo with the below).

```javascript
describe('`disabled`')
  // other tests
   describe('when not provided')
      test('should still be able to focus the component')
```

packages/components/pie-input/test/component/pie-input.spec.ts
jamieomaguire: What's strange is the Playwright Docs all use await in their locator examples, which is probably how it found it's way into the code base

https://playwright.dev/docs/locators

jamieomaguire (approved):

kevinrodrigues (approved):

PR #1248, opened 2024-02-05
feat(pie-chip): DSW-1643 add the chip component

raoufswe:
- adds the basic functionality of the chip component including variants, selected and disabled, loading, and dismissible states
- events are out of the scope of this PR #:handshake: 

### Reviewer 1

### Reviewer 2


packages/components/pie-chip/src/chip.scss
```diff
@@ -1 +1,147 @@
...
+            --chip-bg-color: hsl(var(#{$bg-color}-h), var(#{$bg-color}-s), calc(var(#{$bg-color}-l) + var(--active-modifier)));
+        }
+    }
+}
+
+*,
+*:before,
+*:after {
+    box-sizing: border-box;
+}
```
jamieomaguire, line 40: Does pie-css not remove the need for this?

packages/components/pie-chip/src/chip.scss
```diff
@@ -1 +1,147 @@
...
+            margin-inline-start: var(--chip-dismissible-offset);
+        }
+    }
+
+    &[isLoading] {
+        & > *:not(pie-spinner) {
+            visibility: hidden;
+        }
+
+        pie-spinner {
```
jamieomaguire, line 126: I think you could add a class to the spinner inside the chip template instead of using the selector - please could you give it a go and update  if it works?

packages/components/pie-chip/test/component/pie-chip.spec.ts
```diff
@@ -4,11 +4,21 @@ import { PieChip, ChipProps } from '../../src/index.ts';
 
 const componentSelector = '[data-test-id="pie-chip"]';
 
+const props: Partial<ChipProps> = {
```
jamieomaguire, line 7: 🔥

jamieomaguire:
Excellent stuff ❤️ just a couple of comments!

packages/components/pie-chip/src/chip.scss
raoufswe: let me check

raoufswe: looks like it isn't needed. 

I also created a ticket [DSW-1730] to update other components. 🤝

packages/components/pie-chip/src/chip.scss
raoufswe: updated 💯

kevinrodrigues (approved):

packages/components/pie-chip/src/index.ts
```diff
@@ -11,10 +15,82 @@ const componentSelector = 'pie-chip';
...
  * @tagname pie-chip
+ * @slot icon - The icon slot
+ * @slot - Default slot
  */
 export class PieChip extends LitElement implements ChipProps {
+    @property()
+    @validPropertyValues(componentSelector, variants, 'primary')
+    public variant: ChipProps['variant'] = 'default';
+
+    @property({ type: Boolean })
```
jamieomaguire, line 26: in other components, we have reflected the disabled prop. I don't know if it's needed as the state cannot be changed internally in this case but it may be worth considering (if you haven't already)

packages/components/pie-chip/src/index.ts
```diff
@@ -11,10 +15,82 @@ const componentSelector = 'pie-chip';
...
+    public disabled = false;
+
+    @property({ type: Boolean })
+    public isSelected = false;
```
jamieomaguire, line 30: Do these all need to be marked as optional `?` to enforce the interface types? If not we can probably clean up all our other components 😂

packages/components/pie-chip/src/index.ts
raoufswe: that's an interesting question. so the chip element isn't a native button element, instead, a div with the role of `button`. referencing [mdn](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-disabled),  div elements don't expect the disabled attribute - even if it were to be given a role="button".

on a side note, we still need to specify the `aria-disabled` attribute  for the disabled state but aria labels will be addressed in another ticket.

packages/components/pie-chip/src/index.ts
raoufswe: imo, I don't think we need to mark it as optional since we've assigned a default value and the prop will always have/need a value within the class implementation. the prop is indeed optional for consumers, but not for the class itself. That's different for things where the class expects it to be optional/ undefined such as `href`

I think the key here is to realize that what's optional for consumers doesn't necessarily correspond to the class contract, that's why the other day I suggested the idea of having a contract/interface for consumers and one for the class implementation but not sure how to procced with that xD

packages/components/pie-chip/src/index.ts
jamieomaguire: Sounds good - definitely sounds like a component we will want to properly test on screenreaders!

jamieomaguire (approved):

PR #1249, opened 2024-02-05
feat(pie-monorepo): DSW-000 added danger check for pr checkboxes

ashleynolan:
## Describe your changes
- Added Danger fail for future PRs to stop them being merged before the PR checklists are complete/up-to-date

## Author Checklist

### DIFF
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -1,6 +1,7 @@
 /* global danger, fail, message */
 import { message, danger } from 'danger';
 
+const { pr } = danger.github;
 const validChangesetCategories = ['Added', 'Changed', 'Removed', 'Fixed'];
 
 // Check for correct Changeset formatting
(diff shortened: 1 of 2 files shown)


dangerfile.js
```diff
@@ -32,3 +33,8 @@ danger.git.created_files.filter((filepath) => filepath.includes('.changeset/') &
             console.log(err);
         });
     });
+
+// Check for empty PR Description checkboxes
+if (pr.body.includes('- [ ]')) {
```
raoufswe, line 38: image

raoufswe (approved):

dangerfile.js
kevinrodrigues, line 38: Do you need to specifically check for the `x` here as i think you could enter anything and this would pass right?

kevinrodrigues (approved):

PR #1251, opened 2024-02-06
feat(pie-assistive-text): DSW-1590 remaining code for assistive-text component

LTurns:

[Added] - Logic and styles for assistive-text component
[Added] - `sanitizeAndRenderHTML` utility for slots in button, link, form-label, tag and assistive-text stories
[Added] - Additional content in [generator-pie-component] README on running visual tests

- [x] If it is a `PIE Docs` change, I have reviewed the Docs site preview

### Reviewer 1

### Reviewer 2


packages/components/pie-assistive-text/src/index.ts
```diff
@@ -13,8 +19,36 @@ const componentSelector = 'pie-assistive-text';
...
+    public variant?: AssistiveTextProps['variant'] = 'default';
+
+    /**
+ * Renders the assistive-text icon content.
+ * @private
+ */
+    private renderContent (): TemplateResult {
+        const { variant } = this;
+        return html`
+            ${variant === 'success' ? html`<icon-check-circle size="s" />` : nothing}
```
jamieomaguire, line 33: out of interest can we apply classes to these icon elements? i.e. `.c-assistiveText-icon`

If so, this might be nicer for styling instead of relying on the internal class used inside the icon itself!

packages/components/pie-assistive-text/src/index.ts
```diff
@@ -13,8 +19,36 @@ const componentSelector = 'pie-assistive-text';
...
-        return html`<h1 data-test-id="pie-assistive-text">Hello world!</h1>`;
+        const {
+            variant,
+        } = this;
+
+        return html`
+        <p 
+            class="c-assistiveText"
+            data-test-id="pie-assistive-text"
+            variant=${variant || 'default'}>
```
jamieomaguire, line 46: would `ifDefined(variant)` work here? I'm not super keen on repeating the default string in the template (or you could use a default values object like in `pie-input`?)

Realistically the chances of it actually being undefined are very low, even in a JS consuming application

packages/components/pie-assistive-text/src/assistive-text.scss
```diff
@@ -1 +1,29 @@
...
+
+    &[variant='error'] {
+        --assistive-text-color: var(--dt-color-content-error);
+    }
+}
+
+
+.c-assistiveText-content {
+    gap: var(--dt-spacing-b);
+    display: flex;
```
jamieomaguire, line 23: there's quite a lot of flex going on here, just to double check is it definitely needed?

packages/components/pie-assistive-text/src/index.ts
LTurns: just tested and `ifDefined` does indeed work here! I agree it's nicer than duplicating the default string, changing now 😄

packages/components/pie-assistive-text/src/assistive-text.scss
LTurns: thank you! just double checked and I don't need the ones at the top anymore 😄 removed 👍

packages/components/pie-assistive-text/src/assistive-text.scss
```diff
@@ -1 +1,27 @@
...
+        --assistive-text-color: var(--dt-color-content-positive);
+    }
+
+    &[variant='error'] {
+        --assistive-text-color: var(--dt-color-content-error);
+    }
+}
+
+
+.c-assistiveText-content {
```
jamieomaguire, line 19: Screenshot 2024-02-06 at 13 20 26

Looks great! Just want to double check, do we want to remove the default margin applied to the `p` tag?

jamieomaguire: Screenshot 2024-02-06 at 13 21 29

Also, are you happy that it meets the whitespace requirements outlined in the design? I only ask because I have trouble understanding them sometimes :D

packages/components/pie-assistive-text/src/assistive-text.scss
```diff
@@ -1 +1,27 @@
...
+    gap: var(--dt-spacing-b);
+    display: flex;
+    align-items: center;
```
jamieomaguire, line 22: I believe the font size in the designs is 14px, can you update please?

packages/components/pie-assistive-text/src/index.ts
```diff
@@ -13,8 +20,36 @@ const componentSelector = 'pie-assistive-text';
  * @tagname pie-assistive-text
  */
```
jamieomaguire, line 21: I think you could add a `* @slot - Default slot ` comment above the class to assist with the `custom-elements.json` file that is generated at build

jamieomaguire (dismissed):

jamieomaguire (dismissed):

apps/pie-storybook/stories/pie-assistive-text.stories.ts
```diff
@@ -2,35 +2,55 @@ import { html } from 'lit';
...
+    argTypes: {
+        variant: {
+            description: 'Set the variant of the assistive text.',
+            control: 'select',
+            options: variants,
+            defaultValue: {
+                summary: 'default',
+            },
+        },
+    },
```
raoufswe, line 30: I think you will need to add arg type for the slot as well so users can type what they want. [here](https://github.com/justeattakeaway/pie/pull/1248/files#diff-35c71a2a7b57dcd5e06095f70907a848a84003a2900089d86b7ca38695bd0687R73) is an exmaple. you will also need to update the template a bit

LTurns: I don't know if that's needed actually, I'm able to type anything in the slot on the deployed storybook already 🎉 
Screenshot 2024-02-07 at 08 37 35

packages/components/pie-assistive-text/src/index.ts
```diff
@@ -11,10 +18,39 @@ const componentSelector = 'pie-assistive-text';
...
+        const {
+            variant,
+        } = this;
+
+        return html`
+        <p 
+            class="c-assistiveText"
+            data-test-id="pie-assistive-text"
+            variant=${ifDefined(variant)}>
+            <span class="c-assistiveText-content">
```
raoufswe, line 49: i might be missing something but do you think we can update the markup to be like

```
<p
            class="c-assistiveText"
            data-test-id="pie-assistive-text"
            variant=${ifDefined(variant)}>
             ${this.renderContent()}
            <slot></slot>
             </p>

```

then move the styles of "c-assistiveText-content" to the parent. this way we can use only one element for the component and encapsulate things in one place. 

also in case it helps, you can also set the size of the icon and its display value using the following css vars. 

```
    --icon-display-override: block;
    --icon-size-override: 20px;
    --icon-size-override: 20px;
```

raoufswe (commented):
well done lizzie, just two things and should be good to go! 💯

apps/pie-storybook/stories/pie-assistive-text.stories.ts
raoufswe: oh cool, then the main reason we need it, so we can [sanitize](https://github.com/justeattakeaway/pie/pull/1248/files#diff-35c71a2a7b57dcd5e06095f70907a848a84003a2900089d86b7ca38695bd0687R105) the slot value passed by the user.

packages/components/pie-assistive-text/src/index.ts
LTurns: thanks @raoufswe I'll look into this! ❤️ I personally prefer using the `size="s"` prop in `pie-icons-webc` than declaring the icon size variables, but I'll try with the override and removing the span 🤞

apps/pie-storybook/stories/pie-assistive-text.stories.ts
jamieomaguire: good spot @raoufswe!

LTurns: ahh okay, I copied this from tag so will change there too in this PR! 👍

packages/components/pie-assistive-text/src/index.ts
LTurns: works without the span 🎉 thank you!

jamieomaguire (approved):

raoufswe (dismissed):

jamieomaguire (approved):

jamieomaguire (approved):

raoufswe (approved):

PR #1255, opened 2024-02-07
feat(pie-input): DSW-1588 add leading and trailing content slots

jamieomaguire:

[Added] - Leading and Trailing content slots to pie-input component

- [X] If it is a core feature, I have added thorough tests
- [X] If it is a component change, I have reviewed the Storybook preview

### Reviewer 1

### Reviewer 2

### DIFF
--- a/apps/pie-storybook/stories/pie-input.stories.ts
+++ b/apps/pie-storybook/stories/pie-input.stories.ts
@@ -5,7 +5,7 @@ import { useArgs as UseArgs } from '@storybook/preview-api';
 
 /* eslint-disable import/no-duplicates */
 import '@justeattakeaway/pie-input';
-import { types, inputModes, InputProps as InputPropsOriginal } from '@justeattakeaway/pie-input';
+import { types, inputModes, InputProps as InputPropsBase } from '@justeattakeaway/pie-input';
 /* eslint-enable import/no-duplicates */
 
 import { type StoryMeta } from '../types';
@@ -14,7 +14,7 @@ import '@justeattakeaway/pie-button';
 import '@justeattakeaway/pie-icons-webc/IconPlaceholder';
 
 // Extending the props type definition to include storybook specific properties for controls
-type InputProps = InputPropsOriginal & {
+type InputProps = InputPropsBase & {
     leadingSlot: typeof slotOptions[number];
     trailingSlot: typeof slotOptions[number];
 };
--- a/apps/pie-storybook/stories/pie-input.stories.ts
+++ b/apps/pie-storybook/stories/pie-input.stories.ts
@@ -128,20 +128,14 @@ const inputStoryMeta: InputStoryMeta = {
             },
         },
         leadingSlot: {
-            description: 'An icon or short text to display at the start of the input.',
+            description: 'An icon or short text to display at the start of the input. <br><b>*Not a component Prop</b>',
             control: 'select',
             options: slotOptions,
-            defaultValue: {
-                summary: 'text',
-            },
         },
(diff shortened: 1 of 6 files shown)


siggerzz (approved):

packages/components/pie-input/src/index.ts
```diff
@@ -122,21 +124,26 @@ export class PieInput extends FormControlMixin(RtlMixin(LitElement)) implements
...
-            autocomplete=${ifDefined(autocomplete)}
-            ?autofocus=${autoFocus}
-            inputmode=${ifDefined(inputmode)}
-            placeholder=${ifDefined(placeholder)}
-            ?readonly=${readonly}
-            @input=${this.handleInput}
-            @change=${this.handleChange}
-            data-test-id="pie-input">`;
+        return html`
+            <div>
```
dandel10n, line 128: Interesting if it is possible to use empty tags instead of actual html elements here similar like it is possible in react `<></>`

dandel10n (dismissed):

apps/pie-storybook/stories/pie-input.stories.ts
```diff
@@ -1,16 +1,23 @@
...
+import { types, inputModes, InputProps as InputPropsOriginal } from '@justeattakeaway/pie-input';
 /* eslint-enable import/no-duplicates */
 
 import { type StoryMeta } from '../types';
 import { createStory, type TemplateFunction } from '../utilities';
 import '@justeattakeaway/pie-button';
+import '@justeattakeaway/pie-icons-webc/IconPlaceholder';
+
+// Extending the props type definition to include storybook specific properties for controls
+type InputProps = InputPropsOriginal & {
```
raoufswe, line 17: nit: i think we have used the `Base` convention before in the codebase
so maybe we can have something like `InputBaseProps`

packages/components/pie-input/src/index.ts
raoufswe: @dandel10n I think that's only applicable for JSX, and not part of the html standards

jamieomaguire: Even if it were possible, I need a wrapper element for the styling later on 😃

siggerzz (approved):

dandel10n (approved):

raoufswe (approved):

apps/pie-storybook/stories/pie-input.stories.ts
```diff
@@ -147,22 +181,37 @@ const Template = ({
         });
     }
 
+    function renderLeadingOrTrailingSlot (slotName: 'leading' | 'trailing', slotValue: typeof slotOptions[number]) {
+        if (slotValue === slotOptions[0]) {
```
kevinrodrigues, line 185: You could place these inside of an object instead of the `if` statement and assign the html markup to each key (there would only be two) then at the end return the slot required or nothing. Would simplify this method imo.

e.g

slot = {
[slotOne]: html`<markup>`
[slotTwo]: html`<markup>`
}

reutnr slot[slotValue] || nothing;

packages/components/pie-input/test/component/pie-input.spec.ts
```diff
@@ -633,6 +637,62 @@ test.describe('PieInput - Component tests', () => {
         });
     });
 
+    test.describe('Slots', () => {
```
kevinrodrigues, line 640: I think there should there be a test to check if the slot renders nothing when nothing is passed in? i.e slot: '' or no slot prop at all.

PR #1257, opened 2024-02-09
feat(pie-form-label): DSW-1585 update pie-form-label to work with form components

fernandofranca:
### @justeattakeaway/pie-switch & @justeattakeaway/pie-input

[Added] - `focusTarget` to enable focusing from a pie-form-label

### @justeattakeaway/pie-form-label

[Added] - focusing logic using `focusTarget`

### @justeattakeaway/pie-webc-core

[Added] - PIEInputElement TS interface

## Notes

- the `focusTarget` prop was added to point to the exact element that should receive focus in the component's shadow DOM and that prop must be added to new form components
- we might not need that prop it if the component DOM structure is quite simple, meaning the target element is the direct child of the web component
- pie-label integration with pie-switch and pie-input can be checked at this aperture deployment:
https://pr74-aperture-nextjs-app.pie.design/label-integration


packages/components/pie-form-label/package.json
```diff
@@ -33,6 +33,8 @@
   "devDependencies": {
     "@custom-elements-manifest/analyzer": "0.9.0",
     "@justeattakeaway/pie-components-config": "0.11.0",
+    "@justeattakeaway/pie-input": "^0.11.0",
+    "@justeattakeaway/pie-switch": "^0.26.0",
```
xander-marjoram, line 37: Apologies if I'm missing the point, but are these dependencies needed?

packages/components/pie-form-label/test/component/pie-form-label.spec.ts
```diff
@@ -24,4 +26,77 @@ test.describe('PieFormLabel - Component tests', () => {
...
+                });
+            });
+            test.describe('when clicked twice', () => {
+                test('the switch "checked" attribute is false', async ({ page }) => {
+                    // Arrange
+                    await page.setContent(markup);
+
+                    // Act
+                    const target = page.locator('#approveSettings');
+                    const label = page.locator('pie-form-label');
```
xander-marjoram, line 92: For these tests it might be good to use something like `expect.soft` before the clicks to verify the initial state

packages/components/pie-input/src/index.ts
```diff
@@ -6,7 +6,7 @@ import { ifDefined } from 'lit/directives/if-defined.js';
 import { live } from 'lit/directives/live.js';
 
 import {
-    validPropertyValues, RtlMixin, defineCustomElement, FormControlMixin, wrapNativeEvent,
+    validPropertyValues, RtlMixin, defineCustomElement, FormControlMixin, wrapNativeEvent, PIEInputElement,
```
xander-marjoram, line 9: I believe this can be a `type` import (along with InputProps and PropertyValues, while you're here 😄)
```suggestion
    validPropertyValues, RtlMixin, defineCustomElement, FormControlMixin, wrapNativeEvent, type PIEInputElement,
```

packages/components/pie-form-label/package.json
xander-marjoram: Ah I see, it's for the tests

fernandofranca: Sadly Playwright will crash without having those as deps. Perhaps there is any other way of solving this?

packages/components/pie-input/src/index.ts
fernandofranca: Good shout

packages/components/pie-form-label/test/component/pie-form-label.spec.ts
fernandofranca: You mean something like this?
`await expect.soft(cookieBannerComponent).toBeVisible();`

packages/components/pie-form-label/test/component/pie-form-label.spec.ts
```diff
@@ -24,4 +26,77 @@ test.describe('PieFormLabel - Component tests', () => {
...
+                    await label.click();
+                    await label.click();
+
+                    // Assert
+                    expect(target).toBeFocused();
```
xander-marjoram, line 97: I just noticed that this test is checking the focus state instead of the checked state, just in case you also missed it

fernandofranca: I fixed that one too. Thanks! 🙇‍♂️

xander-marjoram (dismissed):

jamieomaguire (approved):

xander-marjoram (approved):

PR #1262, opened 2024-02-09
feat(pie-webc): DSW-1657 add button and modal to component bundle

xander-marjoram:

### Reviewer 1
- No changes for individual components.

### Reviewer 2
- No changes for individual components.

### DIFF
--- a/packages/components/pie-webc/src/index.ts
+++ b/packages/components/pie-webc/src/index.ts
@@ -1,1 +1,7 @@
-// Component exports to live in this file
+import { PieButton } from '@justeattakeaway/pie-button';
+import { PieModal } from '@justeattakeaway/pie-modal';
+
+export {
+    PieButton,
+    PieModal,
+};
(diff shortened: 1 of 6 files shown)


packages/components/pie-webc/src/index.ts
```diff
@@ -1 +1,7 @@
-// Component exports to live in this file
+import { PieButton } from '@justeattakeaway/pie-button';
+import { PieModal } from '@justeattakeaway/pie-modal';
```
jamieomaguire, line 2: I'm thinking this means if I call `import '@justeattakeaway/pie-webc;'` then every customElement.define function is called automatically?

I don't necessarily see any issue with this, just curious

xander-marjoram: That's quite likely at the moment. I think this will be investigated fully in the upcoming tree-shaking tickets. It'll be interesting to know if bundlers are clever enough to separate them! If it does manage to split them out then I would imagine that only the imported components are registered.

jamieomaguire (approved):

siggerzz (approved):

PR #1272, opened 2024-02-13
feat(pie-webc-core): DSW-1732 dispatch custom event helper function

dandel10n:

---
---

[Added] - dispatchCustomEvent helper function

---
"@justeattakeaway/pie-cookie-banner": patch
"@justeattakeaway/pie-modal": patch
---

[Changed] - local dispatchCustomEvent function replaced with a helper from `pie-webc-core` package

- [x]  If it is a core feature, I have added thorough tests

### DIFF
--- a/packages/components/pie-webc-core/src/functions/dispatchCustomEvent.ts
+++ b/packages/components/pie-webc-core/src/functions/dispatchCustomEvent.ts
@@ -1,0 +1,22 @@
+/**
+ * Dispatch a custom event.
+ *
+ * To be used whenever we have custom events we want to
+ * bubble up through the component.
+ * Bubbles and composed event props are set to true.
+ * @param {Element} element The element to dispatch the event from.
+ * @param {string} eventType
+ * @param {any} detail
+ */
+export function dispatchCustomEvent (element: Element, eventType: string, detail?: CustomEventInit['detail']) : void {
+    if (!eventType.startsWith('pie-')) {
+        console.warn('A custom event name should start with `pie-`');
+    }
+    const event = new CustomEvent(eventType, {
+        bubbles: true,
+        composed: true,
+        detail,
+    });
+
+    element.dispatchEvent(event);
+}
--- a/packages/components/pie-webc-core/src/test/functions/dispatchCustomElement/dispatchCustomEvent.browser.spec.ts
+++ b/packages/components/pie-webc-core/src/test/functions/dispatchCustomElement/dispatchCustomEvent.browser.spec.ts
@@ -1,0 +1,45 @@
+import { test, expect } from '@sand4rt/experimental-ct-web';
+import { MockComponent } from './MockComponent.ts';
+
+test.describe('dispatchCustomEvent function', () => {
+    test('should dispatch an event when clicked', async ({ mount, page }) => {
+        // Arrange
+        const events : Array<Event> = [];
+
+        await mount(MockComponent, {
+            on: {
+                'pie-mock-event': (event: Event) => events.push(event),
+            },
+        });
+
+        // Act
+        await page.locator('[data-test-id="dispatch-custom-event-mock"]').click();
+
+        // Assert
+        expect(events).toHaveLength(1);
+    });
+
+    test('should call console.warn when event name do not start with "pie-"', async ({ mount, page }) => {
+        // Arrange
+        let result = '';
+        const expected = 'A custom event name should start with `pie-`';
+
+        await mount(MockComponent, {
+            props: {
+                eventName: 'mock-event',
+            },
+        });
+
+        page.on('console', (msg) => {
+            if (msg.type() === 'warning') {
+                result = msg.text();
+            }
+        });
+
+        // Act
+        await page.locator('[data-test-id="dispatch-custom-event-mock"]').click();
+
+        // Assert
+        expect(result).toMatch(expected);
+    });
+});
--- a/packages/components/pie-webc-core/vite.config.js
+++ b/packages/components/pie-webc-core/vite.config.js
@@ -18,7 +18,7 @@ export default defineConfig({
         globals: true,
         exclude: [
             '**/test/{accessibility,component,system,visual}/*.spec.{js,ts}',
-            '**/test/mixins/**/*.browser.spec.{js,ts}',
+            '**/test/**/*.browser.spec.{js,ts}',
             '**/node_modules/**/*'
         ],
     },
--- a/.changeset/itchy-days-explain.md
+++ b/.changeset/itchy-days-explain.md
@@ -1,0 +1,6 @@
+---
+"@justeattakeaway/pie-cookie-banner": patch
+"@justeattakeaway/pie-modal": patch
+---
+
+[Changed] - local dispatchCustomEvent function replaced with a helper from `pie-webc-core` package
--- a/.changeset/new-tomatoes-give.md
+++ b/.changeset/new-tomatoes-give.md
@@ -1,0 +1,5 @@
+---
+"@justeattakeaway/pie-webc-core": minor
+---
+
+[Added] - dispatchCustomEvent helper function.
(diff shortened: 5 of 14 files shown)


packages/components/pie-webc-core/src/functions/customEventDispatcher.ts
```diff
@@ -0,0 +1,18 @@
...
+ * To be used whenever we have behavioural events we want to
+ * bubble up through the component.
+ * @param {Element} element The element to dispatch the event from.
+ * @param {string} eventType
+ * @param {any} detail
+ */
+export function dispatchCustomEvent (element: Element, eventType: string, detail?: CustomEventInit['detail']) : void {
+    const event = new CustomEvent(eventType, {
+        bubbles: true,
+        composed: true,
```
jamieomaguire, line 13: I am unsure if we should assume that every custom event should have `bubbles` and `composed` set to true. 

Are we sure this is worth extracting and reusing?

packages/components/pie-webc-core/src/functions/customEventDispatcher.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * Dispatch a custom event.
+ *
+ * To be used whenever we have behavioural events we want to
+ * bubble up through the component.
+ * @param {Element} element The element to dispatch the event from.
+ * @param {string} eventType
+ * @param {any} detail
+ */
+export function dispatchCustomEvent (element: Element, eventType: string, detail?: CustomEventInit['detail']) : void {
```
jamieomaguire, line 10: we should add some unit tests for this function

packages/components/pie-webc-core/src/functions/customEventDispatcher.ts
dandel10n: Currently for all the use cases it is true. I think we will be able to upgrade it in future if there will be need for it

packages/components/pie-webc-core/src/functions/customEventDispatcher.ts
dandel10n: I added a browser test taking formControlMixin.browser.spec.ts as an example but looks like while it passes as part of browser tests in fails as part of unit tests. I am not sure what I did wrong, should I exclude it from unit tests somehow?

packages/components/pie-webc-core/vite.config.js
```diff
@@ -18,7 +18,7 @@ export default defineConfig({
         globals: true,
         exclude: [
             '**/test/{accessibility,component,system,visual}/*.spec.{js,ts}',
-            '**/test/mixins/**/*.browser.spec.{js,ts}',
+            '**/test/*.browser.spec.{js,ts}',
```
dandel10n, line 21: @siggerzz hey, I updated this path as now we have browser tests not only in mixins folder. I also removed this condition from root vite config. Do you have any objections?

packages/components/pie-webc-core/package.json
```diff
@@ -10,7 +10,7 @@
     "build": "run -T vite build",
     "lint:scripts": "run -T eslint .",
     "lint:scripts:fix": "run -T eslint . --fix",
-    "test": "run -T vitest run --config ../../../vite.config.js",
+    "test": "run -T vitest run --config ./vite.config.js",
```
dandel10n, line 13: @siggerzz I also pointed to the pie-webc-core vite config here. I don't know why it was pointing to the core one. Was there a specific reason?

packages/components/pie-webc-core/src/functions/customEventDispatcher.ts
dandel10n: All fixed, thanks for your help @jamieomaguire 
For the ci tests to pass I had to leave link to the root vite config in webc-core package.json but updated it to include not only mixin folder

raoufswe (dismissed):

packages/components/pie-webc-core/vite.config.js
siggerzz: That's fine with me! 🎉

siggerzz (approved):

packages/components/pie-webc-core/src/functions/dispatchCustomEvent.ts
```diff
@@ -0,0 +1,18 @@
...
+ * bubble up through the component.
+ * @param {Element} element The element to dispatch the event from.
+ * @param {string} eventType
+ * @param {any} detail
+ */
+export function dispatchCustomEvent (element: Element, eventType: string, detail?: CustomEventInit['detail']) : void {
+    const event = new CustomEvent(eventType, {
+        bubbles: true,
+        composed: true,
+        detail,
```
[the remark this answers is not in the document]
jamieomaguire: [For reference](https://github.com/justeattakeaway/pie/tree/main/packages/components#events)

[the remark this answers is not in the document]
dandel10n: I commented on your initial review about bubbles and composed [here](https://github.com/justeattakeaway/pie/pull/1272#discussion_r1489077798). I think that as we are starting small and all the current use cases need both key values to be true we can leave it as is. If in future there will be the need to set either of them to false, the helper can be modified leaving the default as true to not break existing use cases. What do you think?

Good point about event name validation, I will add it in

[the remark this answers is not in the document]
jamieomaguire: Ah, sorry I missed that! 

Yep, that sounds fine to me 👍 

It might be worth mentioning in the JSdoc that bubbles and composed will be true for now.

packages/components/pie-webc-core/src/test/functions/dispatchCustomElement/dispatchCustomEvent.browser.spec.ts
```diff
@@ -0,0 +1,45 @@
+import { test, expect } from '@sand4rt/experimental-ct-web';
+import { MockComponent } from './MockComponent.ts';
+
+test.describe('dispatchCustomEvent function', () => {
+    test('should dispatch an event when called', async ({ mount, page }) => {
+        // Arrange
+        const events : Array<Event> = [];
+
+        await mount(MockComponent, {
```
jamieomaguire, line 9:
```suggestion
        const component = await mount(MockComponent, {
```

packages/components/pie-webc-core/src/test/functions/dispatchCustomElement/dispatchCustomEvent.browser.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+            on: {
+                'pie-mock-event': (event: Event) => events.push(event),
+            },
+        });
+
+        // Act
+        await page.locator('[data-test-id="dispatch-custom-event-mock"]').click();
```
jamieomaguire, line 16:
```suggestion
        await component.click();
```

jamieomaguire: please can you give these two commits a try?

jamieomaguire (dismissed):

packages/components/pie-webc-core/src/test/functions/dispatchCustomElement/dispatchCustomEvent.browser.spec.ts
dandel10n: Interestingly only `await component.locator('[data-test-id="dispatch-custom-event-mock"]').click();` works. I compared the code with pie-button and don't see why there `await component.click();` does the trick 🤔

packages/components/pie-webc-core/src/test/functions/dispatchCustomElement/dispatchCustomEvent.browser.spec.ts
```diff
@@ -0,0 +1,45 @@
+import { test, expect } from '@sand4rt/experimental-ct-web';
+import { MockComponent } from './MockComponent.ts';
+
+test.describe('dispatchCustomEvent function', () => {
+    test('should dispatch an event when called', async ({ mount, page }) => {
```
fernandofranca, line 5: `should dispatch an event when clicked` perhaps?

fernandofranca (approved):

fernandofranca (approved):

raoufswe (approved):

PR #1303, opened 2024-02-23
chore(pie-assistive-text): DSW-1773 update conditional rendering

siggerzz:
This PR fixes an issue with pie-assistive-text where the conditional rendering implementation for icons was causing slots not to render in SPA applications. Moving the conditional logic to the template seems to fix the issue.

I've tested my changes in Aperture using the snapshot, and the slot now renders as expected across all aperture apps 👍🏼 

No visual tests to review

### DIFF
--- a/packages/components/pie-assistive-text/src/index.ts
+++ b/packages/components/pie-assistive-text/src/index.ts
@@ -21,7 +21,7 @@ const componentSelector = 'pie-assistive-text';
  * @slot - Default slot
  */
 export class PieAssistiveText extends LitElement implements AssistiveTextProps {
-    @property()
+    @property({ type: String })
     @validPropertyValues(componentSelector, variants, 'default')
     public variant?: AssistiveTextProps['variant'] = 'default';
 
@@ -32,8 +32,8 @@ export class PieAssistiveText extends LitElement implements AssistiveTextProps {
     private renderIcon (): TemplateResult {
         const { variant } = this;
         return html`
-            ${variant === 'success' ? html`<icon-check-circle class="c-assistiveText-icon" size="s" />` : nothing}
-            ${variant === 'error' ? html`<icon-alert-circle class="c-assistiveText-icon" size="s" />` : nothing}`;
+            ${variant === 'success' ? html`<icon-check-circle class="c-assistiveText-icon" size="s" ></icon-check-circle>` : nothing}
+            ${variant === 'error' ? html`<icon-alert-circle class="c-assistiveText-icon" size="s"></icon-alert-circle>` : nothing}`;
     }
 
     render () {
--- a/.changeset/forty-bats-float.md
+++ b/.changeset/forty-bats-float.md
@@ -1,0 +1,5 @@
+---
+"@justeattakeaway/pie-assistive-text": patch
+---
+
+[Fixed] - Issue where self-closing tag was causing rendering issues of variant icon in SPA's


raoufswe (dismissed):

packages/components/pie-assistive-text/src/index.ts
```diff
@@ -46,12 +35,12 @@ export class PieAssistiveText extends LitElement implements AssistiveTextProps {
             class="c-assistiveText"
             data-test-id="pie-assistive-text"
             variant=${ifDefined(variant)}>
-            ${this.renderIcon()}
-            <slot></slot>
-        </p>`;
+                ${variant === 'success' ? html`<icon-check-circle class="c-assistiveText-icon" size="s"></icon-check-circle>` : nothing}
```
fernandofranca, line 38: I understand `variant` can only be one thing, either success or error, however this doesn't seem the best pattern to solve the problem as it can become prone to inconsistency.

Can we solve this with something like this?

```
const iconTag = variant === 'success' ? 'icon-check-circle' : 'icon-alert-circle';

return html`<${iconTag} class="c-assistiveText-icon" size="s"></${iconTag}>`
```

siggerzz: I did try this, but it's not actually string interpolation as far as I can tell.

I have however refactored to follow Lit recommendations for conditional rendering - https://lit.dev/docs/templates/conditionals/#conditionals-with-if-statements

siggerzz: Just updated as part Lit recommendations for conditional rendering 👍🏼 - https://lit.dev/docs/templates/conditionals/#conditionals-with-if-statements

LTurns: I think pie-chip has a similar situation in that the slots are in between conditional rendering. So maybe we just need a conditional like `${variant != default ? this.renderIcon() : nothing}`? 😄 

https://github.com/justeattakeaway/pie/blob/main/packages/components/pie-chip/src/index.ts

packages/components/pie-assistive-text/src/index.ts
```diff
@@ -46,12 +35,12 @@ export class PieAssistiveText extends LitElement implements AssistiveTextProps {
...
+                ${variant === 'error' ? html`<icon-alert-circle class="c-assistiveText-icon" size="s"></icon-alert-circle>` : nothing}
```
jamieomaguire, line 39: Extracting this to a function shouldn't on its own cause the rendering error (this is how we render the modal too and is a common pattern for breaking down large rendering logic). Is your observation that simply moving this conditional logic to a function causes the issue? If so this might be worth flagging with Lit

packages/components/pie-assistive-text/src/index.ts
```diff
@@ -25,17 +25,6 @@ export class PieAssistiveText extends LitElement implements AssistiveTextProps {
...
     public variant?: AssistiveTextProps['variant'] = 'default';
 
-    /**
- * Renders the assistive-text icon content.
- * @private
- */
-    private renderIcon (): TemplateResult {
-        const { variant } = this;
-        return html`
-            ${variant === 'success' ? html`<icon-check-circle class="c-assistiveText-icon" size="s" />` : nothing}
```
jamieomaguire, line 35: I've just noticed we're using self-closing tags for the icons in the previous code. This is prone to causing rendering errors. Please could you test if simply adding closing tags to the icon elements fixes the issue @siggerzz ?

jamieomaguire (approved):

siggerzz:
Tested the snapshot and working as expected across all apps :)

raoufswe (approved):

LTurns (approved):

PR #1215, opened 2024-01-25
refactor(pie-monorepo): DSW-1645 Remove Jest in favour of Vitest

siggerzz:
## Author Checklist (complete before requesting a review)
- [ ] I have performed a self-review of my code
- [ ] If it is a core feature, I have added thorough tests
- [ ] If it is a `PIE Docs` change, I have reviewed the Docs site preview
- [ ] If it is a component change, I have reviewed the Storybook preview
- [ ] If there are visual test updates, I have reviewed them properly before approving

### Reviewer 1
- [ ] If it is a `PIE Docs` change, I have reviewed the PR preview
- [ ] If there are visual test updates, I have reviewed them

### Reviewer 2
- [ ] If it is a `PIE Docs` change, I have reviewed the PR preview
- [ ] If there are visual test updates, I have reviewed them

PR #1305, opened 2024-02-26
fix(pie-monorepo): DSW-000 issue where vitest dep was causing test:coverage to fail

siggerzz:
This PR fixes an issue where a missing dependency was causing the `test:coverage` script to fail.

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "@types/react": "18.2.56",
     "@typescript-eslint/eslint-plugin": "5.62.0",
     "@typescript-eslint/parser": "5.62.0",
-    "@vitest/coverage-v8": "1.2.1",
+    "@vitest/coverage-c8": "0.29.8",
     "autoprefixer": "10.4.17",
     "babel-loader": "8.3.0",
     "commitizen": "4.3.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -84,7 +84,7 @@
     "@types/react": "18.2.56",
     "@typescript-eslint/eslint-plugin": "5.62.0",
     "@typescript-eslint/parser": "5.62.0",
-    "@vitest/coverage-v8": "1.2.1",
+    "@vitest/coverage-c8": "0.29.8",
```
siggerzz, line 87: This may look like a typo, but it is intentional!

raoufswe (approved):

package.json
raoufswe: the package seems to be deprecated, maybe we need to find another solution 
https://www.npmjs.com/package/@vitest/coverage-c8

siggerzz: Yep, Unfortunately, our version of Vitest is super out-of-date. I'd recommend we update when we going to Vitest 1.x.
  
  
  I did test upgrading to Vitest as part of [this PR](https://github.com/justeattakeaway/pie/pull/1215), but there's too many issues with the package currently in our repo, and we need to wait for a bug to be resolved.

raoufswe (approved):

sebastienrocks (approved):

PR #1309, opened 2024-02-29
feat(pie-chip): DSW-1781 add needed aria properties

kevinrodrigues:


raoufswe (approved):

dandel10n (dismissed):

packages/components/pie-chip/src/index.ts
```diff
@@ -75,6 +78,7 @@ export class PieChip extends LitElement implements ChipProps {
                     <button
                         @click="${this._handleCloseButtonClick}"
                         ?disabled=${this.disabled}
+                        aria-label="${this.aria?.close || nothing}"
```
xander-marjoram, line 81: I think `ifDefined` is more appropriate to use for prop values and `nothing` is when you don't want to render an element

packages/components/pie-chip/src/index.ts
```diff
@@ -92,10 +96,15 @@ export class PieChip extends LitElement implements ChipProps {
 
         return html`
             <div
+                aria-atomic="true"
+                aria-busy="${isLoading ? 'true' : 'false'}"
```
xander-marjoram, line 100: Would this work? The lack of leading `?` might mean that it ends up being a string and not a boolean
```suggestion
                aria-busy="${isLoading}"
```

packages/components/pie-chip/src/index.ts
```diff
@@ -92,10 +96,15 @@ export class PieChip extends LitElement implements ChipProps {
...
+                aria-current="${isSelected ? 'true' : 'false'}"
+                aria-label="${this.aria?.label || nothing}"
```
xander-marjoram, line 102: Same point about `ifDefined` here

packages/components/pie-chip/test/component/pie-chip.spec.ts
```diff
@@ -79,4 +83,171 @@ test.describe('PieChip - Component tests', () => {
...
+                    });
+
+                    const closeButton = page.locator(closeBtnSelector);
+
+                    // Act && Assert
+                    await expect(closeButton).toHaveAttribute('aria-label', 'Chip Close');
+                });
+            });
+
+            test.describe('when an aria close value is not provided', () => {
```
xander-marjoram, line 108: I would say that there is still a value being provided here, it's just an empty string

packages/components/pie-chip/test/component/pie-chip.spec.ts
```diff
@@ -79,4 +83,171 @@ test.describe('PieChip - Component tests', () => {
...
+                            ...dismissibleProps,
+                            aria: {
+                                close: 'Chip Close',
+                            },
+                        } as ChipProps,
+                    });
+
+                    const closeButton = page.locator(closeBtnSelector);
+
+                    // Act && Assert
```
xander-marjoram, line 103: Super nit 😄 
```suggestion
                    // Act & Assert
```

packages/components/pie-chip/test/component/pie-chip.spec.ts
```diff
@@ -79,4 +83,171 @@ test.describe('PieChip - Component tests', () => {
...
+                await expect(chip).toHaveAttribute('aria-atomic', 'true');
+            });
+        });
+
+        test.describe('aria-label', () => {
+            test.describe('when passed in', () => {
+                test('should render on the component with the correct value', async ({ mount, page }) => {
+                    // Arrange
+                    await mount(PieChip, {
+                        props,
```
xander-marjoram, line 164: I can see that the value is being passed in inside `props`, but I think this test would be easier to read and understand if a different value was being passed in. Maybe the default values should be removed, and then they are given in the relevant tests?

packages/components/pie-chip/test/component/pie-chip.spec.ts
```diff
@@ -79,4 +83,171 @@ test.describe('PieChip - Component tests', () => {
...
+                test('should not render on the component', async ({ mount, page }) => {
+                    // Arrange
+                    await mount(PieChip, {
+                        props: {} as ChipProps,
+                    });
+
+                    const chip = page.locator(componentSelector);
+
+                    // Act && Assert
+                    await expect(chip).not.toHaveAttribute('aria-label', 'Chip Label');
```
xander-marjoram, line 184: It looks like this is checking only for this specific value, so if there was another value the test would still pass?

This might not be exact code, but I think I would prefer something like this:
```js
await expect(chip.attributes()).not.toContain('aria-label');
```

packages/components/pie-chip/src/index.ts
raoufswe: the aria-busy value is a true or false string not a boolean 
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-busy

xander-marjoram: Yeah, exactly! For a boolean we would do `?aria-busy=${isLoading}` so I'm wondering if we can just do `aria-busy=${isLoading}` to get the string value.

packages/components/pie-chip/test/component/pie-chip.spec.ts
kevinrodrigues: Yes good catch!

packages/components/pie-chip/test/component/pie-chip.spec.ts
kevinrodrigues: Not sure i follow this one. The values are passed from the prop set up above. I could pass in brand new props here (if that's what you mean?) but the whole test file follows this approach at the moment. I think passing in a brand new set of props does actually make it easier to follow to limit scrolling up and down to check but it does also shorten the file up a bit because your not repeating the same values across each test.

kevinrodrigues: Fixed.

raoufswe (approved):

xander-marjoram (approved):

PR #1316, opened 2024-03-04
fix(pie-monorepo): DSW-000 stop danger flagging renovate pr checkboxes

ashleynolan:

- [Fixed] - danger flagging automated prs


dangerfile.js
```diff
@@ -34,7 +34,7 @@ danger.git.created_files.filter((filepath) => filepath.includes('.changeset/') &
         });
     });
 
-// Check for empty PR Description checkboxes
-if (pr.body.includes('- [ ]')) {
+// Check for empty PR Description checkboxes - but not for automated version PRs
+if (pr.body.includes('- [ ]') && (pr.user.login !== 'dependabot[bot]' && pr.user.login !== 'renovate[bot]')) {
```
xander-marjoram, line 38: Doesn't have to be this PR but might be good to extract some of this to make it a little more readable so this line ends up being something like this:
```
if (prBodyContainsUncheckedItem && !isAuthorDependabot) {
```

xander-marjoram (approved):

dangerfile.js
ashleynolan: I'll tidy this up in a follow-up 👍

raoufswe (approved):

PR #1319, opened 2024-03-05
fix(pie-icons-react): DSW-000 fix incorrect "types" location in package.json

jansul:
Hey, we're trying to use `@justeattakeaway/pie-icons-react` in one of our Typescript projects. When we try and import one of the icons we receive the following error: 

> error TS7016: Could not find a declaration file for module '@justeattakeaway/pie-icons-react'. '/Users/xxx.xxx/xxx/node_modules/@justeattakeaway/pie-icons-react/dist/index.js' implicitly has an 'any' type.

I attempted to debug using `tsc --traceResolution`. It seems to try and load `esm/index.d.ts` as referenced in the[ package.json file](https://github.com/jansul/justeattakeaway-pie/blob/main/packages/tools/pie-icons-react/package.json#L6), however this does not seem to actually exist in the distributed files.

This PR updates the above to instead point to `esm/icons/index.d.ts`. There is also a copy in `dist/icons/index.d.ts`, however the esm version seemed more appropriate given it was already looking for a file there. I don't know too much about ES modules though 😅 

I've tested this by modifying `package.json` locally, and the types seem to work as expected:

image

### DIFF
--- a/.changeset/purple-oranges-pay.md
+++ b/.changeset/purple-oranges-pay.md
@@ -1,0 +1,5 @@
+---
+"@justeattakeaway/pie-icons-react": patch
+---
+
+[Fixed] - Incorrect "types" location in package.json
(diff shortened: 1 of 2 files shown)


jansul:
Replaced by https://github.com/justeattakeaway/pie/pull/1320

PR #1321, opened 2024-03-06
fix(pie-monorepo): DSW-000 fix issue with root execution of certain build scripts didn't run

siggerzz:
We have a slight quirk in the monorepo, where in the root package.json we need to specify the `pie-monorepo` package in the workspaces package.json like so:

```
  "workspaces": {
    "packages": [
      "apps/**/*",
      "configs/**/*",
      "packages/**/*",
      "./" <<<-------- This line
    ]
  },
  ```
  
 The reason for doing this, is that without it, running `yarn changeset` would not pick up `pie-monorepo` as a versionable package.

One of the downsides to this approach, is that because Turborepo also uses yarn workspaces to determine the packages a given task needs to run for, it can result in a recursion issue unless the `--filter=!pie-monorepo` flag is provided for any build scripts that rely on Turborepo.

The absence of this flag meant that a number of scripts weren't able to execute when run from the root of the monorepo.

### CODE REVIEW

File: .changeset/brave-owls-wink.md [hunk 1]
```diff
@@ -0,0 +1,6 @@
+---
+"pie-monorepo": patch
+---
+
+[Fixed] - Issue where `create:manifest` & `build:react-wrapper` commands failured when executed from the monorepo root directory.
```
xander-marjoram, line 5:
```suggestion
[Fixed] - Issue where `create:manifest` & `build:react-wrapper` commands failed when executed from the monorepo root directory.
```

dandel10n (approved):

fernandofranca (approved):

PR #1326, opened 2024-03-07
chore(pie-monorepo): DSW-1783 refactor generation of playwright html

siggerzz:
This PR moves the `playwright` folder used by component tests into the `@justeattakeaway/pie-components-config` package.

A new script called `test:browsers-setup` gets executed prior to the component tests to copy this file into the working directory of any dependant packages prior to test execution.

Ideally, we would just tell the tests to reference the file directly from config, but due to some limitations in Playwright, it looks like they've hardcoded the logic to expect this folder in the same directory as the playwright config, meaning we have to be slightly creative and copy the files over.


packages/tools/pie-icons-webc/playwright/index.html
```diff
@@ -52,6 +52,5 @@
 </head>
 <body>
     <div id="root"></div>
-    <script type="module" src="./index.ts"></script>
```
xander-marjoram, line 55: Should this file also have been deleted?

configs/pie-components-config/package.json
```diff
@@ -5,10 +5,14 @@
...
   "author": "Just Eat Takeaway.com - Design System Team",
   "description": "Shared configuration files for PIE components",
+  "bin": {
+    "playwright-lit-setup": "./scripts/playwright-lit-setup.js"
+  },
   "files": [
     "custom-elements-manifest.config.js",
     "vite.config.js",
     "tsconfig.json",
+    "playwright/**",
```
xander-marjoram, line 15: Would this also include any cache folder?

packages/tools/pie-icons-webc/playwright/index.html
siggerzz: Yeah you're right. I was hoping this could be deleted as it's just an empty file, but I looks like playwright probably uses this file to inject the component code.

configs/pie-components-config/package.json
siggerzz: Addressed this with a new .npmignore file 👍🏼

xander-marjoram (approved):

.gitignore
```diff
@@ -162,7 +162,8 @@ browsers-report/
 lit-browsers-report/
 visual-report/
 lit-visual-report/
-react-browsers-report
+react-browsers-report/
+packages/components/**/playwright/
```
jamieomaguire, line 166: do you need to do the cache remove thing we spoke about last week?

siggerzz: Nope, this should be fine :)

jamieomaguire (approved):

PR #1333, opened 2024-03-15
feat(pie-cookie-banner): DSW-1754 Added page object to cookie banner

JoshuaNg2332:
- Added page object support to cookie banner
- Refactored existing component tests to support page objects


.changeset/dull-vans-rush.md
```diff
@@ -0,0 +1,5 @@
+---
+"@justeattakeaway/pie-cookie-banner": minor
+---
+
+[Added] - Page object supporting files and refactors
```
xander-marjoram, line 5: Is a changeset required? I reckon we can get away without one seeing as there are no component changes

packages/components/pie-cookie-banner/test/helpers/page-object/pie-cookie-banner.page.ts
```diff
@@ -0,0 +1,135 @@
...
+    readonly bodyAcceptAllLinkLocator: Locator;
+    readonly bodyNecessaryOnlyLinkLocator;
+    readonly bodyManagePreferencesLinkLocator: Locator;
+    readonly bodyCookieStatementLinkLocator: Locator;
+    readonly bodyCookieTechnologiesLinkLocator: Locator;
+    readonly modalDescriptionLocator: Locator;
+
+    constructor (page: Page) {
+        super(page);
+        this.page = page;
```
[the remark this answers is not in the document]
xander-marjoram: And in theory you could probably get rid of `this.page` on L8 because it should also come from the super class.

packages/components/pie-cookie-banner/test/helpers/page-object/pie-cookie-banner.page.ts
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    async isPreferenceToggleDisabled (preferenceIds: string) {
+        const preferenceToggle = this.page.locator(preferenceIds);
+        return preferenceToggle.isChecked();
+    }
+
+    async getAcceptAllTextContent () {
+        const acceptAllText = await this.acceptAllButtonLocator.textContent();
+        return String(acceptAllText).trim();
```
xander-marjoram, line 102: Out of interest, why is `String` needed here?

xander-marjoram: Just thinking that if `acceptAllText` comes back as `null`, this will return `"null"`.
Would this do what you need it to?
```suggestion
        return acceptAllText?.trim();
```
It returns either `undefined` or the trimmed string.

packages/components/pie-cookie-banner/test/helpers/page-object/selectors.ts
```diff
@@ -0,0 +1,51 @@
+const cookieBanner = {
+    selectors: {
+        container: {
+            description: 'The selector for the cookie banner container',
+            dataTestId: 'pie-cookie-banner',
```
xander-marjoram, line 5: Would it be a bad idea to import these into the component?

JoshuaNg2332: Yeah, so ideally, we'd be going against the page object model by importing these into our component directly as generally, consumers will only be using the functionality we'd provide through our page object. By exposing these in the tests, we'd essentially be giving them free reign to start exploring the rest of the component when the main focus for consumers is to just use our testing structure.

packages/components/pie-cookie-banner/test/helpers/page-object/pie-cookie-banner.page.ts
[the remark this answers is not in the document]
JoshuaNg2332: Just tested and that works. Makes sense! Thanks.

.changeset/dull-vans-rush.md
JoshuaNg2332: Is not adding a page object support considered adding a feature? It does change how we test our components?

xander-marjoram: Yeah, it's adding (test) functionality and improving the component, for sure, but I don't think it changes anything that is published to npm (currently?) so I don't think it makes sense to publish a new version.

packages/components/pie-cookie-banner/test/component/pie-cookie-banner.spec.ts
```diff
@@ -323,31 +320,31 @@ test.describe('PieCookieBanner - Component tests', () => {
             );
 
             // Act
-            const acceptAllButton = component.locator('[data-test-id="actions-accept-all"]');
-            const necessaryOnlyButton = component.locator('[data-test-id="actions-necessary-only"]');
+            const acceptAllButton = await pieCookieBannerComponent.getAcceptAllAttribute('variant');
```
xander-marjoram, line 323: Nit: I would rename these variables to be `acceptAllButtonVariant` and `necessaryOnlyButtonVariant` to make the expectations read a bit more smoothly

packages/components/pie-cookie-banner/test/component/pie-cookie-banner.spec.ts
```diff
@@ -356,42 +353,42 @@ test.describe('PieCookieBanner - Component tests', () => {
...
+                await mount(
                     PieCookieBanner,
                     {
                         props: {},
                     },
                 );
 
                 // Act
-                const cookieStatementLink = component.locator('[data-test-id="banner-description"] [data-test-id="cookie-statement-link"]');
+                const cookieBannerDescription = await pieCookieBannerComponent.getBannerCookieStatementLinkAttribute('href');
```
xander-marjoram, line 364: This should stay as `cookieBannerLink`, or maybe `cookieBannerHref`, right?

packages/components/pie-cookie-banner/test/component/pie-cookie-banner.spec.ts
```diff
@@ -149,47 +145,48 @@ test.describe('PieCookieBanner - Component tests', () => {
...
         // Arrange
         await mount(PieCookieBanner, {
             props: {} as CookieBannerProps,
         });
 
         // Act
-        await page.click(managePreferencesSelector);
-        const necessaryPreferenceToggle = await page.locator(getPreferenceItemSelector('necessary'));
+        await pieCookieBannerComponent.clickManagePreferencesAction();
+        const necessaryPreferenceToggleChecked = await pieCookieBannerComponent.isPreferenceToggleChecked(getPreferenceItemSelector('necessary'));
```
xander-marjoram, line 171: Is there any way this `getPreferenceItemSelector` could be worked into the cookie banner page object?

That would lead to something like this, which looks reasonable to me:
`pieCookieBannerComponent.isPreferenceToggleChecked('necessary')`

packages/components/pie-modal/test/helpers/page-object/selectors.ts
```diff
@@ -20,6 +20,10 @@ const modal = {
             description: 'The selector for the modal content',
             dataTestId: 'modal-content-inner',
         },
+        description: {
+            description: 'The selector for the modal description',
```
xander-marjoram, line 24: I wonder if this could just be a JSON file a bit like this:

Not for this PR but just something to consider in future to see if it might simplify the imports, etc.

```json
{
    "container": {
        "description": "The selector for the modal container",
        "dataTestId": "pie-modal"
    },
    "backButton": {
        "description": "The selector for the back button",
        "dataTestId": "modal-back-button"
    },
    "closeButton": {
        "description": "The selector for the close button",
        "dataTestId": "modal-close-button"
    },
    "header": {
        "description": "The selector for the modal header",
        "dataTestId": "modal-header"
    },
    "content": {
        "description": "The selector for the modal content",
        "dataTestId": "modal-content-inner"
    },
    "description": {
        "description": "The selector for the modal description",
        "dataTestId": "modal-description"
    }
}
```

xander-marjoram: Then you might be able to do something like `import selectors from '/path/to/modal/selectors.json';`

Although then that only cuts it down from `modal.selectors.content.description` to `selectors.content.description`, so maybe the easier thing to do would just be to rename the export and then remove one level of nesting.

```ts
const selectors = {
    container: {
        description: 'The selector for the modal container',
        dataTestId: 'pie-modal',
    },
    backButton: {
        description: 'The selector for the back button',
        dataTestId: 'modal-back-button',
    },
    closeButton: {
        description: 'The selector for the close button',
        dataTestId: 'modal-close-button',
    },
    header: {
        description: 'The selector for the modal header',
        dataTestId: 'modal-header',
    },
    content: {
        description: 'The selector for the modal content',
        dataTestId: 'modal-content-inner',
    },
    description: {
        description: 'The selector for the modal description',
        dataTestId: 'modal-description',
    },
};

export {
    selectors,
};
```

packages/components/pie-cookie-banner/test/component/pie-cookie-banner.spec.ts
xander-marjoram: I think my point was this doesn't seem to be getting the description, as it's looking for the `href` attribute on a link, rather than the text content of a description. The other tests below that do similar all have `cookieBannerLink`, which I like.

packages/components/pie-cookie-banner/test/component/pie-cookie-banner.spec.ts
```diff
@@ -13,40 +15,34 @@ import {
...
             props: {} as CookieBannerProps,
         });
 
         // Act
-        const cookieBanner = page.locator(componentSelector);
+        const cookieBanner = await pieCookieBannerComponent.isCookieBannerVisible();
 
         // Assert
-        expect(cookieBanner).toBeVisible();
+        expect(cookieBanner).toBeTruthy();
```
jamieomaguire, line 41: can this not be `toBe(true)`? is a bit more explicit

packages/components/pie-cookie-banner/test/component/pie-cookie-banner.spec.ts
```diff
@@ -106,38 +100,38 @@ test.describe('PieCookieBanner - Component tests', () => {
...
+            await pieCookieBannerComponent.clickManagePreferencesAll(elementLevel.name);
+
+            const modal = pieModalComponent.isModalVisible();
+            const cookieBanner = await pieCookieBannerComponent.isCookieBannerVisible();
 
             // Assert
-            expect(modal).toBeVisible();
-            expect(cookieBanner).not.toBeVisible();
+            expect.soft(modal).toBeTruthy();
+            expect.soft(cookieBanner).toBeFalsy();
```
jamieomaguire, line 111: I think we've introduced some ambiguity here. The previous test was very easy to understand in it's assertions and this seems to be a lot less clear.

Why do we check (softly) that the cookiebanner is falsy?

packages/components/pie-cookie-banner/test/component/pie-cookie-banner.spec.ts
JoshuaNg2332: Yeah, playwright's native , `toBeTrue` uses truthy instead of true but Truthy is already an explicit command?

`Value is truthy, i.e. not false, 0, null, etc.`

packages/components/pie-cookie-banner/test/component/pie-cookie-banner.spec.ts
JoshuaNg2332: Ideally soft checks should be made before the final assert as sometimes you can get into a situation where if you have multiple `expect` assertions in a test, you can easily lose track as to which of these has caused an error and can also fail early. Ideally, you want to be doing soft asserts before the final hard assert as then if one of these tests were to fail, you'll easily be able to identify which of these assertions are causing the failure.

@siggerzz and I have also discussed about the ambiguity of our testing names with use of `falsy` and `truthy` and have both agreed for it to be something we should look into in the future as we'd much rather be able to have page objects but also use the native playwright functions in the tests.

packages/components/pie-cookie-banner/test/component/pie-cookie-banner.spec.ts
jamieomaguire: so the result could be `''` instead of `true`?

packages/components/pie-cookie-banner/test/helpers/page-object/pie-cookie-banner.page.ts
```diff
@@ -0,0 +1,138 @@
...
+    readonly descriptionLocator: Locator;
+    readonly acceptAllButtonLocator: Locator;
+    readonly necessaryOnlyButtonLocator: Locator;
+    readonly managePrefsLinkLocator: Locator;
+    readonly bodyAcceptAllLinkLocator: Locator;
+    readonly bodyNecessaryOnlyLinkLocator;
+    readonly bodyManagePreferencesLinkLocator: Locator;
+    readonly bodyCookieStatementLinkLocator: Locator;
+    readonly bodyCookieTechnologiesLinkLocator: Locator;
+    readonly modalDescriptionLocator: Locator;
```
jamieomaguire, line 23: I think you can also mark these as private if consumers don't need knowledge of them :)

packages/components/pie-cookie-banner/test/helpers/page-object/pie-cookie-banner.page.ts
```diff
@@ -0,0 +1,138 @@
...
+        this.managePrefsLinkLocator = page.getByTestId(cookieBanner.selectors.managePreferencesLink.dataTestId);
+        this.bodyAcceptAllLinkLocator = page.getByTestId(cookieBanner.selectors.bodyAcceptAllLink.dataTestId);
+        this.bodyNecessaryOnlyLinkLocator = page.getByTestId(cookieBanner.selectors.bodyNecessaryOnlyLink.dataTestId);
+        this.bodyManagePreferencesLinkLocator = page.getByTestId(cookieBanner.selectors.bodyManagePreferencesLink.dataTestId);
+        this.bodyCookieStatementLinkLocator = page.getByTestId(cookieBanner.selectors.bodyCookieStatementLink.dataTestId);
+        this.bodyCookieTechnologiesLinkLocator = page.getByTestId(cookieBanner.selectors.bodyCookieTechnologiesLink.dataTestId);
+        this.modalDescriptionLocator = page.getByTestId(cookieBanner.selectors.modalDescription.dataTestId);
+    }
+
+    async isCookieBannerVisible () {
```
jamieomaguire, line 40: NIT: I would add typing to these methods. I.e. `: boolean` to show what type they return

packages/components/pie-cookie-banner/test/helpers/page-object/pie-cookie-banner.page.ts
```diff
@@ -0,0 +1,138 @@
...
+
+    async clickManagePreferencesLink () {
+        await this.bodyManagePreferencesLinkLocator.click();
+    }
+
+    async clickManagePreferencesAll (elementLevel: string) {
+        await (elementLevel === 'action' ? this.clickManagePreferencesAction() : this.clickManagePreferencesLink());
+    }
+
+    async clickPreferenceToggle (preferenceIds: PreferenceIds) {
```
jamieomaguire, line 88: Small question, what is the stacktrace like in the tests if any of these page object methods fail? Do we need to handle errors to provide useful messaging?

JoshuaNg2332: The stack trace comes through just like it would if the native function was run in the test. So for example, if it was trying to find a locator and couldn't find it, the stack trace would indicate that you are trying to invoke this function that calls on x particular locator and state that it cannot find it. 
For this reason, I don't think we need to add error handling.

packages/components/pie-cookie-banner/test/component/pie-cookie-banner.spec.ts
```diff
@@ -13,40 +15,34 @@ import {
...
+
+    test('should render successfully', async ({ mount }) => {
         // Arrange
         await mount(PieCookieBanner, {
             props: {} as CookieBannerProps,
         });
 
         // Act
-        const cookieBanner = page.locator(componentSelector);
+        const cookieBanner = await pieCookieBannerComponent.isCookieBannerVisible();
```
jamieomaguire, line 38:
```suggestion
        const cookieBannerIsVisible = await pieCookieBannerComponent.isCookieBannerVisible();
```

packages/components/pie-cookie-banner/test/component/pie-cookie-banner.spec.ts
```diff
@@ -13,40 +15,34 @@ import {
...
 
         // Assert
-        expect(cookieBanner).toBeVisible();
+        expect(cookieBanner).toBe(true);
```
jamieomaguire, line 41:
```suggestion
        expect(cookieBannerIsVisible).toBe(true);
```

packages/components/pie-cookie-banner/test/helpers/page-object/pie-cookie-banner.page.ts
```diff
@@ -0,0 +1,282 @@
...
+     *                          within the body has been successfully clicked.
+     */
+    async clickBodyAcceptAllLink () : Promise<void> {
+        await this.bodyAcceptAllLinkLocator.click();
+    }
+
+    /**
+     * Clicks either the "Accept All" button or the "Accept All" link within the body, based on the specified element level.
+     *
+     * @param {string} elementLevel The level of the element to click, either 'action' or 'body'.
```
jamieomaguire, line 73: This is an excellent opportunity to use a string union type! It could be `action | body`

[reference](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html)

packages/components/pie-cookie-banner/test/helpers/page-object/pie-cookie-banner.page.ts
```diff
@@ -0,0 +1,284 @@
...
+
+    /**
+     * Checks whether the preference toggle associated with the specified preference IDs is disabled.
+     *
+     * @param {PreferenceIds} preferenceIds The preference IDs used to locate the preference toggle.
+     * @returns {Promise<boolean>} A Promise that resolves to a boolean value indicating whether the preference toggle is disabled (`true`) or not (`false`).
+     */
+    async isPreferenceToggleDisabled (preferenceIds: PreferenceIds) : Promise<boolean> {
+        const preferenceToggle = this.page.locator(getPreferenceItemSelector(preferenceIds));
+        return preferenceToggle.isChecked();
```
raoufswe, line 194: i might be missing something, but based on the method names, we should check for preferenceToggle.isDisabled() right?

JoshuaNg2332: Oops yeah, I think you're right! I'll fix now.

raoufswe (approved):

ashleynolan (approved):

jamieomaguire (approved):

PR #1348, opened 2024-03-22
fix(pie-chip): DSW-000 hover and active styles are applied for disabled state

raoufswe:

- Fixes an issue raised by design that the hover and active styles are applied when disabled is true.

### Reviewer 1

### Reviewer 2


packages/components/pie-chip/src/chip.scss
```diff
@@ -1,7 +1,7 @@
 @use '@justeattakeaway/pie-css/scss' as p;
 
 @mixin chip-interactive-states($bg-color, $mode: 'default') {
-    &:hover:not(:disabled) {
+    &:hover:not([disabled]) {
```
raoufswe, line 4: the css`:disabled` selector is only applied if the element can be natively disabled

ashleynolan (approved):

.changeset/twenty-pianos-invent.md
```diff
@@ -0,0 +1,5 @@
+---
+"@justeattakeaway/pie-chip": minor
```
ashleynolan, line 2: If you wanted to, you could use patch instead of minor here (as it's a small fix). Not a biggie though 😆

dandel10n (approved):

packages/components/pie-chip/src/chip.scss
jamieomaguire: good spot!

jamieomaguire (approved):

.changeset/twenty-pianos-invent.md
raoufswe: i will keep that in mind, thanks for flagging it 🥇

PR #1367, opened 2024-03-28
feat(pie-icons-webc): DSW-000 set height and width on icon hosts

jamieomaguire:
Applies height and width to web component `host` along with a display inline block. Otherwise, the web component hosts are incorrectly sized. Image attached to demo

Screenshot 2024-03-28 at 15 14 02

### Reviewer 1

### Reviewer 2

### DIFF
--- a/packages/tools/pie-icons-webc/build.js
+++ b/packages/tools/pie-icons-webc/build.js
@@ -36,13 +36,17 @@ const componentSelector = '${kebabCase(name)}';
  * @tagname ${kebabCase(name)}
  */
 export class ${name} extends LitElement implements IconProps {
-    // The following styles make sure that the icon will be sized correctly
+    // The following styles make sure that the icon will be sized correctly.
+    // "--icon-size-override" css var can be used in parent components to make sure the icon
+    // stays the correct size for cases where the icon is added by consumers via a slot,
+    // for example in Button, IconButton, Chip and Tag.
+    // Might be changed later to assigning a size prop to an icon slot
+    // (POC: https://github.com/justeattakeaway/pie/pull/1128)
     static styles = css\`
         :host {
             display: inline-block;
         }
         :host svg {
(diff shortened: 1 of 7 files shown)


packages/tools/pie-icons-webc/build.js
```diff
@@ -58,6 +61,8 @@ export class ${name} extends LitElement implements IconProps {
         super.connectedCallback();
         if (this._svg?.getAttribute('width') === null) {
             const svgSize = getSvgProps('${svgClasses}', '', null, '${name}');
+            this.style.width = \`\${\`var(--icon-size-override,\${svgSize.width}\`}px)\`;
+            this.style.height = \`\${\`var(--icon-size-override,\${svgSize.height}\`}px)\`;
             this._svg?.setAttribute('width', svgSize.width);
             this._svg?.setAttribute('height', svgSize.height);
```
dandel10n, line 67: Feels weird that we set here both width and height attributes and at the same time set width and height through the inline styles 🤔

packages/tools/pie-icons-webc/build.js
```diff
@@ -38,6 +38,9 @@ const componentSelector = '${kebabCase(name)}';
 export class ${name} extends LitElement implements IconProps {
     // The following styles make sure that the icon will be sized correctly
     static styles = css\`
+        :host {
+            display: inline-block;
+        }
         :host svg {
```
dandel10n, line 44: Could you please add a comment here to clarify the reasoning for this override vars, something like this:
// Variables can be used in parent components to make sure that the icon stays the correct size for cases where the icon is added by consumers via a slot, for example in IconButton 
// Might be changed later to assigning a size prop to an icon slot (POC: https://github.com/justeattakeaway/pie/pull/1128)

dandel10n (approved):
Looks good, the only comment I have is about the comment that would give more context about current implementation

PR #1373, opened 2024-04-02
chore(pie-storybook): DSW-000 fix type declarations for storybook background colours

xander-marjoram:
Use `satisfies` for type guarding instead of asserting with `as`.

Previously, TypeScript did not identify that `'background-dark'` was not in the `BackgroundColor` string union. In fact, you could add something like `{ test: 'hello' }` to the background values and TypeScript would not complain.

Changing `as` to `satisfies` ensures proper type guarding, correctly identifying that `'background-dark'` was not an expected colour name.

Now if you try adding `{ test: 'hello' }` to the background values object, you get the following sensible error message:
```
Object literal may only specify known properties, and 'test' does not exist in type 'BackgroundValue'.
```

> [!NOTE]
> This only affects the type checking. Functionality is unaffected as it was previously working.

- [x] If it is a `PIE storybook` change, I have reviewed the Storybook preview

### Reviewer 1
- [x] If it is a `PIE Storybook` change, I have reviewed the Storybook preview

### Reviewer 2
- [x] If it is a `PIE Storybook` change, I have reviewed the Storybook preview


kevinrodrigues (approved):
Nice.

jamieomaguire (approved):

PR #1374, opened 2024-04-02
feat(pie-input): DSW-1584 required prop and focus invalid

jamieomaguire:
- adds a new `required` prop to `pie-input`
- updates storybook to show the new required prop
- adds a new mechanism for managing forms on the page and enabling focusing the first invalid input in a form on submit (validity is determined based on the `ValidityState` exposed by the control 
- integrates the new `PieFormManager` into the form control mixin so that new form controls automatically set it up if required
- unit and browser testing

## PieFormManager
This code may eventually be removed if we decide to not focus invalid inputs. However the mechanism may still be useful if we want to perform any other DS related tasks on form submits. 

The class is designed to be attached to the global window object and hold weak references to forms associated with our components (weakly held so that garbage collection can prevent memory leaks). It essentially applies and removes submit listeners to the stored forms.

This approach ensures that form control components do not need to store any code relating to form submission.

A (terribly drawn) flowchart outlining the lifecycle of a form control that uses the `formControlMixin`:

pieforms

pieFormManager

### Reviewer 1

### Reviewer 2


apps/pie-storybook/stories/pie-input.stories.ts
```diff
@@ -194,6 +194,13 @@ const inputStoryMeta: InputStoryMeta = {
             control: 'select',
             options: sizes,
         },
+        required: {
+            description: 'If true, the input is required to have a value before submitting the form. The form will autofocus a required field if it is missing a value. Will have no effect if the input is placed inside a form.',
```
raoufswe, line 198: is this accurate? 

```
Will have no effect if the input is placed inside a form.
```

packages/components/pie-input/README.md
```diff
@@ -93,6 +93,7 @@ import { PieInput } from '@justeattakeaway/pie-input/dist/react';
 | `min` | `number` | - | The minimum value of the input. Only applies when type is `number`. If the value provided is lower, the input is invalid. |
 | `max` | `number` | - | The maximum value of the input. Only applies when type is `number`. If the value provided is higher, the input is invalid. |
 | `size` | `'small'`, `'medium'`, `'large'` | `medium` | The size of the input field. Can be `small`, `medium`, or `large`. Defaults to `medium`. |
+| `required` | `boolean` | `false` | If true, the input is required to have a value before submitting the form. The form will autofocus a required field if it is missing a value. Will have no effect if the input is placed inside a form. |
```
xander-marjoram, line 96: Same as Raouf's comment but here

apps/pie-storybook/stories/pie-input.stories.ts
xander-marjoram: In readme and defs too, as I'm sure you're aware

jamieomaguire: lol! yeah thanks for spotting. will update now 👍 🚀

packages/components/pie-webc-core/src/forms/pie-form-manager.ts
```diff
@@ -0,0 +1,88 @@
...
+
+        this._forms.set(form, data);
+    }
+
+    /**
+     * Removes a form from the form manager and the submit event listener from it.
+     * @param form - The form to remove from the form manager.
+     */
+    public deleteForm (form: HTMLFormElement): void {
+        const data = this._forms.get(form);
```
xander-marjoram, line 58: Nit: Could you use `getForm` here?

xander-marjoram: And potentially in `addForm` too

jamieomaguire: lovely spot - updating now

fernandofranca:
This looks great though I'm not sure we should do validation/focus management. Perhaps I missed some key factor that makes this part necessary.

packages/components/pie-input/test/component/pie-input.spec.ts
```diff
@@ -796,6 +796,98 @@ test.describe('PieInput - Component tests', () => {
...
+                    props: {
+                        required: true,
+                    } as InputProps,
+                });
+
+                // Act
+                const input = component.locator('input');
+
+                // Assert
+                expect((await input.getAttribute('required'))).toBe('');
```
raoufswe, line 824: i might be missing something, but based on [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute), `getAttribute` will return `null `or empty string when the attribute isn't found but the test is a positive clause. 

maybe we can use [hasAttribute](https://developer.mozilla.org/en-US/docs/Web/API/Element/hasAttribute)

packages/components/pie-input/test/component/pie-input.spec.ts
```diff
@@ -1113,5 +1205,23 @@ test.describe('PieInput - Component tests', () => {
...
+                </form>
+            `);
+
+            const inputLocator = page.locator('pie-input');
+
+            // Act
+            await page.click('button[type="submit"]');
+
+            // Assert
+            await expect(inputLocator).toBeFocused();
```
raoufswe, line 1224: 🔥

packages/components/pie-input/test/component/pie-input.spec.ts
jamieomaguire: Yeah happy to change

jamieomaguire: Screenshot 2024-04-09 at 14 54 10

I don't think our version of playwright supports that method!

packages/components/pie-webc-core/src/mixins/formControl/formControlMixin.ts
```diff
@@ -47,11 +48,46 @@ export const FormControlMixin =
...
+            override connectedCallback (): void {
+                super.connectedCallback();
+
+                if (this.form) {
+                    // Ensure that the form is managed by the PieFormManager
+                    if (!window.pieFormManager) {
+                        window.pieFormManager = new PieFormManager();
+                    }
+
+                    if (window.pieFormManager) {
```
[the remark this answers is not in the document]
jamieomaguire: Yes! good spot

packages/components/pie-webc-core/src/mixins/formControl/formControlMixin.ts
```diff
@@ -47,11 +48,46 @@ export const FormControlMixin =
...
+                        window.pieFormManager.addForm(this.form);
```
raoufswe: a question: have we considered the use-case where two different components are calling the mixin? Do we need to check if the form hasn't been added yet or is there a check somewhere else / not needed?

packages/components/pie-input/test/component/pie-input.spec.ts
raoufswe: are we sure that the empty string here means the `required` attribute is found not `doesn't exist` as per the MDN spec?

packages/components/pie-webc-core/src/mixins/formControl/formControlMixin.ts
jamieomaguire: If I understand your question correctly, this line hopefully answers?

https://github.com/justeattakeaway/pie/pull/1374/files#diff-405085fc57869377bc65fdd5d8f99a1b2ae2512d6afbb446713cad440f7466f7R39-R41

Any input is allowed to try to add its form to the manager, but it can only be added once

packages/components/pie-input/test/component/pie-input.spec.ts
jamieomaguire: According to the spec you linked:

"All modern web browsers return null when the specified attribute does not exist on the specified element."

And we can see this here:
Screenshot 2024-04-09 at 14 51 19

I am unsure how else to test this but I am fully open to suggestions

raoufswe: well the browser is a source of truth and it's returning an empty string when it's assigned LOL. I would say we should use the hasAttrubte but seems it's missing. 🤝

packages/components/pie-webc-core/src/mixins/formControl/formControlMixin.ts
raoufswe: yep exactly nice!

packages/components/pie-webc-core/src/test/forms/pie-form-manager.spec.ts
```diff
@@ -0,0 +1,198 @@
...
+
+            const testForm = document.getElementById('testForm') as HTMLFormElement;
+            const manager = new PieFormManager();
+            const addEventListenerSpy = vi.spyOn(testForm, 'addEventListener');
+
+            // Act
+            manager.addForm(testForm);
+
+            // Assert
+            expect(addEventListenerSpy).toHaveBeenCalledWith('submit', expect.any(Function));
```
xander-marjoram, line 26: What's triggering the form submission here?

packages/components/pie-webc-core/src/test/forms/pie-form-manager.spec.ts
```diff
@@ -0,0 +1,198 @@
+import {
+    describe,
+    it,
+    expect,
+    vi,
+} from 'vitest';
```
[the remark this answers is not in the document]
jamieomaguire: ah good point - removing now

packages/components/pie-webc-core/src/test/forms/pie-form-manager.spec.ts
jamieomaguire: There is no form submission, this is checking that `addEventListener` was called for the submit event

packages/components/pie-webc-core/src/test/forms/pie-form-manager.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+
+            const testForm = document.getElementById('testForm') as HTMLFormElement;
+            const manager = new PieFormManager();
+            manager.addForm(testForm);
+
+            // Act
+            const result = manager.getForm(testForm);
+
+            // Assert
+            expect(result?.form).toStrictEqual(testForm);
```
xander-marjoram, line 54: Do we also want the listener, for completeness?

packages/components/pie-input/test/component/pie-input.spec.ts
jamieomaguire: I fully agree 😃 hopefully when we update playwright we can revisit!

packages/components/pie-webc-core/src/test/forms/pie-form-manager.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+            // Act
+            const result = manager.getForm(testForm);
+
+            // Assert
+            expect(result).toBeUndefined();
+        });
+    });
+
+    describe('deleteForm', () => {
+        it('does not return the PieFormData object associated with a form after it has been deleted', () => {
```
xander-marjoram, line 93: I think this test is identical to the one above?

packages/components/pie-webc-core/src/test/forms/pie-form-manager.spec.ts
jamieomaguire: for-completeness

packages/components/pie-webc-core/src/test/forms/pie-form-manager.spec.ts
jamieomaguire: correct! removing it now

xander-marjoram (approved):

PR #1388, opened 2024-04-09
feat(pie-notification): DSW-1901 support full width notification (banner)

thejfreitas:
This PR adds a new prop to the pie-notification component `position` which applies the constraints of the design spects that the notification component does not have border radius when set to `full-width`

### Description

Screenshot 2024-04-09 at 9 59 03 AM
Screenshot 2024-04-09 at 9 58 55 AM
Screenshot 2024-04-09 at 9 58 15 AM
Screenshot 2024-04-09 at 9 57 52 AM

═══════════════════════════════════════════════════════════

### Reviewer 1 - @raoufswe 


packages/components/pie-notification/test/component/pie-notification.spec.ts
```diff
@@ -473,6 +473,53 @@ test.describe('PieNotification - Component tests', () => {
...
+                await mount(PieNotification, {
+                    props: {},
+                });
+
+                // Act
+                const notification = page.locator(componentSelector);
+
+                // Assert
+                expect(notification).toBeVisible();
+                expect(notification).toHaveCSS('border-radius', '12px');
```
xander-marjoram, line 488: These feel a bit more like visual tests to me. I'd be interested to know what other people think though.

apps/pie-storybook/stories/pie-notification.stories.ts
```diff
@@ -52,6 +53,14 @@ const notificationStoryMeta: NotificationStoryMeta = {
                 summary: 'neutral',
             },
         },
+        position: {
+            description: 'Notifications can appear anywhere inline within the content or full width which should appear at the top of the interface, under the header',
```
ashleynolan, line 57:
```suggestion
            description: 'Specifies whether the notification should be displayed inline, within the content, or full width (recommended at the top of the interface, under the header)',
```

packages/components/pie-notification/test/component/pie-notification.spec.ts
thejfreitas: In my opinion, although it looks quite close to a visual test I believe this is suitable as a unit test since I am testing the behaviour of the component based on its inputs as well what is the default behaviour if no prop is passed. IMHO these unit tests and visual tests added in this PR complement each other and I don't see that as duplication.

packages/components/pie-notification/src/index.ts
```diff
@@ -48,6 +49,10 @@ export class PieNotification extends LitElement implements NotificationProps {
     @validPropertyValues(componentSelector, variants, 'neutral')
     public variant: NonNullable<NotificationProps['variant']> = 'neutral';
 
+    @property()
+    @validPropertyValues(componentSelector, positions, 'inline-content')
+    public position: NonNullable<NotificationProps['position']> = 'inline-content';
```
jamieomaguire, line 54: I am interested in this pattern of using NonNullable. Could you explain a bit more about how it works please @thejfreitas ? :)

packages/components/pie-notification/test/component/pie-notification.spec.ts
jamieomaguire: @thejfreitas I see your point and it's entirely valid. What I wonder is do we expect this test to fail but the visual test to pass? I think the answer is probably not, and therefore I would personally prefer we leave this to visual testing.

Another issue is if the design token value changes then we have to update this test as we're referencing a raw value.

packages/components/pie-notification/test/visual/pie-notification.spec.ts
```diff
@@ -16,6 +16,7 @@ import { IconHeartFilled } from '@justeattakeaway/pie-icons-webc';
 import {
     variants,
     headingLevels,
+    positions,
```
jamieomaguire, line 19: @thejfreitas one thing I have noticed is we don't have any visual tests that set the component to right to left. 

Please could you take a look at the [pie-input visual tests](https://github.com/justeattakeaway/pie/blob/4fca85052c6d9a334346128c870ef13643888a19/packages/components/pie-input/test/visual/pie-input.spec.ts#L16) to see an example of how we can set the page to RTL in order to test it?

siggerzz (dismissed):

packages/components/pie-notification/src/index.ts
thejfreitas: To make it explicit that in the component's class scope this property is not nullable and all internal methods will assume that this prop will have a value regardless if is optional for our users.

packages/components/pie-notification/test/component/pie-notification.spec.ts
thejfreitas: That's my intention if somehow the value/token changes for whatever reason this will make sure that we need to update the token to a new one in this component. I had this scenario in the past in PIE React and as I said above IMHO they complement each other but I am ok to remove it.

@ashleynolan @xander-marjoram @jamieomaguire please react with 👍🏼 to keep or 👎🏼 to remove this test I am fine with both options.

packages/components/pie-notification/test/visual/pie-notification.spec.ts
thejfreitas: I believe this prop is not related to reading direction as well don't interfere on the reading experience, however I took the opportunity to provide a visual test for this case scenario. As well I didn't provided a RTL test for each prop because I didn't wanted to double the amount of visual tests since Percy is kinda limited of the number of screenshots. If you think that this component deserves more RTL visual tests I believe that a new ticket would fit better for this purpose. :) 

Screenshot 2024-04-26 at 11 55 49 AM

packages/components/pie-notification/test/component/pie-notification.spec.ts
xander-marjoram: Hi @thejfreitas, sorry for the slow response. My understanding is that if the tokens change then we want consumers to update automatically. I'd be in favour of removing this test as the visual test should cover it.

raoufswe (approved):

packages/components/pie-notification/src/notification.scss
```diff
@@ -38,6 +38,10 @@
         --notification-direction: row;
     }
 
+    &[position='full-width'] {
+        --notification-border-radius: var(--dt-radius-00);
```
xander-marjoram, line 42: We can update this to use an alias token in a follow-up PR 😄

xander-marjoram (approved):

xander-marjoram:
This looks good. I noticed that there isn't currently a storybook control for the new `position` prop, can this be added in another PR?

packages/components/pie-notification/src/notification.scss
raoufswe: i will add it as part of this PR https://github.com/justeattakeaway/pie/pull/1435 🫰

raoufswe:
@xander-marjoram i will add it as part of this PR https://github.com/justeattakeaway/pie/pull/1435 🤝

.changeset/honest-students-protect.md
```diff
@@ -0,0 +1,5 @@
+---
+"@justeattakeaway/pie-notification": patch
```
ashleynolan, line 2: Super small note (or NIT 😆 ) – patch releases should only be used for tiny bug fixes or one-line updates, rather than added features (which should be minor releases).

thejfreitas:
> This looks good. I noticed that there isn't currently a storybook control for the new `position` prop, can this be added in another PR?

@xander-marjoram Looks like the control were lost on some rebase/merge.

.changeset/honest-students-protect.md
thejfreitas: @ashleynolan noted, thanks for the feedback

PR #1391, opened 2024-04-11
chore(pie-monorepo): DSW-1868 add bundle visualiser to build scripts

jamieomaguire:
Adds a rollup bundle visualiser that runs during the build of our components, webc, webc-core and webc-icons packages. In future we may want to limit this to only run locally and not during CI. However it seemed like it would involve adding more environment context to our build process which I think is probably out of scope for this change.

### Example outputs

Screenshot 2024-04-11 at 13 19 25
Screenshot 2024-04-11 at 13 18 58
Screenshot 2024-04-11 at 13 18 42

### DIFF
--- a/.changeset/ninety-boats-yawn.md
+++ b/.changeset/ninety-boats-yawn.md
@@ -1,0 +1,8 @@
+---
+"@justeattakeaway/pie-webc-core": minor
+"@justeattakeaway/pie-components-config": minor
+"@justeattakeaway/pie-icons-webc": minor
+"pie-monorepo": minor
+---
+
+[Added] - Bundle visualiser that runs during build for webc icons, webc core and our components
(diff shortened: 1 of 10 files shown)


.changeset/ninety-boats-yawn.md
```diff
@@ -0,0 +1,8 @@
+---
+"@justeattakeaway/pie-webc-core": minor
+"@justeattakeaway/pie-components-config": minor
```
xander-marjoram, line 3: Do we need a patch bump for the components so that this is picked up?

jamieomaguire: I'm unsure, this doesn't give consumers anything so I think not?

xander-marjoram (approved):

maledr5 (approved):

ashleynolan (approved):

PR #1429, opened 2024-05-02
ci(pie-monorepo): DSW-000 filter internal packages from /test-aperture

siggerzz:
This change to the /test-aperture logic ensures that packages that are no intended to be used by consuming applications don't be added as part of the automatically generated PR's.

### DIFF
--- a/.github/workflows/changeset-snapshot/test-aperture.js
+++ b/.github/workflows/changeset-snapshot/test-aperture.js
@@ -38,8 +38,7 @@ module.exports = async ({ github, context }, execa) => {
                 repo: 'pie-aperture',
                 event_type: 'pie-trigger',
                 client_payload: {
-                  'pie-branch': process.env.PIE_BRANCH,
-                  'pie-pr-number': process.env.PIE_PR_NUMBER,
+                  'pie-branch': process.env.GITHUB_REF_NAME,
                   'snapshot-version': snapshotVersion,
                   'snapshot-packages': packageNames.join(' ')
                 }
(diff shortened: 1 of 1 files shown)


.github/workflows/changeset-snapshot/test-aperture.js
```diff
@@ -9,7 +9,7 @@ module.exports = async ({ github, context }, execa) => {
     const newTags = Array
         .from(stdout.matchAll(/New tag:\s+([^\s\n]+)/g))
         .map(([_, tag]) => tag)
-        .filter((tag) => !/^wc-.+$|pie-(monorepo|docs|storybook)/.test(tag));
+        .filter((tag) => !/^wc-.+$|pie-(monorepo|docs|storybook|git-hooks-scripts|webc-core|webc-testing|icons-configs|wrapper-react)|generator-pie-component/.test(tag));
```
jamieomaguire, line 12: Its fine for now but i think this is a bit brittle (like we discussed yesterday for the web-c automation script). Do you have any ideas on how we could avoid having to remember to update this list?

siggerzz: Completely agree. As discussed yesterday, I think adding an extra property to the pieMetadata in package.json would probably make the most sense. We could also take another look what which packages are private/public, but there's a risk this will have a knock-on effect with package resolutions.

siggerzz:
Closing this PR and will refactor the logic once pie-webc is integrated into aperture.

PR #1451, opened 2024-05-16
feat(pie-monorepo): DSW-1644 provide defaults object for components props

fernandofranca:

[Added] - Generic type for components default props
[Changed] - components to use a single source of truth for props default values
[Changed] - component stories to use a single source of truth for props default values

### Reviewer 1


packages/components/pie-cookie-banner/src/index.ts
```diff
@@ -27,14 +29,25 @@ import {
...
+// TODO: Move this object to defs.ts
+// It can be moved as soon as we manage to figure a safe way to avoid the error TS2821
+// Import assertions are only supported when the '--module' option is set to 'esnext' or 'nodenext'
+export const defaultProps: Readonly<DefaultProps> = {
+    hasPrimaryActionsOnly: false,
+    defaultPreferences: {},
+    locale: defaultLocale,
+    cookieStatementLink: '',
+    cookieTechnologiesLink: '',
+};
```
fernandofranca, line 49: Browser tests for `pie-cookie-banner` fail when they find the import for the default JSON locale, **ONLY IF** it is in the `defs.ts` file. When the import is in the `index.ts` file the error doesn't happen. 😄

What is also peculiar is the error happening only in browser tests, never in compilation...

```
TypeError: Module "file:///.../pie/packages/components/pie-cookie-banner/locales/en-gb.json" needs an import assertion of type "json"
```

As suggested by the error message, a solution is to add `assert { type: 'json' };` to the import, however it will cause a TS error that requires to update our `tsconfig` `module` to `ESNext` (currently we use `ES2022`).

I'm not entirely sure if this is the reasonable or safe solution, so at the cost of consistency, I kept the JSON import in `index.ts`.

If we decide to go with `tsconfig` changes or some other settings update to browser tests, I'd suggest to add to follow-up PR so we avoid scope creep.

apps/pie-storybook/stories/pie-assistive-text.stories.ts
```diff
@@ -49,7 +49,7 @@ const Template : TemplateFunction<AssistiveTextProps> = ({
     slot,
 }) => html`
     <pie-assistive-text
-        variant="${variant}">
+        variant="${variant || defaultProps.variant}">
```
xander-marjoram, line 52: Would we be able to use something like `ifDefined` here instead? I would expect the assistive text component to handle the default variant if it isn't provided
```suggestion
        variant="${ifDefined(variant)}">
```

packages/components/pie-tag/src/index.ts
```diff
@@ -18,18 +20,18 @@ const componentSelector = 'pie-tag';
  */
 export class PieTag extends LitElement implements TagProps {
     @property({ type: String })
-    @validPropertyValues(componentSelector, variants, 'neutral')
-    public variant: TagProps['variant'] = 'neutral';
+    @validPropertyValues(componentSelector, variants, defaultProps.variant)
```
xander-marjoram, line 23: This is much better, nice one

apps/pie-storybook/stories/pie-assistive-text.stories.ts
xander-marjoram: Also I noticed that in pie-card's story file, the HTML template is missing either `defaultProps` or `isDefined`

packages/components/pie-card/src/defs.ts
```diff
@@ -58,3 +60,12 @@ export interface CardProps {
      */
     padding?: PaddingValue | `${PaddingValue},${PaddingValue}`;
 }
+
+export type DefaultProps = ComponentDefaultPropsGeneric<CardProps, 'tag' | 'variant' | 'disabled' | 'isDraggable'>;
+
+export const defaultProps: Readonly<DefaultProps> = {
```
xander-marjoram, line 66: I notice that we're using `Readonly` here. It is also applied in the definition of `ComponentDefaultPropsGeneric`: https://github.com/justeattakeaway/pie/pull/1451/files#diff-1e0d8ad270e50cba2b32b4be90b4ae7916c97dd38197a058ded70b2e05fda04d

Is it necessary to have `Readonly` in both places?

packages/components/pie-divider/src/index.ts
```diff
@@ -29,8 +32,8 @@ export class PieDivider extends LitElement implements DividerProps {
                 data-test-id="pie-divider"
                 aria-hidden="true"
                 class="c-divider"
-                variant=${variant || 'default'}
-                orientation=${orientation || 'horizontal'}
+                variant=${ifDefined(variant || defaultProps.variant)}
```
xander-marjoram, line 35: Out of interest, was there a warning in your editor that meant you needed to add `ifDefined` here?

packages/components/pie-notification/src/index.ts
```diff
@@ -46,28 +47,28 @@ export class PieNotification extends LitElement implements NotificationProps {
     public isOpen = true;
 
     @property()
-    @validPropertyValues(componentSelector, variants, 'neutral')
-    public variant: NonNullable<NotificationProps['variant']> = 'neutral';
+    @validPropertyValues(componentSelector, variants, defaultProps.variant)
+    public variant: NonNullable<NotificationProps['variant']> = defaultProps.variant;
```
xander-marjoram, line 51: I think `isOpen` also needs updating

apps/pie-storybook/stories/pie-assistive-text.stories.ts
fernandofranca: Well noticed! 👍

packages/components/pie-card/src/defs.ts
fernandofranca: Fixed the leftovers 👍

packages/components/pie-divider/src/index.ts
fernandofranca: Nope, truly excessive indeed 🤣

apps/pie-storybook/stories/pie-button.stories.ts
```diff
@@ -190,9 +185,9 @@ const Template: TemplateFunction<ButtonProps> = ({
     responsiveSize,
 }) => html`
 <pie-button
-    size="${size}"
-    variant="${variant}"
-    type="${type}"
+    size="${size || defaultProps.size}"
```
xander-marjoram, line 188: I think we should use `ifDefined` in all of these stories, like you've done for assistive-text above, sorry if I wasn't clear with my previous comment!

packages/components/pie-checkbox/src/index.ts
```diff
@@ -87,7 +87,7 @@ export class PieCheckbox extends RtlMixin(LitElement) implements CheckboxProps {
                 name=${ifDefined(name)}
                 ?disabled=${disabled}
                 ?required=${required}
-                .indeterminate=${indeterminate}
+                .indeterminate=${indeterminate || defaultProps.indeterminate}
```
xander-marjoram, line 90: I'm unsure if this is necessary, given we also set the default on L46

packages/components/pie-chip/src/index.ts
```diff
@@ -98,15 +100,15 @@ export class PieChip extends LitElement implements ChipProps {
...
+                aria-busy="${ifDefined(isLoading)}"
+                aria-current="${ifDefined(isSelected)}"
                 aria-label="${ifDefined(this.aria?.label)}"
                 aria-live="polite"
                 class="c-chip"
                 data-test-id="pie-chip"
                 tabindex="0"
                 role="button"
-                variant="${variant}"
+                ?variant="${variant}"
```
xander-marjoram, line 111: I think `?` is for binding boolean attributes

packages/components/pie-divider/src/index.ts
```diff
@@ -29,8 +32,8 @@ export class PieDivider extends LitElement implements DividerProps {
                 data-test-id="pie-divider"
                 aria-hidden="true"
                 class="c-divider"
-                variant=${variant || 'default'}
-                orientation=${orientation || 'horizontal'}
+                variant=${ifDefined(variant)}
+                orientation=${ifDefined(orientation)}
```
xander-marjoram, line 36: This is probably okay, but I think `variant` and `orientation` should always be defined because we're giving them the default values on L21 and L25

apps/pie-storybook/stories/pie-button.stories.ts
xander-marjoram: Just to elaborate, because I won't be around to answer questions on Monday (if you have them).

In storybook, we want to be able to test that components apply default values properly when we don't pass in the prop. So in this example, if we use the storybook control to set `size` to `undefined`, `ifDefined` wouldn't pass in a value at all, and therefore the component has to decide which value to use, falling back to the default.

packages/components/pie-checkbox/src/index.ts
xander-marjoram: And the same comment for the spinner component
https://github.com/justeattakeaway/pie/pull/1451/files#diff-766e0e088420ac03dd430a518fb113540ebc722ea339697d32c2ebb342015bddR44-R45

packages/components/pie-notification/src/index.ts
```diff
@@ -43,31 +44,31 @@ export * from './defs';
...
+    public isDismissible = defaultProps.isDismissible;
 
     @property({ type: Boolean })
-    public isCompact = false;
+    public isCompact = defaultProps.isCompact;
 
     @property({ type: String })
     public heading!: string;
 
     @property()
```
jamieomaguire, line 66: I know this isn't related to your change, but should this property decorator have an arg of `{ type: String }` inside?

jamieomaguire (approved):

packages/components/pie-divider/src/index.ts
```diff
@@ -1,8 +1,11 @@
 import { LitElement, html, unsafeCSS } from 'lit';
 import { property } from 'lit/decorators.js';
+import { ifDefined } from 'lit/directives/if-defined.js';
```
xander-marjoram, line 3: Looks like this import can be removed now

apps/pie-storybook/stories/pie-modal.stories.ts
```diff
@@ -192,7 +187,7 @@ const BaseStoryTemplate = (props: ModalProps) : TemplateResult => {
         <pie-modal
             .aria="${aria}"
             heading="${heading}"
-            headingLevel="${headingLevel}"
+            headingLevel="${headingLevel || defaultProps.headingLevel}"
```
xander-marjoram, line 190:
```suggestion
            headingLevel="${ifDefined(headingLevel)}"
```

packages/components/pie-notification/src/index.ts
fernandofranca: Should I add an empty `heading` default prop and have that `!` removed?

jamieomaguire (approved):

xander-marjoram (approved):

PR #1504, opened 2024-06-07
feat(pie-checkbox): DSW-2082 add defaultChecked property

xander-marjoram:

- [Added] - `defaultChecked` property.
- [Changed] - Clarified usage of `checked` property. This was being used for both the initial and current states.
- [Added] - Form reset behaviour. If the checkbox is associated with a form and the form is reset, `checked` is updated to match the value of `defaultChecked`.

### Reviewer 1 - Kev

### Reviewer 2


apps/pie-storybook/stories/pie-checkbox.stories.ts
```diff
@@ -13,10 +13,11 @@ type CheckboxStoryMeta = StoryMeta<CheckboxProps>;
 
 const defaultArgs: CheckboxProps = {
     ...defaultProps,
-    name: '',
+    name: 'name',
```
xander-marjoram, line 16: This is needed for the form association to work in storybook, it took me a while to work out why it was broken!

packages/components/pie-checkbox/src/index.ts
```diff
@@ -51,7 +53,7 @@ export class PieCheckbox extends FormControlMixin(RtlMixin(LitElement)) implemen
     public aria: CheckboxProps['aria'];
 
     @query('input[type="checkbox"]')
-    private checkbox?: HTMLInputElement;
+    public checkbox!: HTMLInputElement;
```
xander-marjoram, line 56: I made it public to help with the tests, but I'm happy to change it back if we like, it's not super important.

Regarding the `!`, I think it's a safe assumption that the input element will always exist, otherwise something has gone extremely wrong!

packages/components/pie-checkbox/src/index.ts
```diff
@@ -125,11 +137,11 @@ export class PieCheckbox extends FormControlMixin(RtlMixin(LitElement)) implemen
             <input
                 type="checkbox"
                 .value=${value}
-                ?checked=${live(checked)}
+                .checked=${!!checked}
```
xander-marjoram, line 140: The `checked` _attribute_ (set in Lit using `?checked`) refers to the initial state for the checkbox ([MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#checked)), whereas the `checked` _property_ (the same as `document.querySelector('input[type="checkbox"]').checked`, set using `.checked` controls the actual checked state.

Took me a while to work that one out!

packages/components/pie-checkbox/src/index.ts
```diff
@@ -125,11 +137,11 @@ export class PieCheckbox extends FormControlMixin(RtlMixin(LitElement)) implemen
...
                 name=${ifDefined(name)}
                 ?disabled=${disabled}
                 ?required=${required}
-                .indeterminate=${indeterminate}
+                .indeterminate=${!!indeterminate}
```
xander-marjoram, line 144: There was an IDE warning that `boolean | undefined` is not assignable to `boolean`.

apps/pie-storybook/stories/pie-checkbox.stories.ts
dandel10n: My bad, sorry. I wanted to update with my latest changes and forgot

packages/components/pie-checkbox/src/index.ts
```diff
@@ -36,7 +35,10 @@ export class PieCheckbox extends FormControlMixin(RtlMixin(LitElement)) implemen
     public name?: CheckboxProps['name'];
 
     @property({ type: Boolean, reflect: true })
-    public checked = defaultProps.checked;
```
dandel10n, line 39: Question so I can understand better: I think I passed it this way because of the lint error "`boolean | undefined` is not assignable to boolean." you mentioned below. You fixed it by bringing it to boolean via `!!`.  I guess we never should use direct referencing to defaultProps like I did here, right?

dandel10n (commented):
Great work Xander 👍 I left few comments more for me to better understand

apps/pie-storybook/stories/pie-checkbox.stories.ts
```diff
@@ -112,10 +127,11 @@ const Template = ({
 
     return html`
         <pie-checkbox
-            .value="${ifDefined(value)}"
```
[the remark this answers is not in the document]
xander-marjoram: I removed this because there was an IDE warning that `ifDefined` has no effect here. It looks like it should only be used for attribute bindings, and this is a property binding:

https://github.com/runem/lit-analyzer/blob/master/docs/readme/rules.md#-no-invalid-directive-binding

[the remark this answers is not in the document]
xander-marjoram: I just double checked and even with `ifDefined` it submits an empty string if you remove the `"on"` from the storybook control. Looks like we'll have to find another way.

apps/pie-storybook/stories/pie-checkbox.stories.ts
```diff
@@ -143,10 +160,11 @@ const ExampleFormTemplate: TemplateFunction<CheckboxProps> = ({
     return html`
     <form id="testForm">
         <pie-checkbox
-            .value="${ifDefined(value)}"
```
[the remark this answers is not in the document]
xander-marjoram: Discussed in the other thread

packages/components/pie-checkbox/src/index.ts
xander-marjoram: @dandel10n It's a good point. I think the issue is that we insist on making our properties optional in the TypeScript types, so if we pass that through into the component we are forced to resolve the possible undefined (even though we know it will always have a value because we assign a default).

I think there are two ways to deal with this:
1. Don't use `CheckboxProps` in this file to force `boolean` props to `boolean | undefined`; as long as we provide a default value it should never be undefined inside the component.
2. Continue to use `CheckboxProps`, but use `Boolean()` or `!!` to make TypeScript see that it will always have a value.

I think I would prefer option number 1 because it's a bit cleaner. We also don't lose JSDoc comments as those are still picked up because the class `implements CheckboxProps`.

As you're away this week, I'll revert it and we can discuss aligning the rest of the props when you're back.

packages/components/pie-checkbox/src/index.ts
xander-marjoram: I think this can be addressed as part of the other discussion: https://github.com/justeattakeaway/pie/pull/1504/files#r1632938723

apps/pie-storybook/stories/pie-checkbox.stories.ts
```diff
@@ -34,47 +35,60 @@ const checkboxStoryMeta: CheckboxStoryMeta = {
...
-            description: 'Indicates whether or not the checkbox is checked.',
+            description: 'The checked state of the checkbox.',
             control: 'boolean',
             defaultValue: {
                 summary: defaultArgs.checked,
             },
         },
+
+        defaultChecked: {
+            description: 'The default checked state of the checkbox.',
```
ashleynolan, line 64: Is it worth being totally explicit, so that:

```suggestion
            description: 'The default checked state of the checkbox (does not reflect if the checkbox is currently checked).',
```

packages/components/pie-checkbox/src/index.ts
```diff
@@ -108,6 +111,23 @@ export class PieCheckbox extends FormControlMixin(RtlMixin(LitElement)) implemen
...
+     */
+    public formResetCallback () : void {
+        const oldValue = this.checked;
+        this.checked = this.defaultChecked;
+
+        // If the value has changed as part of the reset, emit a change event
+        if (oldValue !== this.checked) {
+            const changeEvent = new Event('change', { bubbles: true, composed: true });
+            this.dispatchEvent(changeEvent);
+        }
```
ashleynolan, line 126: Not much of a difference, but may be cleaner?

```suggestion
        // If the current value is different to the default, update it and emit a change event
        if (this.checked !== this.defaultChecked) {
            this.checked = this.defaultChecked;
            const changeEvent = new Event('change', { bubbles: true, composed: true });
            this.dispatchEvent(changeEvent);
        }
```

ashleynolan (commented):
Looks good – couple of very small comments, otherwise 🚢

packages/components/pie-checkbox/src/index.ts
xander-marjoram: Nice, I like it. I actually went one step further and did an early return if `this.checked === this.defaultChecked`.

apps/pie-storybook/stories/pie-checkbox.stories.ts
xander-marjoram: Updated to be _even more_ explicit 😄

packages/components/pie-checkbox/test/component/pie-checkbox.spec.ts
```diff
@@ -86,6 +86,55 @@ test.describe('PieCheckbox - Component tests', () => {
...
+                        checked: true,
+                    } as CheckboxProps,
+                });
+
+                // Act
+                const checkbox = component.locator(componentSelector);
+                const isChecked = await page.evaluate(() => document.querySelector('pie-checkbox')?.checkbox?.checked);
+
+                // Assert
+                expect(await checkbox.isChecked()).toBe(true);
```
kevinrodrigues, line 133: Do we need to verify that the prop works as intended here as part of this & the above test? I think we probably don't need to IMO because the next line would fail if the prop wasn't being applied correctly to the component.

kevinrodrigues (dismissed):

kevinrodrigues (approved):

ashleynolan (approved):

PR #1574, opened 2024-07-03
feat(pie-textarea): DSW-2130 add resize prop and functionality

xander-marjoram:
To-do
- [x] Default heights with parameterisation
- [x] Minimum height for manual mode
- [x] Disable horizontal resize in manual mode 
- [x] Disable manual resize in auto mode 
- [x] Minimum height for auto resize mode
- [x] Automatic height adjustment in auto mode
- [x] Browser tests
- [x] Visual tests
- [x] Readme update
- [x] Mobile-specific fixed heights?

Resize handle re-styling - Won't do. Will create a new ticket if required.

- [Added] - resize functionality for textarea
- [Changed] - update textarea storybook story

- [x] I have reviewed visual test updates properly before approving them
Tested in Storybook. Aperture pages will be added in DSW-2214.

### Reviewer 1 @leksaBoiko 

### Reviewer 2 @kevinrodrigues 


packages/components/pie-textarea/src/index.ts
```diff
@@ -23,21 +26,44 @@ export class PieTextarea extends RtlMixin(LitElement) implements TextareaProps {
...
-                class="c-textarea"
-                data-test-id="pie-textarea-shell"
-                data-pie-size=${ifDefined(size)}>
+                class="c-textareaWrapper"
+                data-test-id="pie-textarea-wrapper"
+                data-pie-size="${size}"
+                data-pie-resize="${resize}">
                 <textarea
                     data-test-id="pie-textarea"
+                    @input=${this.handleResize}
```
xander-marjoram, line 66: Open question about how often we want this to be triggered. I tried `@change` but it felt too unresponsive because it waits for focus to leave the textarea before resizing.

packages/components/pie-textarea/src/index.ts
```diff
@@ -23,21 +26,44 @@ export class PieTextarea extends RtlMixin(LitElement) implements TextareaProps {
 
     @property({ type: String })
     @validPropertyValues(componentSelector, sizes, defaultProps.size)
-    public size?: TextareaProps['size'] = defaultProps.size;
+    public size = defaultProps.size;
+
```
leksaBoiko, line 30: I wonder if you added `disabled` prop to `defaultProps`, maybe you can describe `disable` prop in  the same way?

```
@property({ type: Boolean, reflect: true })
public disabled = defaultProps.disabled;
```

xander-marjoram: Ah yes, I must have missed that. Good idea.

packages/components/pie-textarea/src/index.ts
```diff
@@ -19,25 +22,48 @@ export class PieTextarea extends RtlMixin(LitElement) implements TextareaProps {
     static shadowRootOptions = { ...LitElement.shadowRootOptions, delegatesFocus: true };
 
     @property({ type: Boolean, reflect: true })
```
raoufswe, line 24: do you know why we are reflecting the disabled prop here while we don't in some other places?

packages/components/pie-textarea/src/index.ts
```diff
@@ -19,25 +22,48 @@ export class PieTextarea extends RtlMixin(LitElement) implements TextareaProps {
...
         } = this;
 
         return html`
             <div
-                class="c-textarea"
-                data-test-id="pie-textarea-shell"
-                data-pie-size=${ifDefined(size)}>
+                class="c-textareaWrapper"
+                data-test-id="pie-textarea-wrapper"
+                data-pie-size="${size}"
```
raoufswe, line 62: i think we usually just assign the prop name to the element such as size and resize. 
has this changed?

xander-marjoram: Do you mean `size="${size}"` instead? I think @ashleynolan was saying that technically this is invalid HTML because `size` is not a valid HTML attribute for the `textarea` element, so we want to avoid potentially being punished in terms of SEO if they start checking the HTML inside the shadow DOM.

packages/components/pie-textarea/src/index.ts
xander-marjoram: Good question. I just looked and it seems that in all of the form components (checkbox, switch, text input and button) we are reflecting. I suppose this is to make sure that if something externally changes the disabled state of the component that this can be noticed by the form.

packages/components/pie-textarea/src/index.ts
leksaBoiko: just wondering, does it make sense to change `size` props to `data-pie-size` inside `pie-modal` and `pie-spinner` components:

https://github.com/justeattakeaway/pie/blob/96ea944afa3526717ca253f66057823afa95839d/packages/components/pie-spinner/src/index.ts#L44
https://github.com/justeattakeaway/pie/blob/96ea944afa3526717ca253f66057823afa95839d/packages/components/pie-modal/src/index.ts#L404

xander-marjoram: Yes, I think so. Shall I add it as an AC to the ticket for aligning props and types?

leksaBoiko: yes, I think DSW-2191 is fine
thank you

leksaBoiko (dismissed):

leksaBoiko (dismissed):

packages/components/pie-textarea/src/index.ts
```diff
@@ -19,25 +22,48 @@ export class PieTextarea extends RtlMixin(LitElement) implements TextareaProps {
...
+    public size = defaultProps.size;
+
+    @property({ type: String })
+    @validPropertyValues(componentSelector, resizeModes, defaultProps.resize)
+    public resize = defaultProps.resize;
+
+    @query('textarea')
+    private _textarea!: HTMLTextAreaElement;
+
+    private handleResize () {
```
kevinrodrigues, line 38: Should this be debounced / limited?

xander-marjoram: Yeah it's an interesting question. I don't think debounce would work very well because it can feel unresponsive as you're typing. Throttling might work if we get the timing right. Maybe something like 100-200ms.

kevinrodrigues: Yup throttling is the one i always mix those up. 👍

packages/components/pie-textarea/src/textarea.scss
```diff
@@ -1,22 +1,89 @@
...
+  --textarea-content-color: var(--dt-color-content-default);
 
-  height: var(--textarea-height);
+  // Default height is two lines of text
+  --textarea-height: calc((var(--textarea-line-height) * 2) + (var(--textarea-padding-block) * 2) + (var(--textarea-border-thickness) * 2));
+
+  line-height: 0; // Remove once there is text outside the textarea
+
+  textarea {
+    font-size: calc(var(--dt-font-body-l-size) * 1px);
```
ashleynolan, line 24:
```suggestion
    @include p.font-size(--dt-font-body-l-size);
```

packages/components/pie-textarea/src/textarea.scss
```diff
@@ -1,22 +1,89 @@
...
 
-.c-textarea {
-  --textarea-height: 72px;
+// Heights are being defined based on the line height of the text and the padding.
+// Changing the `size` property affects the padding and therefore the height of the textarea.
+// Default height is two lines of text.
+// Minimum height in manual resize mode is one line of text.
+// Maximum height in auto resize mode is six lines of text.
+.c-textareaWrapper {
+  --textarea-line-height: calc(var(--dt-font-body-l-line-height) * 1px);
```
ashleynolan, line 9:
```suggestion
  --textarea-line-height: #{p.line-height(--dt-font-body-l-line-height)};
```

kevinrodrigues (approved):

leksaBoiko (approved):

kevinrodrigues (approved):

PR #1581, opened 2024-07-08
feat(pie-checkbox-group): DSW-2182 main component functionality

dandel10n:
**"@justeattakeaway/pie-checkbox-group": minor
"pie-storybook": patch**
---

[Added] - main logic for checkbox-group

---
"@justeattakeaway/pie-checkbox": minor
---

[Added] - event listener for pie-checkbox-group-disabled custom event
[Removed] - aria prop since aria attributes should be passed to the component directly

**Note:** css styles along with Percy tests update will come in a separate PR

The Aperture PR: https://github.com/justeattakeaway/pie-aperture/pull/176

- [x] I have added thorough tests where applicable (component)

### Reviewer 2

### DIFF
--- a/apps/pie-storybook/stories/pie-checkbox-group.stories.ts
+++ b/apps/pie-storybook/stories/pie-checkbox-group.stories.ts
@@ -21,8 +21,12 @@ const checkboxGroupStoryMeta: CheckboxGroupStoryMeta = {
     title: 'Checkbox Group',
     component: 'pie-checkbox-group',
     argTypes: {
+        name: {
+            description: 'The name associated with the group.',
+            control: 'text',
+        },
         label: {
-            description: 'The visible label for the checkbox group',
+            description: 'The visible label for the checkbox group.',
             control: 'text',
         },
         status: {
@@ -60,6 +64,7 @@ const checkboxGroupStoryMeta: CheckboxGroupStoryMeta = {
 export default checkboxGroupStoryMeta;
 
 const Template = ({
+    name,
     label,
     assistiveText,
     status,
@@ -68,6 +73,7 @@ const Template = ({
     <p>Please note, the checkboxes are separate components. See
     <pie-link href="/?path=/story/checkbox--default">pie-checkbox</pie-link>.</p>
     <pie-checkbox-group
+        name="${ifDefined(name)}"
         label="${ifDefined(label)}"
         assistiveText="${ifDefined(assistiveText)}"
         status=${ifDefined(status)}
--- a/apps/pie-storybook/stories/pie-checkbox-group.stories.ts
+++ b/apps/pie-storybook/stories/pie-checkbox-group.stories.ts
@@ -56,7 +56,7 @@ const checkboxGroupStoryMeta: CheckboxGroupStoryMeta = {
     parameters: {
         design: {
             type: 'figma',
-            url: 'https://www.figma.com/design/aD4m0j97Ruw8Q4S5lED2Bl/Checkbox-audit?node-id=9938-14138&t=1V1VEYmtArFih6AX-4',
+            url: 'https://www.figma.com/file/aD4m0j97Ruw8Q4S5lED2Bl/Checkbox-audit?node-id=9938-14138&t=1V1VEYmtArFih6AX-4',
         },
     },
 };
--- a/apps/pie-storybook/stories/pie-checkbox-group.stories.ts
+++ b/apps/pie-storybook/stories/pie-checkbox-group.stories.ts
@@ -1,8 +1,8 @@
 import { html } from 'lit';
-
+import { ifDefined } from 'lit/directives/if-defined.js';
 /* eslint-disable import/no-duplicates */
 import '@justeattakeaway/pie-checkbox-group';
-import { CheckboxGroupProps } from '@justeattakeaway/pie-checkbox-group';
+import { CheckboxGroupProps, defaultProps, statusTypes } from '@justeattakeaway/pie-checkbox-group';
 /* eslint-enable import/no-duplicates */
 
 import { type StoryMeta } from '../types';
@@ -10,27 +10,88 @@ import { createStory } from '../utilities';
 
 type CheckboxGroupStoryMeta = StoryMeta<CheckboxGroupProps>;
 
-const defaultArgs: CheckboxGroupProps = {};
+const defaultArgs: CheckboxGroupProps = {
+    ...defaultProps,
+    aria: {
+        label: '',
+        labelledby: '',
+    },
+    disabled: false,
+};
 
 const checkboxGroupStoryMeta: CheckboxGroupStoryMeta = {
     title: 'Checkbox Group',
     component: 'pie-checkbox-group',
-    argTypes: {},
+    argTypes: {
+        label: {
+            description: 'The visible label for the checkbox group',
+            control: 'text',
+        },
+        status: {
+            description: 'The status of the checkbox group component / assistive text. Can be default, success or error.',
+            control: 'select',
+            options: statusTypes,
+            defaultValue: {
+                summary: defaultProps.status,
+            },
+        },
+        disabled: {
+            description: 'If true, disables the whole checkbox group. If you need to disable only one checkbox pass the disabled prop to the needed checkbox.',
+            control: 'boolean',
+            defaultValue: {
+                summary: defaultArgs.disabled,
+            },
+        },
+        assistiveText: {
+            description: 'An optional assistive text to display below the checkbox group.',
+            control: 'text',
+            defaultValue: {
+                summary: '',
+            },
+        },
+        aria: {
+            description: 'ARIA object to pass label/labelledby/describedby aria attributes',
+            control: 'object',
+        },
+    },
     args: defaultArgs,
     parameters: {
         design: {
             type: 'figma',
-            url: '',
+            url: 'https://www.figma.com/design/aD4m0j97Ruw8Q4S5lED2Bl/Checkbox-audit?node-id=9938-14138&t=1V1VEYmtArFih6AX-4',
         },
     },
 };
 
 export default checkboxGroupStoryMeta;
 
-// TODO: remove the eslint-disable rule when props are added
-// eslint-disable-next-line no-empty-pattern
-const Template = ({}: CheckboxGroupProps) => html`
-    <pie-checkbox-group></pie-checkbox-group>
+const Template = ({
+    label,
+    assistiveText,
+    status,
+    disabled,
+    aria,
+}: CheckboxGroupProps) => html`
+    <pie-checkbox-group
+        label="${ifDefined(label)}"
+        .aria="${aria}"
+        assistiveText="${ifDefined(assistiveText)}"
+        status=${ifDefined(status)}
+        ?disabled="${disabled}"
+    >
+        <pie-checkbox
+            name="checkbox-one"
+            label="checkbox 1">
+        </pie-checkbox>
+        <pie-checkbox
+            name="checkbox-two"
+            label="checkbox 2">
+        </pie-checkbox>
+        <pie-checkbox
+            name="checkbox-three"
+            label="checkbox 3">
+        </pie-checkbox>
+    </pie-checkbox-group>
 `;
 
 export const Default = createStory<CheckboxGroupProps>(Template, defaultArgs)();
--- a/apps/pie-storybook/stories/pie-checkbox-group.stories.ts
+++ b/apps/pie-storybook/stories/pie-checkbox-group.stories.ts
@@ -12,10 +12,6 @@ type CheckboxGroupStoryMeta = StoryMeta<CheckboxGroupProps>;
 
 const defaultArgs: CheckboxGroupProps = {
     ...defaultProps,
-    aria: {
-        label: '',
-        labelledby: '',
-    },
     disabled: false,
 };
 
@@ -49,10 +45,6 @@ const checkboxGroupStoryMeta: CheckboxGroupStoryMeta = {
                 summary: '',
             },
         },
-        aria: {
-            description: 'ARIA object to pass label/labelledby/describedby aria attributes',
-            control: 'object',
-        },
     },
     args: defaultArgs,
     parameters: {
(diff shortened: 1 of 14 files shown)


packages/components/pie-checkbox-group/src/defs.ts
```diff
@@ -1,3 +1,44 @@
...
+     * The status of the checkbox component / assistive text. Can be default, success or error.
+     */
+    status?: typeof statusTypes[number];
+}
+
+export type DefaultProps = ComponentDefaultPropsGeneric<CheckboxGroupProps, 'status'>;
+
+export const defaultProps: DefaultProps = {
+    // a default value for the html <input type="checkbox" /> value attribute.
+    // https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#value
```
xander-marjoram, line 42: I think this comment refers to the `value` prop so I'm not sure if it's needed here?

packages/components/pie-checkbox-group/src/index.ts
```diff
@@ -1,20 +1,97 @@
...
+            disabled,
+            aria,
+        } = this;
+        return html`
+            <fieldset
+                ?disabled=${disabled}
+                aria-labelledby=${label ? nothing : aria?.labelledby || nothing}
+                aria-label=${aria?.label || nothing}
+                aria-invalid=${status === 'error' ? 'true' : 'false'}
+                aria-errormessage="${status === 'error' ? assistiveTextIdValue : nothing}"
```
xander-marjoram, line 80: This prop has `""` around its value, I think the other aria ones probably should too?

packages/components/pie-checkbox-group/src/index.ts
```diff
@@ -1,20 +1,97 @@
...
+    }
+
+    private _handleDisabled () : void {
+        if (this._slottedChildren) {
+            [...this._slottedChildren]
+            .forEach((child) => child.setAttribute('disabled', 'true'));
+        }
+    }
+
+    protected firstUpdated (_changedProperties: PropertyValues<this>): void {
```
xander-marjoram, line 50: I think `updated` is called immediately after `firstUpdated`, so I wonder if we need both?

packages/components/pie-checkbox-group/test/component/pie-checkbox-group.spec.ts
```diff
@@ -1,14 +1,30 @@
...
+        const assistiveTextComponent = await mount(PieAssistiveText);
+        await assistiveTextComponent.unmount();
+    });
+
     test('should render successfully', async ({ mount, page }) => {
         // Arrange
         await mount(PieCheckboxGroup, {
             props: {} as CheckboxGroupProps,
+            slots: {
+                default: '<input type="checkbox" />',
```
xander-marjoram, line 26: Would this test work with `pie-checkbox`? Is it worth testing both?

packages/components/pie-checkbox-group/test/component/pie-checkbox-group.spec.ts
```diff
@@ -17,4 +33,101 @@ test.describe('PieCheckboxGroup - Component tests', () => {
...
+        test.describe('disabled', () => {
+            test.describe('when true', () => {
+                test('should disable the slotted input component', async ({ mount }) => {
+                    // Arrange
+                    const component = await mount(PieCheckboxGroup, {
+                        props: {
+                            disabled: true,
+                        } as CheckboxGroupProps,
+                        slots: {
+                            default: '<input type="checkbox" data-test-id="test-checkbox" />',
```
xander-marjoram, line 120: I think testing with `pie-checkbox` would be good here too. And perhaps checking that individual checkboxes can still be disabled even when `pie-checkbox-group`'s `disabled` prop is false?

packages/components/pie-checkbox-group/src/defs.ts
```diff
@@ -1,3 +1,44 @@
...
+     */
+    aria?: AriaProps;
+
+    /**
+     * The status of the checkbox component / assistive text. Can be default, success or error.
+     */
+    status?: typeof statusTypes[number];
+}
+
+export type DefaultProps = ComponentDefaultPropsGeneric<CheckboxGroupProps, 'status'>;
```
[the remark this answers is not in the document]
xander-marjoram: Also I renamed `ComponentDefaultPropsGeneric` to `ComponentDefaultProps` as part of #1582

leksaBoiko (commented):
https://github.com/justeattakeaway/pie/assets/168754213/737f6b45-7851-4ffd-9010-a7af01c14b54

packages/components/pie-checkbox-group/src/defs.ts
dandel10n: Oops

packages/components/pie-checkbox-group/src/index.ts
dandel10n: I refactored the aria attributes to follow the same approach as in Text Input component ✅

apps/pie-storybook/stories/pie-checkbox-group.stories.ts
```diff
@@ -1,36 +1,91 @@
...
+}: CheckboxGroupProps) => html`
+    <p>Please note, the checkboxes are separate components. See
+    <pie-link href="/?path=/story/checkbox--default">pie-checkbox</pie-link>.</p>
+    <pie-checkbox-group
+        label="${ifDefined(label)}"
+        assistiveText="${ifDefined(assistiveText)}"
+        status=${ifDefined(status)}
+        ?disabled="${disabled}"
+    >
+        <pie-checkbox
```
ashleynolan, line 76: Not specifically to do with this file, but I noticed when moving through focusing checkboxes, the focus can get a bit hidden by the next checkbox. Not something necessarily to fix in this PR, but I'd expect the focus to be above the checkboxes.

Screenshot 2024-07-12 at 14 10 39

packages/components/pie-checkbox-group/src/index.ts
```diff
@@ -1,20 +1,108 @@
...
 export * from './defs';
 
 const componentSelector = 'pie-checkbox-group';
+const assistiveTextId = 'assistive-text';
 
 /**
  * @tagname pie-checkbox-group
  */
-export class PieCheckboxGroup extends RtlMixin(LitElement) implements CheckboxGroupProps {
+export class PieCheckboxGroup extends FormControlMixin(RtlMixin(LitElement)) implements CheckboxGroupProps {
```
ashleynolan, line 21: Another generic question 😄 

Are we planning to handle checkbox nesting levels as part of checkbox group? So for consumers that want folder structure groupings, where the indeterminant status could be used?

packages/components/pie-checkbox/src/index.ts
```diff
@@ -147,18 +141,20 @@ export class PieCheckbox extends FormControlMixin(RtlMixin(LitElement)) implemen
             name,
             label,
             disabled,
+            disabledByParent,
```
ashleynolan, line 144: Just wondering why we need both `disabled` and `disabledByParent`? Would just using `disabled` be ok, or are there use cases where we need both?

apps/pie-storybook/stories/pie-checkbox-group.stories.ts
dandel10n: The styling for the component will come in a separate PR and there will be enough space between checkboxes 👍

packages/components/pie-checkbox/src/index.ts
dandel10n: Yes, here is a slack discussion around it: https://justeat.slack.com/archives/C05DE57E27P/p1720621318557419

Summary: for the scenarios where only one checkbox in a group should be disabled we have to separate the disabled state of the checkbox from the disabled by the checkbox group.

packages/components/pie-checkbox-group/src/index.ts
dandel10n: Not on this stage I think, since we don't have any designs for that type of the group. But we can work on it later if there will be the use case for that

packages/components/pie-checkbox/src/index.ts
ashleynolan: I guess the only thing about it being a prop, is that I see props as being an interface for consumers, but this is a prop that only really exists for `checkboxGroup` to use.

Just an idea (not sure if it would work or not), but could you handle this with events instead, so that it keeps the prop interface a bit cleaner? So when disabled is set, send an event to the checkboxes, and keep it more as a type of internal state, rather than a prop?

packages/components/pie-checkbox-group/src/index.ts
ashleynolan: I think it'd be worth bringing up with design, as I can see this being a common use-case in back-office tools (and will likely mean we just need to set the structure of the slot markup for it). Obviously a separate ticket though, just wanted to check  😄  

cc @RLumleyJET

packages/components/pie-checkbox/src/index.ts
dandel10n: @ashleynolan I rewrote the logic to use custom event dispatcher and listener

packages/components/pie-checkbox-group/src/index.ts
dandel10n: We discussed this with Rachel and for now we won't add it until we get a request for this

apps/pie-storybook/stories/pie-checkbox-group.stories.ts
```diff
@@ -1,36 +1,97 @@
...
+                summary: '',
+            },
+        },
+    },
     args: defaultArgs,
     parameters: {
         design: {
             type: 'figma',
-            url: '',
+            url: 'https://www.figma.com/design/aD4m0j97Ruw8Q4S5lED2Bl/Checkbox-audit?node-id=9938-14138&t=1V1VEYmtArFih6AX-4',
```
leksaBoiko, line 59:
```suggestion
            url: 'https://www.figma.com/file/aD4m0j97Ruw8Q4S5lED2Bl/Checkbox-audit?node-id=9938-14138&t=1V1VEYmtArFih6AX-4',
```

dandel10n: Interesting, for me the https://www.figma.com/file/aD4m0j97Ruw8Q4S5lED2Bl/Checkbox-audit?node-id=9938-14138&t=1V1VEYmtArFih6AX-4 link redirects to https://www.figma.com/design/aD4m0j97Ruw8Q4S5lED2Bl/Checkbox-audit?node-id=9938-14138&t=1V1VEYmtArFih6AX-4

sebastienrocks: `/design/` seems fine to me

sebastienrocks (approved):

apps/pie-storybook/stories/pie-checkbox-group.stories.ts
leksaBoiko: but it doesn't work for me
Screenshot 2024-07-18 at 11 17 51

packages/components/pie-checkbox-group/README.md
```diff
@@ -74,18 +75,55 @@ import { PieCheckboxGroup } from '@justeattakeaway/pie-checkbox-group/dist/react
 
 | Property | Type | Default | Description |
 |---|---|---|---|
-| - | - | - | - |
+| `name` | string | - | The name associated with the group.
+| `label` | string | - | The label value of the component |
+| assistiveText | string | - | An optional assistive text to display below the checkbox group. |
```
xander-marjoram, line 80:
```suggestion
| `assistiveText` | string | - | An optional assistive text to display below the checkbox group. |
```

apps/pie-storybook/stories/pie-checkbox-group.stories.ts
dandel10n: Oh, I see, good catch 👍

packages/components/pie-checkbox-group/README.md
```diff
@@ -74,18 +75,55 @@ import { PieCheckboxGroup } from '@justeattakeaway/pie-checkbox-group/dist/react
...
+| `disabled` | boolean | - | Same as the HTML disabled attribute - indicates whether or not the checkbox group is disabled. |
+| `assistiveText` | `string` | `''` | Allows assistive text to be displayed below the checkbox group. |
```
xander-marjoram, line 82: This is a duplicate

packages/components/pie-checkbox-group/src/index.ts
```diff
@@ -1,20 +1,113 @@
...
+    public status = defaultProps.status;
+
+    @property({ type: Boolean, reflect: true })
+    public disabled = defaultProps.disabled;
+
+    get _slot () {
+        return this.shadowRoot?.querySelector('slot');
+    }
+
+    get _slottedChildren () {
```
xander-marjoram, line 52: I think we could use the [`queryAssignedElements` decorator](https://lit.dev/docs/api/decorators/#queryAssignedElements) for this? Happy for you to do this in the next PR if you like

xander-marjoram: And maybe `@query` would work for `_slot` too

xander-marjoram: As a bonus, I think `queryAssignedElements` returns an array so you might not need to spread it below

packages/components/pie-checkbox-group/src/index.ts
```diff
@@ -1,20 +1,113 @@
...
+
+    private _handleStatus () : void {
+        if (this._slottedChildren) {
+            [...this._slottedChildren]
+            .forEach((child) => child.setAttribute('status', this.status));
+        }
+    }
+
+    protected updated (_changedProperties: PropertyValues<this>): void {
+        super.updated(_changedProperties);
```
xander-marjoram, line 74: Apparently you don't need to call `super` in `updated` https://lit.dev/docs/components/lifecycle/#updated

packages/components/pie-checkbox-group/src/index.ts
dandel10n: 👌 this is perfect

xander-marjoram (approved):

PR #1587, opened 2024-07-10
feat(pie-button): DSW-000 Changing styles to use classmap

ashleynolan:

- [Changed] - Styles now use classmap
- [Fixed] - Fixed background colour when button is in `xsmall` or `small` size and responsive mode is true, where the bg-color wasn't updating to orange as required

### Reviewer 2

### DIFF
--- a/packages/components/pie-button/src/button.scss
+++ b/packages/components/pie-button/src/button.scss
@@ -1,14 +1,13 @@
 @use '@justeattakeaway/pie-css/scss' as p;
-
-$breakpoint-wide: 768px;
+@use '@justeattakeaway/pie-css/scss/settings' as *;
 
 /**
 * Mixin for adding the responsive behaviour.
 * Takes in the the particular settings for the wide breakpoint.
 */
 @mixin responsive-wide {
-    &[isResponsive] {
-        @media (min-width: $breakpoint-wide) {
+    &.is-responsive {
+        @include media('>md') {
             @content;
         }
     }
@@ -94,12 +93,14 @@ $breakpoint-wide: 768px;
     // used to specify whether the button should be full width or not
     inline-size: var(--btn-inline-size);
 
-    // Variant
-    &[variant='primary'] {
+    // ************************
+    // Button Variant Modifiers
+    // ************************
+    &.o-btn--primary {
         @include p.button-interactive-states('--dt-color-interactive-brand');
 
-        &[size='xsmall'],
-        &[size='small-productive'] {
+        &.o-btn--xsmall,
+        &.o-btn--small-productive {
             /**
             * Where the font-size is smaller,
             * update the button backgrounds so that the text is accessible
@@ -108,16 +109,27 @@ $breakpoint-wide: 768px;
 
             @include p.button-interactive-states('--dt-color-interactive-primary', 'inverse');
         }
+
+        // This is to handle the background colour change getting overridden above, but then being put in responsive mode
+        // can't use :not for this unfortunately, as then the smaller narrow state would lose its background colour too
+        &.o-btn--xsmall.is-responsive.o-btn--expressive,
+        &.o-btn--small-productive.is-responsive {
+            @include responsive-wide {
+                --btn-bg-color: var(--dt-color-interactive-brand);
+
+                @include p.button-interactive-states('--dt-color-interactive-brand');
+            }
+        }
     }
 
-    &[variant='secondary'] {
+    &.o-btn--secondary {
         --btn-bg-color: var(--dt-color-interactive-secondary);
         --btn-text-color: var(--dt-color-content-interactive-secondary);
 
         @include p.button-interactive-states('--dt-color-interactive-secondary');
     }
 
-    &[variant='outline'] {
+    &.o-btn--outline {
         --btn-bg-color: transparent;
         --btn-text-color: var(--dt-color-content-interactive-secondary);
 
@@ -128,7 +140,7 @@ $breakpoint-wide: 768px;
         @include p.button-interactive-states('--dt-color-black', 'transparent');
     }
 
-    &[variant='ghost'] {
+    &.o-btn--ghost {
         --btn-bg-color: transparent;
         --btn-text-color: var(--dt-color-content-link);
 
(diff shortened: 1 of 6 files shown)


packages/components/pie-button/src/button.scss
```diff
@@ -108,16 +109,27 @@ $breakpoint-wide: 768px;
 
             @include p.button-interactive-states('--dt-color-interactive-primary', 'inverse');
         }
+
+        // This is to handle the background colour change getting overridden above, but then being put in responsive mode
+        // can't use :not for this unfortunately, as then the smaller narrow state would lose its background colour too
+        &.o-btn--xsmall.is-responsive.o-btn--expressive,
+        &.o-btn--small-productive.is-responsive {
```
xander-marjoram, line 116: Does `.is-responsive` already come from the mixin?
```suggestion
        &.o-btn--xsmall.o-btn--expressive,
        &.o-btn--small-productive {
```

packages/components/pie-button/src/index.ts
```diff
@@ -226,18 +226,22 @@ export class PieButton extends FormControlMixin(LitElement) implements ButtonPro
             responsiveSize,
         } = this;
 
+        const classes = {
+            'o-btn': true,
+            'o-btn--fullWidth': isFullWidth,
+            'is-responsive': isResponsive,
+            [`o-btn--${responsiveSize}`]: ((isResponsive && responsiveSize) || false),
```
xander-marjoram, line 233: Can `|| false` be removed?

packages/components/pie-button/src/index.ts
```diff
@@ -1,8 +1,8 @@
 import {
     LitElement, html, unsafeCSS, nothing, PropertyValues, TemplateResult,
 } from 'lit';
+import { classMap } from 'lit-html/directives/class-map.js';
```
xander-marjoram, line 4: This is the [suggested](https://lit.dev/docs/templates/directives/#classmap) import path, I don't think we have `lit-html` listed as a dependency.
```suggestion
import { classMap } from 'lit/directives/class-map.js';
```

packages/components/pie-button/src/button.scss
ashleynolan: Just checked, and yep it does 😆  Good spot

ashleynolan: Fixed

xander-marjoram (dismissed):

packages/components/pie-button/src/button.scss
```diff
@@ -137,15 +149,15 @@ $breakpoint-wide: 768px;
         @include p.button-interactive-states('--dt-color-black', 'transparent');
     }
 
-    &[variant='inverse'] {
+    &.o-btn--inverse {
```
jamieomaguire, line 152: It could be useful for part of this change be to document our CSS naming conventions in the wiki. Would you be happy to do that? It could be a port from Fozzie, but I don't think we currently document for web components.

ashleynolan: Yeh, sure. I'll spin this up as a separate PR 👍

packages/components/pie-button/src/index.ts
```diff
@@ -226,18 +226,22 @@ export class PieButton extends FormControlMixin(LitElement) implements ButtonPro
...
         } = this;
 
+        const classes = {
+            'o-btn': true,
+            'o-btn--fullWidth': isFullWidth,
+            'is-responsive': isResponsive,
+            [`o-btn--${responsiveSize}`]: Boolean(isResponsive && responsiveSize),
+            [`o-btn--${variant}`]: true,
+            [`o-btn--${size}`]: true,
+            'is-loading': isLoading,
```
dandel10n, line 236: Maybe we can pass only dynamic class names to the object and combine it with static values like `class="o-btn o-btn--${variant} o-btn--${size} ${classMap(classes)}"`?

packages/components/pie-button/src/button.scss
```diff
@@ -241,6 +268,9 @@ $breakpoint-wide: 768px;
         }
     }
 
+    // ************
+    // Focus styles
+    // ************
```
jamieomaguire, line 273: I am sure the answer is yes, but does vite strip these comments out?

packages/components/pie-button/src/index.ts
jamieomaguire: I think its a reasonable idea, my only argument against this would be it's nice being able to see all the class names in a single place

packages/components/pie-button/src/index.ts
```diff
@@ -226,18 +226,22 @@ export class PieButton extends FormControlMixin(LitElement) implements ButtonPro
             responsiveSize,
         } = this;
 
+        const classes = {
+            'o-btn': true,
+            'o-btn--fullWidth': isFullWidth,
+            'is-responsive': isResponsive,
```
jamieomaguire, line 232: I'm a little skeptical of the `is-responsive` class name. Is this a convention we want to establish versus utility class names?

packages/components/pie-button/src/button.scss
ashleynolan: Here you go https://github.com/justeattakeaway/pie/wiki/PIE-%E2%80%93-CSS-Naming-Scheme

packages/components/pie-button/src/index.ts
ashleynolan: That's pretty much why I ended up using one object for the classes – I preferred being able to see the different classes being set altogether. Would be good to know if people are for or against it though before we refactor in other places 😄

packages/components/pie-button/src/button.scss
ashleynolan: Yeh, they are Sass comments, so not valid CSS, so get stripped out in the compile step. Only `/* */` style comments would be left in.

packages/components/pie-button/src/index.ts
ashleynolan: I don't mind really – would you expect it to be something like `o-btn--responsive` instead?

jamieomaguire: I think so yeah, unless its we make it a util class local to this component.

jamieomaguire (approved):

xander-marjoram (approved):

PR #1621, opened 2024-07-23
feat(pie-textarea): DSW-2132 add textarea form functionality

leksaBoiko:
[Added] - textarea component form functionality

The Aperture PR: https://github.com/justeattakeaway/pie-aperture/pull/180


packages/components/pie-textarea/src/defs.ts
```diff
@@ -20,17 +20,57 @@ export interface TextareaProps {
...
+     * If true, the textarea is required to have a value before submitting the form. If there is no value, then the component validity state will be invalid.
+     */
+    required?: boolean;
 }
 
 /**
  * The default values for the `TextareaProps` that are required (i.e. they have a fallback value in the component).
  */
-type DefaultProps = ComponentDefaultProps<TextareaProps>;
+type DefaultProps = ComponentDefaultProps<TextareaProps, 'value' | 'disabled' | 'size' | 'resize'>;
```
xander-marjoram, line 67: I wonder if it might be better to `Omit` some props instead? Also some things like required, readonly, autoFocus can default to false

packages/components/pie-textarea/src/index.ts
```diff
@@ -33,6 +45,21 @@ export class PieTextarea extends RtlMixin(LitElement) implements TextareaProps {
...
+    public autoFocus?: TextareaProps['autoFocus'];
+
+    @property({ type: Boolean })
+    public readonly?: TextareaProps['readonly'];
+
+    @property({ type: String })
+    public defaultValue?: TextareaProps['defaultValue'];
+
+    @property({ type: Boolean })
+    public required?: TextareaProps['required'] = false;
```
xander-marjoram, line 61: If there is a default value that can go in `defaultProps`

packages/components/pie-textarea/src/index.ts
```diff
@@ -33,6 +45,21 @@ export class PieTextarea extends RtlMixin(LitElement) implements TextareaProps {
...
     public resize = defaultProps.resize;
 
+    @property({ type: String })
+    public autocomplete?: TextareaProps['autocomplete'];
+
+    @property({ type: Boolean })
+    public autoFocus?: TextareaProps['autoFocus'];
+
+    @property({ type: Boolean })
+    public readonly?: TextareaProps['readonly'];
```
xander-marjoram, line 55: I don't think we need `?` on the prop declarations in this file because that should already come from the interface

packages/components/pie-textarea/src/index.ts
```diff
@@ -43,21 +70,89 @@ export class PieTextarea extends RtlMixin(LitElement) implements TextareaProps {
...
+
+    /**
+     * Called when the form that owns this component is reset.
+     * Resets the value to the default value.
+     */
+    public formResetCallback (): void {
+        this.value = this.defaultValue ?? defaultProps.value;
+
+        // This ensures the input value is updated when the form is reset.
+        // Otherwise, there is a bug where values like 'e1212' for number inputs do not correctly reset.
```
xander-marjoram, line 99: I wonder if this bug also applied to textarea elements?

packages/components/pie-textarea/src/index.ts
```diff
@@ -43,21 +70,89 @@ export class PieTextarea extends RtlMixin(LitElement) implements TextareaProps {
...
+        if (this._textarea) {
+            this._textarea.value = this.value;
+        }
+
+        this._internals.setFormValue(this.value);
+    }
+
+    protected firstUpdated (_changedProperties: PropertyValues<this>): void {
+        super.firstUpdated(_changedProperties);
```
xander-marjoram, line 108: It's [not necessary](https://lit.dev/docs/components/lifecycle/#firstupdated) to call `super` inside `firstUpdated`

apps/pie-storybook/stories/pie-textarea.stories.ts
```diff
@@ -42,6 +95,41 @@ const textareaStoryMeta: TextareaStoryMeta = {
...
+            },
+        },
+        defaultValue: {
+            description: 'An optional default value to use when the textarea is reset.',
+            control: 'text',
+            defaultValue: {
+                summary: '',
+            },
+        },
+        required: {
```
xander-marjoram, line 126: I noticed that `autocomplete` and `autoFocus` are missing storybook controls. Was this intentional? I think autoFocus might be useful to have, and perhaps autocomplete inside the form story

packages/components/pie-textarea/src/index.ts
leksaBoiko: I checked other components and they `?` on the prop declarations
I followed the same code style

packages/components/pie-textarea/src/index.ts
leksaBoiko: I thought about it and seems that I don't need a separate `defaultValue` property because, the `value` property itself can be used to set the initial value.

apps/pie-storybook/stories/pie-textarea.stories.ts
```diff
@@ -42,6 +93,48 @@ const textareaStoryMeta: TextareaStoryMeta = {
...
         },
+        name: {
+            description: 'The name of the textarea (used as a key/value pair with `value`). This is required in order to work properly with forms.',
+            control: 'text',
+            defaultValue: {
+                summary: defaultArgs.name,
+            },
+        },
+        value: {
+            description: 'The value of the textarea (used as a key/value pair in HTML forms with `name`).',
```
jamieomaguire, line 104: just triple checking, is this correct for textarea values in forms - it's still a key/value pair with `name`?

packages/components/pie-textarea/src/index.ts
```diff
@@ -43,21 +67,82 @@ export class PieTextarea extends RtlMixin(LitElement) implements TextareaProps {
...
+    public formDisabledCallback (disabled: boolean): void {
+        this.disabled = disabled;
+    }
+
+    /**
+     * Called when the form that owns this component is reset.
+     * Resets the value to the default value.
+     */
+    public formResetCallback (): void {
+        this.value = defaultProps.value;
```
jamieomaguire, line 93: This deviates from the text-input behaviour, which checks if a `defaultValue` prop exits and uses that instead if it does. Which behaviour do we want here?

packages/components/pie-textarea/src/index.ts
```diff
@@ -67,9 +152,16 @@ export class PieTextarea extends RtlMixin(LitElement) implements TextareaProps {
...
-                    data-test-id="pie-textarea"
-                    @input=${this.handleResize}
+                    name=${ifDefined(name)}
+                    autocomplete=${ifDefined(autocomplete)}
+                    .value=${live(value)}
+                    ?autofocus=${autoFocus}
+                    ?readonly=${readonly}
+                    ?required=${required}
                     ?disabled=${disabled}
+                    @input=${this.handleInput}
```
jamieomaguire, line 162: `handleResize` is no longer called on input. Is this correct?

packages/components/pie-textarea/src/textarea.scss
```diff
@@ -21,6 +21,7 @@
   line-height: 0; // Remove once there is text outside the textarea
 
   textarea {
+    width: 100%;
```
jamieomaguire, line 24: I am curious why this is necessary, and if the element should be `display: block`/`inline-block` instead?

leksaBoiko: Screenshot 2024-07-26 at 16 46 49

display: block/inline-block doesn't work, you can check this in pie-aperture form

packages/components/pie-textarea/src/index.ts
leksaBoiko:
```
export const defaultProps: DefaultProps = {
    value: '',
};
```
by default `value` is an empty string, so when I need to reset form, I equal the current value to an empty string

apps/pie-storybook/stories/pie-textarea.stories.ts
leksaBoiko: When the form is submitted, the content of the <textarea> is sent as a key/value pair where the key is the value of the name attribute, and the value is the content entered by the user.

packages/components/pie-textarea/src/index.ts
jamieomaguire: Here is a scenario:

1. I load up a user info page from the server.

2. The `first name` field has a value of `jamie`

3. I edit the field to say `john`

4. I click `reset` on the form

5. The first name input should have a value of `jamie` as this was the initial value. Not an empty string.


Does that make sense? @xander-marjoram just in case you think differently

xander-marjoram: This is a great catch, I think we definitely need to capture this behaviour but I think it's worth its own ticket; I'll write one this morning!

jamieomaguire: Cheers! Happy to approve this PR and address separately in that ticket 👍

xander-marjoram:
This is looking great @leksaBoiko!

One thing: I've discovered a bug which I think needs to be fixed in a follow-up PR.

Scenario: The textarea is focused inside a form, when you press the enter key
Expected behaviour: A new line is added to the text inside the textarea
Current behaviour: The form is submitted

packages/components/pie-textarea/src/index.ts
```diff
@@ -43,21 +66,92 @@ export class PieTextarea extends RtlMixin(LitElement) implements TextareaProps {
...
+    public formResetCallback (): void {
+        this.value = defaultProps.value;
+
+        this._internals.setFormValue(this.value);
+    }
+
+    protected firstUpdated (): void {
+        this._internals.setFormValue(this.value);
+
+        this._textarea.addEventListener('keydown', this.handleKeyDown);
```
jamieomaguire, line 100: for any event listeners we add, we will want to get rid of them in the disconnected callback

xander-marjoram (approved):

xander-marjoram (approved):

jamieomaguire (approved):

PR #1622, opened 2024-07-23
chore(eslint-config-pie): DSW-000 add rule for consistent type imports

xander-marjoram:
I discovered the eslint rule `@typescript-eslint/consistent-type-imports` and I thought it would be good to add it to our PIE config package.

Adding `type` wherever possible is beneficial because it reduces the amount of code that TypeScript needs to include in the output JS; i.e., if an import is only needed to provide TS typing then JavaScript doesn't need it at all.

[This is the main change](https://github.com/justeattakeaway/pie/pull/1622/files#diff-980653a4f3839eadd474d5b807e3adda20b0658ec0ba89f97f39a8dbd3cf2f06), everything else is just updating to adhere to the new rule.


packages/components/pie-checkbox/src/index.ts
```diff
@@ -1,5 +1,5 @@
 import {
-    LitElement, html, unsafeCSS, PropertyValues, nothing,
```
xander-marjoram, line 2: This was unused

dandel10n (dismissed):

packages/tools/pie-icons-vue/.gitignore
```diff
@@ -1,2 +1,2 @@
-generated/*
-icons/*
+generated
```
xander-marjoram, line 1: This ignores the actual folders as well as their contents

ashleynolan (dismissed):

ashleynolan (approved):

packages/components/pie-assistive-text/src/defs-react.ts
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
+import type React from 'react';
```
jamieomaguire, line 1: gosh I'd be curious to see if this makes a wild different in bundle output

jamieomaguire (approved):

packages/components/pie-assistive-text/src/defs-react.ts
xander-marjoram: I was curious about this too, so I tested it and it turns out it does not reduce the size, although I think that's because we already explicitly exclude `react` packages from the bundle.

jamieomaguire (dismissed):

fernandofranca (approved):

raoufswe (dismissed):
🥇

kevinrodrigues (dismissed):

raoufswe (approved):

kevinrodrigues (approved):

PR #1644, opened 2024-07-30
chore(pie-monorepo): DSW-000 snyk policy file

dandel10n:
Snyk policy file to add ignored vulnerabilities with the reasoning.
Docs: https://docs.snyk.io/manage-risk/policies/the-.snyk-file


packages/.snyk
```diff
@@ -0,0 +1,16 @@
+ # Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities
+version: v1.25.0
+# ignores vulnerabilities until expiry date; change duration by modifying expiry date
+ignore:
+    SNYK-JS-IP-7148531:
```
xander-marjoram, line 5: Do you think we will end up with dozens of these entries for the `0.0.0-use.local` issue?

dandel10n: For all the closed PRs only these two snyk codes were reported that is why I added them to be ignored:
Screenshot 2024-07-31 at 14 29 11

siggerzz (approved):

xander-marjoram (approved):

PR #1676, opened 2024-08-06
refactor(pie-cookie-banner): DSW-2207 refactored experimental package from cookie banner using page object modal

JoshuaNg2332:
- Refactored `@sand4rt/experimental-ct-web` out of `component` and `accessibility` tests
- Updated `base.page.ts` file to include a new load function for component objects to use
- Updated Cookie Banner page object to support new load function

As part of this update, going forward, if you want to test Cookie Banner locally through component or accessibility tests, you will need to run `yarn dev --filter=pie-storybook` in another terminal window before you can run your tests locally.

### Reviewer 1

### Reviewer 2

### DIFF
--- a/packages/components/pie-cookie-banner/test/accessibility/pie-cookie-banner.spec.ts
+++ b/packages/components/pie-cookie-banner/test/accessibility/pie-cookie-banner.spec.ts
@@ -1,15 +1,10 @@
-
-import { test, expect } from '@justeattakeaway/pie-webc-testing/src/playwright/webc-fixtures.ts';
-import { PieCookieBanner, CookieBannerProps } from '../../src/index.ts';
+import { test, expect } from '@justeattakeaway/pie-webc-testing/src/playwright/playwright-fixtures.ts';
+import { CookieBannerComponent } from '../helpers/page-object/pie-cookie-banner.page.ts';
 
 test.describe('PieCookieBanner - Accessibility tests', () => {
-    test('a11y - should test the PieCookieBanner component WCAG compliance', async ({ makeAxeBuilder, mount }) => {
-        await mount(
-            PieCookieBanner,
-            {
-                props: {} as CookieBannerProps,
-            },
-        );
+    test('a11y - should test the PieCookieBanner component WCAG compliance', async ({ makeAxeBuilder, page }) => {
+        const cookieBannerPage = new CookieBannerComponent(page);
+        await cookieBannerPage.load('');
 
         const results = await makeAxeBuilder().analyze();
 
--- a/packages/components/pie-cookie-banner/test/component/pie-cookie-banner.spec.ts
+++ b/packages/components/pie-cookie-banner/test/component/pie-cookie-banner.spec.ts
@@ -1,5 +1,4 @@
-
-import { test, expect } from '@sand4rt/experimental-ct-web';
+import { test, expect } from '@justeattakeaway/pie-webc-testing/src/playwright/playwright-fixtures.ts';
 import { readFile } from 'fs/promises';
 import { CookieBannerComponent, Level } from 'test/helpers/page-object/pie-cookie-banner.page.ts';
 import { ModalComponent } from '@justeattakeaway/pie-modal/test/helpers/page-object/pie-modal.page.ts';
@@ -8,9 +7,6 @@ import {
     ON_COOKIE_BANNER_MANAGE_PREFS, ON_COOKIE_BANNER_PREFS_SAVED,
     preferences,
 } from '../../src/defs.ts';
-import {
-    PieCookieBanner, CookieBannerProps,
-} from '../../src/index.ts';
 
 const englishLocale = JSON.parse(await readFile(new URL('../../locales/en-gb.json', import.meta.url)));
 const spanishLocale = JSON.parse(await readFile(new URL('../../locales/es-es.json', import.meta.url)));
@@ -28,11 +24,9 @@ test.describe('PieCookieBanner - Component tests', () => {
         pieModalComponent = new ModalComponent(page);
     });
 
-    test('should render successfully', async ({ mount }) => {
+    test('should render successfully', async () => {
         // Arrange
-        await mount(PieCookieBanner, {
-            props: {} as CookieBannerProps,
-        });
+        await pieCookieBannerComponent.load({});
 
         // Act
         const isCookieBannerVisible = await pieCookieBannerComponent.isCookieBannerVisible();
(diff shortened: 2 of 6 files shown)


packages/components/pie-cookie-banner/test/accessibility/pie-cookie-banner.spec.ts
```diff
@@ -1,15 +1,10 @@
...
-    test('a11y - should test the PieCookieBanner component WCAG compliance', async ({ makeAxeBuilder, mount }) => {
-        await mount(
-            PieCookieBanner,
-            {
-                props: {} as CookieBannerProps,
-            },
-        );
+    test('a11y - should test the PieCookieBanner component WCAG compliance', async ({ makeAxeBuilder, page }) => {
+        const cookieBannerPage = new CookieBannerComponent(page);
+        await cookieBannerPage.load('');
```
xander-marjoram, line 7: Is there not a type error here?
Also feels like we could just have a default empty value for no query strings

JoshuaNg2332: Oops yeah, clearly missed that one.
Yeah I think we can. I'll make changes now and clean up the rest of the tests.

packages/components/pie-webc-testing/src/helpers/page-object/base-page.ts
```diff
@@ -1,10 +1,77 @@
...
+        const pageUrl = buildUrl(this.componentName, this.composePath(queries), this.args);
+        await this.open(pageUrl);
+    }
+
+    async open (url: string) {
+        await this.page.goto(url, { waitUntil: 'networkidle' });
+        return this;
+    }
+
+    composePath (queries: Record<string, unknown>) {
```
kevinrodrigues, line 33: Might be worth documenting this method.

JoshuaNg2332: Yup. I'll add it in as a follow up PR.

packages/components/pie-webc-testing/src/helpers/page-object/base-page.ts
```diff
@@ -1,10 +1,77 @@
...
+        if (!queries) {
+            return '';
+        }
+
+        const flattenQueries = (obj: Record<string, unknown>, prefix = ''): string[] => Object.keys(obj).flatMap((key) => {
```
kevinrodrigues, line 38: You could use object.entries in the same way instead of object.keys and then in the flatMap pull out the value removing the need for the line below. Although that method coverts it to an array but i think that's ok in your case?

kevinrodrigues (approved):

kevinrodrigues (dismissed):

packages/components/pie-webc-testing/src/helpers/page-object/base-page.ts
```diff
@@ -1,10 +1,86 @@
 import { type Page } from '@playwright/test';
+import { buildUrl } from './storybook-extensions';
+
+declare global {
+    interface Window { __eventsArray: Array<string>; }
```
xander-marjoram, line 5: I think this is great for now but it might be worth thinking about how we want to handle multiple components pushing events at the same time, if we ever want to do that.

packages/components/pie-webc-testing/src/helpers/page-object/base-page.ts
xander-marjoram: I think an example input -> output would be valuable

packages/components/pie-webc-testing/src/helpers/page-object/base-page.ts
```diff
@@ -1,10 +1,86 @@
...
+                return flattenQueries(value as Record<string, unknown>, newKey);
+            }
+            return `${newKey}:${value}`;
+        });
+
+        return `&args=${flattenQueries(queries).join(';')}`;
+    }
+
+    /**
+     * Emits an on push event that has been passed through from the test
```
xander-marjoram, line 61:
```suggestion
     * Listens for the given event. When it is emitted, the name of the event is pushed to an array on the `window` object.
```

packages/components/pie-cookie-banner/test/component/pie-cookie-banner.spec.ts
```diff
@@ -43,44 +37,34 @@ test.describe('PieCookieBanner - Component tests', () => {
...
             // Act
             await pieCookieBannerComponent.clickAcceptAll(level);
+
+            const events = await pieCookieBannerComponent.getCapturedEvents();
             const isCookieBannerVisible = await pieCookieBannerComponent.isCookieBannerVisible();
 
             // Assert
             expect.soft(isCookieBannerVisible).toBe(false);
-            expect(events).toHaveLength(1);
+            expect(events).toContain(ON_COOKIE_BANNER_ACCEPT_ALL);
```
xander-marjoram, line 53:
```suggestion
            expect(events).toHaveLength(1);
            expect(events).toContain(ON_COOKIE_BANNER_ACCEPT_ALL);
```

xander-marjoram: Or actually, as that's the only event we're listening for, we can probably get away with just checking the length as we were before

packages/components/pie-cookie-banner/test/component/pie-cookie-banner.spec.ts
```diff
@@ -134,38 +110,28 @@ test.describe('PieCookieBanner - Component tests', () => {
...
 
+        const events = await pieCookieBannerComponent.getCapturedEvents();
         const isCookieBannerVisible = await pieCookieBannerComponent.isCookieBannerVisible();
         const isModalVisible = await pieModalComponent.isModalVisible();
 
         // Assert
         expect(isModalVisible).toBe(false);
         expect(isCookieBannerVisible).toBe(false);
-        expect(cookieBannerPrefsSavedEvent).toMatchObject(expectedCookieBannerPrefsSavedEvent);
+        expect(events).toContain(ON_COOKIE_BANNER_PREFS_SAVED);
```
xander-marjoram, line 129: Probably another length check here would suffice

packages/components/pie-cookie-banner/test/component/pie-cookie-banner.spec.ts
```diff
@@ -251,9 +215,9 @@ test.describe('PieCookieBanner - Component tests', () => {
     });
 
     test.describe('`locale` prop', () => {
-        test('should render text in the default (English) language when the locale is not set', async ({ mount }) => {
+        test('should render text in the default (English) language when the locale is not set', async () => {
             // Arrange
-            await mount(PieCookieBanner, {});
+            await pieCookieBannerComponent.load({ defaultPreferences: null });
```
xander-marjoram, line 220: Was it necessary to add this?
```suggestion
            await pieCookieBannerComponent.load();
```

packages/components/pie-cookie-banner/test/component/pie-cookie-banner.spec.ts
```diff
@@ -197,9 +161,9 @@ test.describe('PieCookieBanner - Component tests', () => {
         });
     });
 
-    test('should toggle the `all` preference node if the other preferences are set to true manually', async ({ mount }) => {
+    test('should toggle the `all` preference node if the other preferences are set to true manually', async () => {
         // Arrange
-        await mount(PieCookieBanner, { props: {} as CookieBannerProps });
+        await pieCookieBannerComponent.load({ defaultPreferences: null });
```
xander-marjoram, line 166: The preferences should be false by default, right? 
```suggestion
        await pieCookieBannerComponent.load();
```

Also there's the `no-restricted-syntax` eslint comment below here, is that still needed? I think we may have set a rule that means that's no longer required

packages/components/pie-cookie-banner/test/component/pie-cookie-banner.spec.ts
xander-marjoram: Actually maybe just make an object that contains all of the preferences but with them explicitly set to false, then you can use that in all these preference tests

packages/components/pie-cookie-banner/test/component/pie-cookie-banner.spec.ts
JoshuaNg2332: Just looked and yeah, we still need that `no-restricted-syntax` line.

JoshuaNg2332: Also this one does require the `defaultPreferences: null` as it's currently working backwards based on the default values. Meaning that we'll need to alter this test once we make the fixies for preferences defaults.

PR #1696, opened 2024-08-14
refactor(pie-modal): DSW-2280 use classes instead of attributes for styling

xander-marjoram:
- [Changed] - Use classes instead of attributes for styling
- [Changed] - tidy up modal render functions

### Reviewer 1

### Reviewer 2

### DIFF
--- a/apps/pie-storybook/stories/pie-modal.stories.ts
+++ b/apps/pie-storybook/stories/pie-modal.stories.ts
@@ -237,7 +237,6 @@ export const FocusManagement = createStory<ModalProps>(FocusableElementsPageStor
     returnFocusAfterCloseSelector: '#focus-3',
 });
 export const LargeTextContent = createBaseModalStory({
-    isFooterPinned: false,
     slot: `Lorem ipsum dolor sit amet consectetur adipisicing elit. Reprehenderit quas inventore quasi ullam, sed ab odio dicta, tempore, ex adipisci atque asperiores suscipit quisquam alias aliquam minus amet ad a?
     Iure consequuntur nihil officia odio, ut dolores reprehenderit tenetur, repellat eveniet dolore, dignissimos aspernatur quo laboriosam eum repellendus ratione libero. Aspernatur in, inventore ratione molestias exercitationem repudiandae omnis nisi illo?
     Laborum, aspernatur labore! Nulla corporis laudantium, odio iure cum maiores veritatis. Facere ullam sequi voluptate ipsa neque? Atque necessitatibus aspernatur quibusdam sit pariatur quo sunt, voluptatem doloribus dolore consequatur temporibus?
--- a/apps/pie-storybook/stories/pie-modal.stories.ts
+++ b/apps/pie-storybook/stories/pie-modal.stories.ts
@@ -15,8 +15,6 @@ import {
 } from '@justeattakeaway/pie-modal';
 /* eslint-enable import/no-duplicates */
 
-import { variants as buttonVariants } from '@justeattakeaway/pie-button';
-
 import { StoryMeta, SlottedComponentProps } from '../types';
 import { createStory, sanitizeAndRenderHTML } from '../utilities';
 
--- a/packages/components/pie-modal/src/index.ts
+++ b/packages/components/pie-modal/src/index.ts
@@ -3,7 +3,6 @@ import {
 } from 'lit';
 import { html, unsafeStatic } from 'lit/static-html.js';
 import { property, query } from 'lit/decorators.js';
-import { classMap } from 'lit/directives/class-map.js';
 import { ifDefined } from 'lit/directives/if-defined.js';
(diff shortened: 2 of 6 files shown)


apps/pie-storybook/stories/pie-modal.stories.ts
```diff
@@ -239,7 +237,6 @@ export const FocusManagement = createStory<ModalProps>(FocusableElementsPageStor
     returnFocusAfterCloseSelector: '#focus-3',
 });
 export const LargeTextContent = createBaseModalStory({
-    isFooterPinned: false,
```
xander-marjoram, line 242: There was a bit of funkiness here. We were setting it to `false` but it was still `true` in the component, even though the storybook control showed `false`. You could toggle it off and on and it would work as expected but removing this means the control accurately reflects the current state.

packages/components/pie-modal/src/index.ts
```diff
@@ -359,85 +356,93 @@ export class PieModal extends RtlMixin(LitElement) implements ModalProps {
...
+
+        const modalClasses = {
+            'c-modal': true,
+            [`c-modal--${this.size}`]: true,
+            'c-modal--top': this.position === 'top',
+            'c-modal--hasActions': Boolean(this.leadingAction?.text || this.supportingAction?.text),
+            'c-modal--hasBackButton': this.hasBackButton,
+            'c-modal--hasStackedActions': this.hasStackedActions,
+            'c-modal--isDismissible': this.isDismissible,
+            'c-modal--isFooterPinned': this.isFooterPinned,
```
xander-marjoram, line 418: I'm assuming we want the class naming to match the props? Otherwise I probably would have gone with something like `c-modal--pinnedFooter`

packages/components/pie-modal/src/index.ts
```diff
@@ -359,85 +356,93 @@ export class PieModal extends RtlMixin(LitElement) implements ModalProps {
...
-        } = this;
-
-        const headingTag = unsafeStatic(headingLevel);
+        const headingTag = unsafeStatic(this.headingLevel);
+        const ariaLabel = (this.isLoading && this.aria?.loading) || undefined;
+
+        const modalClasses = {
+            'c-modal': true,
+            [`c-modal--${this.size}`]: true,
+            'c-modal--top': this.position === 'top',
```
xander-marjoram, line 413: `position: centre` doesn't really do anything, so I'm only applying the class conditionally. I could probably do the same with the size, only applying it for small or large, thoughts?

packages/components/pie-modal/src/index.ts
ashleynolan: Doesn't have to match – I'd go for what you suggested

packages/components/pie-modal/src/index.ts
```diff
@@ -359,85 +356,93 @@ export class PieModal extends RtlMixin(LitElement) implements ModalProps {
...
-            hasBackButton,
-            hasStackedActions,
-            heading,
-            headingLevel,
-            isDismissible,
-            isFooterPinned,
-            isFullWidthBelowMid,
-            isLoading,
-            leadingAction,
-            position,
```
jamieomaguire, line 401: As discussed offline, I like this change however it does deviate from the way we reference properties in other components.

I think this could be a nice change but I would prefer that we stay consistent. So I would suggest either chat with the team to get some agreement on which pattern to go with, and if that requires changes lets create a ticket and tackle them all. 

Perhaps there is some kind of linting we can add in the future around this to enforce whichever pattern we choose.

packages/components/pie-modal/src/modal.scss
```diff
@@ -136,12 +136,8 @@
         @media (min-width: $breakpoint-wide) {
             --modal-footer-padding: var(--dt-spacing-e);
         }
-    }
 
-    // When hasStackedActions is set
-    // change the direction of the footer flex container so buttons are full width of modal
-    &[hasstackedactions] {
-        & .c-modal-footer {
+        &.c-modal-footer--stackedActions {
```
xander-marjoram, line 140: This stacked actions class is now applied to the footer in `renderModalFooter`. I thought this was a bit nicer than adding the class to the modal and then applying these styles to the footer when it is inside a modal with that class.

packages/components/pie-modal/src/modal.scss
```diff
@@ -181,16 +177,15 @@
     }
 
     // Ensure correct padding when there is a back button in front of the heading
-    &[hasbackbutton] .c-modal-heading {
+    .c-modal-backBtn + .c-modal-heading {
```
xander-marjoram, line 180: The back button already has a class, so we can use that and the native next-sibling selector instead of having a whole other class to achieve the same thing

ashleynolan (approved):
🚀  :shipit:

jamieomaguire (approved):

PR #1697, opened 2024-08-15
feat(pie-textarea): DSW-2301 add defaultValue property

leksaBoiko:

[Added] - `defaultValue` property 

### Reviewer 1 @kevinrodrigues 


packages/components/pie-textarea/test/component/pie-textarea.spec.ts
```diff
@@ -320,6 +320,25 @@ test.describe('PieTextarea - Component tests', () => {
             });
         });
 
+        test.describe('defaultValue', () => {
+            test('should correctly reset the textarea value to the default value if one is provided when the form is reset', async ({ page }) => {
```
kevinrodrigues, line 324: Do you need a test for just the default value?

`should set the "defaultValue" when a default value is used`

leksaBoiko: `defaultValue` is an optional default value to use when the textarea is reset.

In my opinion this test covers this case, because: 
1. Set the `defaultValue`
2. Type a `value`
3. Reset form
4. Get a `defaultValue` as it was expected

leksaBoiko: I moved this test from "Props" test group to "Form integration"

kevinrodrigues (dismissed):

raoufswe:
I'm not sure if I am missing something but when I pass a default value to the component storybook, the component isn't populated with the value 

https://pr1697-storybook.pie.design/?path=/story/textarea--default&args=defaultValue:meh

Screenshot 2024-08-19 at 17 05 50

leksaBoiko:
> I'm not sure if I am missing something but when I pass a default value to the component storybook, the component isn't populated with the value
> 
> https://pr1697-storybook.pie.design/?path=/story/textarea--default&args=defaultValue:meh
> 

@raoufswe only when `pie-textarea `is inside a form and the form is reset, the `value` is updated to match the `defaultValue`, similar to `pie-text-input` component behaviour

xander-marjoram:
@raoufswe @leksaBoiko Perhaps we could disable the control for `defaultValue` on the base story? I did something similar for the divider, where I disabled the variant control because we already had another story demonstrating the different values, and I added a note pointing the user to the other story if they wanted to interact with it.

https://github.com/justeattakeaway/pie/pull/1704/files#diff-de4acbabc3b96a43b592e821c2f2aab56c0997666b5b53684de6024de2a42590

packages/components/pie-textarea/src/index.ts
```diff
@@ -34,6 +34,9 @@ export class PieTextarea extends FormControlMixin(RtlMixin(LitElement)) implemen
     @property({ type: String })
     public value = defaultProps.value;
 
+    @property({ type: String })
+    public defaultValue?: TextareaProps['defaultValue'];
```
xander-marjoram, line 38: Nit: the `?` already comes through from `defs.ts`
```suggestion
    public defaultValue: TextareaProps['defaultValue'];
```

xander-marjoram (approved):

PR #1703, opened 2024-08-15
feat(pie-button): DSW-2305 add tag prop which lets the button render as an anchor

xander-marjoram:
- [Added] - tag prop which lets the button render as an anchor tag
- [Added] - `href`, `rel` and `target` props which can be passed to the anchor element

### Reviewer 2


packages/components/pie-button/src/button.scss
```diff
@@ -72,7 +72,7 @@
     }
 
     position: relative;
-    display: flex;
+    display: inline-flex;
```
xander-marjoram, line 75: Even though the button was correctly sized, the web component host was a block element so you could click next to the button and it would still register the click.

e.g., here: https://aperture-nuxt.pie.design/components/button

packages/components/pie-button/src/index.ts
```diff
@@ -197,8 +213,9 @@ export class PieButton extends FormControlMixin(LitElement) implements ButtonPro
      */
     private renderSpinner (): TemplateResult {
         const { size, variant, disabled } = this;
-        const spinnerSize = size && size.includes('small') ? 'small' : 'medium'; // includes("small") matches for any small size value and xsmall
-        let spinnerVariant;
+
+        const spinnerSize: SpinnerProps['size'] = size && size.includes('small') ? 'small' : 'medium'; // includes("small") matches for any small size value and xsmall
```
xander-marjoram, line 217: Importing the type for child components actually works pretty nicely

packages/components/pie-webc-testing/src/helpers/defs.ts
```diff
@@ -1,6 +1,21 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-export type PropObject = {
-    [key: string]: any;
```
xander-marjoram, line 3: This is now an optionally-generic type you can pass the component prop type into. This makes sure you can only assign valid values when creating variations inside the visual and accessibility tests.

packages/components/pie-button/src/index.ts
jamieomaguire: love it

packages/components/pie-button/src/index.ts
```diff
@@ -157,17 +173,17 @@ export class PieButton extends FormControlMixin(LitElement) implements ButtonPro
...
-        if (!this.isLoading && this.form) {
-            if (this.type === 'submit') {
-                // only submit the form if either formnovalidate is set, or the form passes validation checks (triggers native form validation)
-                if (this.formnovalidate || this.form.reportValidity()) {
-                    this._simulateNativeButtonClick('submit');
-                }
-            }
+        if (!this.form) return;
+        if (this.isLoading) return;
+        if (this.tag !== 'button') return;
```
jamieomaguire, line 178: absolutely just a preference but I feel like this tag check should be first (or loading)

packages/components/pie-button/src/index.ts
xander-marjoram: We already have it in `dependencies` so shouldn't add anything that isn't already there 👍🏻

apps/pie-storybook/stories/pie-button.stories.ts
```diff
@@ -25,6 +26,15 @@ const buttonStoryMeta: ButtonStoryMeta = {
     title: 'Button',
     component: 'pie-button',
     argTypes: {
+        tag: {
+            description: 'Choose the HTML element that will be used to render the button.<br>For this story, the prop has the value of `button`. See the Anchor story to interact with the component when this prop has a value of `a`.',
```
jamieomaguire, line 30: We probably don't want this description on the Anchor story (which is what currently happens)
Screenshot 2024-08-30 at 14 36 24

packages/components/pie-button/src/index.ts
```diff
@@ -207,23 +224,60 @@ export class PieButton extends FormControlMixin(LitElement) implements ButtonPro
...
+
+        return html`
+            <a
+                href="${ifDefined(href)}"
+                rel="${ifDefined(rel)}"
+                target="${ifDefined(target)}"
+                class="${classMap(classes)}">
+                ${iconPlacement === 'leading' ? html`<slot name="icon"></slot>` : nothing}
+                <slot></slot>
+                ${iconPlacement === 'trailing' ? html`<slot name="icon"></slot>` : nothing}
```
jamieomaguire, line 246: I've been considering this for a while, and I'm becoming less convinced that using a slot for button text is the best approach, especially when it comes to link buttons.

While I understand that the convention is to use a slot, I can't think of many situations where we would want anything other than plain text - particularly for the link variant.

What do you think?

xander-marjoram: Yes I think I agree. As you said to me this might be useful for something like `nuxt-link`, etc, but there might be other workaround for that like adding more possible values for "tag".

Maybe we could suggest this as a change when we move to v1?

xander-marjoram: We discussed this on Slack and we're going to raise this as a separate discussion with the team.

dandel10n (approved):

PR #1708, opened 2024-08-16
feat(pie-docs): DSW-2268 refactors navigation to include link to inde…

maledr5:
Refactor the navigation mechanism so that:
- main navigation items can link to index pages
- caret can independently open the accordion without linking

Decided to keep the focus navigation that will select the `summary` element as well as the `link` and `button`. Is one additional tab step but I did not wanted to remove the a11y advantages of using `<details>` and `<summary>` elements.

**Note on percy changes:** 
- The menu was moved 2px down so all screenshots have changed.
- I have left an [empty index page for Components](https://pr1708-docs.pie.design/components/) so to showcase how it will work if one of the main menu items has content. I will be working on the index page content next so this navigation and the content changes can be merged together. But I wanted to get a review on this changes first.

**Note on resizeObserver removal:**
There is no record of the decision on why we implemented the animation only for desktop and not mobile screens. As this solution relies on the initial animation script, I removed the check for smaller screens (>=600). Without this check I could not see a use for the resizeObserver either, I wasn't able to understand why we would need it even before as we set the accordions once and do not change them on resize any more. If anybody has more context let me know. 

- [x] I have reviewed the new navigation interaction

### Reviewer 2
- [x] I have reviewed the new navigation interaction


apps/pie-docs/src/assets/js/accordions.js
```diff
@@ -122,27 +128,11 @@ class Accordion {
...
-
-            // ensure we don't duplicate accordion instances
-            if (width > 600 && !accordions.length) {
-            // TODO: use fozzie JS breakpoint helpers instead of hardcoding 600px
-                document.querySelectorAll('.c-nav details').forEach((el) => {
-                    accordions.push(new Accordion(el));
-                });
-            }
-        });
-    });
```
raoufswe, line 139: we don't need this logic anymore?

apps/pie-docs/src/_includes/nav.njk
```diff
@@ -50,16 +66,17 @@
...
+                            {{ summary(category.title, category.url, category.children[0].url, category.hasIndexPage, category.key) }}
                             {% if category.children.length %}
                             {# build up a list of subCategories for each subpages (used in component pages) #}
                                 <ul class="c-nav-list">
                                     {% for subpage in category.children %}
                                         {% if subpage.url %}
                                             {% if subpage.children.length and subpage.subPageDropdown %}
-                                                <details {% if page.url and page.url.includes(subpage.key) %} open{% endif %}>
-                                                    {{ summary(subpage.title) }}
+                                                <details>
```
raoufswe, line 78: same as above

apps/pie-docs/src/assets/js/accordions.js
maledr5: Thanks @raoufswe I left a comment around this in the description :) , here it is:

There is no record of the decision on why we implemented the animation only for desktop and not mobile screens. As this solution relies on the initial animation script, I removed the check for smaller screens (>=600). Without this check I could not see a use for the resizeObserver either. I wasn't able to understand why we would need it before as we set the accordions once and do not change them on resize any more. If anybody has more context let me know.

Do you know why we need it or remember context around this decision?

apps/pie-docs/src/_includes/nav.njk
```diff
@@ -50,16 +66,17 @@
             {% for category in section %}
                 {% if category.url %}
                     <li>
-                        <details {% if page.url and page.url.includes(category.url) %} open{% endif %}>
-                            {{ summary(category.title, category.key) }}
+                        <details>
```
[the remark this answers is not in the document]
maledr5: The issue here is that now we navigate. The moment we navigate we lose the animation. So what I do now is trigger the animation after the navigation happens (after the re-render). if I add this line again, as it is a default html function that will make the list visible, with the current solution creates this weird jump and of course removes the animation.

The only solution I can think about is remembering the previous state using browser local storage: storing the previous url and only triggering the animation when it changes. Any thoughts?

https://github.com/user-attachments/assets/810088f7-1608-444c-a30e-0d4776516d14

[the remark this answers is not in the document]
maledr5: Added this line back and removed animation when clicking on links as agreed with the team :)

apps/pie-docs/src/_includes/nav.njk
maledr5: Added this line back and removed animation when clicking on links as agreed with the team :)

apps/pie-docs/src/assets/js/accordions.js
raoufswe: maybe @jamieomaguire has any idea? 👀

raoufswe (dismissed):

apps/pie-docs/src/assets/js/accordions.js
jamieomaguire: Ash asked for it, I delivered :P 

In all seriousness, it was something we felt was a nice UX improvement. Why we left out mobile, I cannot remember unfortunately. 

I think it is perfectly fine for us to remove this for now and revisit after chatting with design :)

jamieomaguire (approved):

jamieomaguire (approved):

raoufswe (approved):

PR #1719, opened 2024-08-21
cosmetic(pie-textarea): DSW-2302 fix positioning of resize handle

xander-marjoram:
- Redistribute the padding between the textarea and the wrapping element to reposition the resize handle

### Reviewer 1 @leksaBoiko 

### Reviewer 2 @maledr5 


configs/pie-components-config/playwright-lit-visual-config.js
```diff
@@ -51,7 +51,7 @@ export function getPlaywrightVisualConfig () {
                 name: 'visual:mobile',
                 grep: /@mobile/,
                 use: {
-                    ...devices['Pixel 5'],
+                    ...devices['iPhone 11 Pro'],
```
xander-marjoram, line 54: Pixel 5 is a 393px width. iPhone 11 Pro is 375px, which matches our `percyWidths` breakpoint.

There was a bit of a bug where the element wasn't properly resizing when Percy set its page to 375px wide, so it's best to align these as close as possible so that we get accurate visual tests.

There are one or two other tests in the docs site that use the `@mobile` flag which this in theory could affect, but I'm not expecting to see a difference.

siggerzz: Docs uses the config defined at the root of the repo. The config you've changed is specific to Lit components, so shouldn't have an impact 👍🏼 - https://github.com/justeattakeaway/pie/blob/main/playwright-visual.config.ts#L65

siggerzz: On a separate note, I've noticed that `@justeattakeaway/pie-components-config` isn't listed as a devDep in all components, meaning visual tests probably won't trigger for all components for this change. This PR is pretty low risk, but it's something we definitely need to sort.

packages/components/pie-textarea/test/component/pie-textarea.spec.ts
```diff
@@ -93,92 +87,6 @@ test.describe('PieTextarea - Component tests', () => {
             });
         });
 
-        test.describe('size', () => {
```
xander-marjoram, line 96: These tests were checking that the attributes were added to the element. As we're not styling with attributes anymore these aren't needed, and I didn't think the tests were useful enough to rewrite them to check for the class.

configs/pie-components-config/playwright-lit-visual-config.js
```diff
@@ -51,7 +51,7 @@ export function getPlaywrightVisualConfig () {
                 name: 'visual:mobile',
                 grep: /@mobile/,
                 use: {
-                    ...devices['Pixel 5'],
+                    ...devices['Galaxy S8'],
```
xander-marjoram, line 54: This only applies to tests with `@mobile` in their name. Currently we don't have any in the components folder so this won't affect any other tests.

The reason I'm adding it is that there is a bit of a problem in how Percy works, where it the HTML snapshot is rendered at a desktop width and sent to the server and re-rendered at a mobile width, where the textarea should be taller. By telling the test to use a narrower viewport it means the initial snapshot more closely matches the final screenshot.

leksaBoiko (approved):

PR #1730, opened 2024-08-27
feat(pie-docs): DSW-2268 adds dynamic display of index page cards

maledr5:
Created a shortcode that will display a card for each element in the navigation under a category.
A category is a main navigation item like `Components` or `Foundations`.
This shortcode can be used as content for the Category page when we support one. Currently this PR adds support for components, and a new will be implementing Foundations.

Future updates will only need to add the image to assets, or ignore any items that are not supposed to be displayed.

- [x] If new links have been added, I have tested them in the preview

### Reviewer 2
- [ ] I have reviewed the Docs site preview
- [ ] If new links have been added, I have tested them in the preview


apps/pie-docs/src/assets/styles/components/_indexPage.scss
```diff
@@ -0,0 +1,50 @@
+@use '../fozzie-config' as f;
+@use '../variables' as v;
+@use '@justeattakeaway/pie-css/scss' as p;
+
+.c-indexPage {
```
jamieomaguire, line 5: Just noticed at 320px width (minimum we support) it feels a little squashed where there is no inline whitespace

Screenshot 2024-08-30 at 15 51 15

apps/pie-docs/src/_11ty/shortcodes/indexPageDisplay.js
```diff
@@ -0,0 +1,73 @@
...
+            const imageDirectory = imageSrcDirectory || defaultImageDirectory;
+            const imgSrc = `${imageDirectory}/${itemKeySlug}/${menuItemSlug}.svg`;
+            const imgMobileSrc = `${imageDirectory}/${itemKeySlug}/${menuItemSlug}-mobile.svg`;
+            const fallbackImage = `${imageDirectory}/${itemKeySlug}/default.svg`;
+
+            const hasSource = fs.existsSync(path.join(__dirname, imgSrc));
+            const hasMobileSource = fs.existsSync(path.join(__dirname, imgMobileSrc));
+
+            const renderFallbackImage = () => {
+                console.warn(`Fallback image used for ${menuItemSlug}. Please ensure image is provided.`);
```
jamieomaguire, line 54: I would prefer we fail the build here. Unless we are happy with a fallback image in prod?

apps/pie-docs/src/_11ty/shortcodes/indexPageDisplay.js
```diff
@@ -0,0 +1,73 @@
...
+
+/** IndexPageDisplay Shortcode - automatic page rendering:
+ * Render index items in a flexible grid.
+ * Every main item in our menu can now render an index page. This shortcode will loop on every child menu item and create an
+ * interactive card with a link to that menu item. Items can be excluded if specified. Images must exist for items not excluded.
+ * Mobile images are optional, desktop image will be rendered if mobile image is not provided.
+ * A default image will be rendered if desktop image is not provided and a warning is printed to the server console.
+ * Images must be added:
+ *  - within a folder with the main menu item name, for example 'components/' 'foundations/'
+ *  - image should have the name of the sub-menu in snakecase, for example Checkbox Group is called 'checkbox-group'
```
jamieomaguire, line 13: These are good instructions, is it worth adding to a readme somewhere?

apps/pie-docs/src/_11ty/shortcodes/indexPageDisplay.js
maledr5: I added because it improved the experience while building but I think it should not be too difficult to build new pages and I rather throw an error indeed. I'll change this to throw.

apps/pie-docs/src/_11ty/shortcodes/indexPageDisplay.js
maledr5: There is some instructions in readme. I will add a note to look into this component for more information.

raoufswe (approved):

apps/pie-docs/src/_11ty/shortcodes/indexPageDisplay.js
```diff
@@ -0,0 +1,78 @@
...
```
fernandofranca, line 13: Looks like kebab-case, I think.

maledr5: 100% 😄 good catch!

PR #1763, opened 2024-09-02
feat(pie-lottie-player): DSW-2365 implement lottie player

fernandofranca:
- [ ] If visual tests are required for the component, I have created a project in Percy
- [ ] I have [created a new label](https://github.com/justeattakeaway/pie/issues/labels) and added it to the [PR Labeler workflow](https://github.com/justeattakeaway/pie/blob/main/.github/project-labeler.yml)

### Reviewer 1

### Reviewer 2


packages/components/pie-lottie-player/src/index.ts
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    disconnectedCallback () {
+        super.disconnectedCallback();
+        this._destroyAnimation();
+    }
+
+    private _canUseDOM () {
+        return typeof window !== 'undefined' && 'document' in window && 'createElement' in window.document;
+    }
```
jamieomaguire, line 46: would the Lit `isServer` directive be useful here instead? [Reference](https://lit.dev/docs/api/misc/#isServer)

packages/components/pie-lottie-player/src/defs.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * By the default animations loop, setting this prop as true will prevent such behaviour
+     */
+    loopDisabled?: boolean;
+
+    /**
+     * By default animations start playing as soons as its data is available, setting this prop as true will prevent such behaviour
+     */
+    autoPlayDisabled?: boolean;
```
jamieomaguire, line 27: Personal preference but I wonder if `loop` and `autoPlay` might be easier to read here (and then set them to false instead of true). Not a big deal though!

packages/components/pie-lottie-player/README.md
```diff
@@ -0,0 +1,91 @@
+<p align="center">
```
jamieomaguire, line 1: Please can we add some basic usage and props table info here? Basic usage in my opinion involves passing some animations to the component

jamieomaguire:
Amazing work @fernandofranca! Just left some comments/thoughts 👍

packages/components/pie-lottie-player/src/defs.ts
fernandofranca: That was exactly the first approach I used. 

Since having both as true by default makes sense for most cases, it didn't seemed necessary to have them reflected whenever the component is initialized. Besides that would make both props required since they are booleans, and that I consider an unnecessary inconvenience for the consumer.
E.g.: 
```HTML
<pie-lottie-player animationSrc="..." loop autoplay><pie-lottie-player>
<!-- If consumer forgets to add autoplay and the animation will likely not behave as expected -->
```
Let me know if I missed something 👍

packages/components/pie-lottie-player/README.md
fernandofranca: My bad, I was pretty sure to have added that.

apps/pie-storybook/package.json
```diff
@@ -5,7 +5,7 @@
   "description": "Storybook for PIE Design System components",
   "scripts": {
     "dev": "storybook dev -p 6006",
-    "build": "storybook build --output-dir dist",
+    "build": "storybook build --output-dir dist && cp -R ./static ./dist",
```
fernandofranca, line 8: This ensures to copy the animation assets stored in the `apps/pie-storybook/static` folder, so they can be deployed as well.

xander-marjoram: I wonder if this can be done through turbo, adding a new script called something like `copy:static-assets` that depends on `build`, and then `dev` depends on the copy step

packages/components/pie-lottie-player/test/accessibility/pie-lottie-player.spec.ts
```diff
@@ -0,0 +1,18 @@
...
+import { PieLottiePlayer, LottiePlayerProps } from '../../src/index.ts';
+
+test.describe('PieLottiePlayer - Accessibility tests', () => {
+    test('a11y - should test the PieLottiePlayer component WCAG compliance', async ({ makeAxeBuilder, mount }) => {
+        await mount(
+            PieLottiePlayer,
+            {
+                props: {} as LottiePlayerProps,
+            },
+        );
```
xander-marjoram, line 12: Can be simplified as we're not passing any props in
```suggestion
import { PieLottiePlayer } from '../../src/index.ts';

test.describe('PieLottiePlayer - Accessibility tests', () => {
    test('a11y - should test the PieLottiePlayer component WCAG compliance', async ({ makeAxeBuilder, mount }) => {
        await mount(PieLottiePlayer);
```

packages/components/pie-webc/package.json
```diff
@@ -197,6 +197,16 @@
       "import": "./react/checkbox-group.js",
       "require": "./react/checkbox-group.js",
       "types": "./react/checkbox-group.d.ts"
+    },
+    "./components/lottie-player.js": {
```
xander-marjoram, line 201: Nit: can these be alphabetised please

packages/components/pie-lottie-player/src/defs.ts
```diff
@@ -0,0 +1,58 @@
...
+     */
+    speed?:number;
+
+    /**
+     * Sets the animation reproduction direction
+     */
+    direction?:AnimationDirection;
+
+    /**
+     * Plays ther animation
```
xander-marjoram, line 41: Would this be more accurate?
```suggestion
     * Plays the animation from the beginning
```

packages/components/pie-lottie-player/src/defs.ts
```diff
@@ -0,0 +1,58 @@
...
+    stop (): void;
+}
+
+export type DefaultProps = ComponentDefaultProps<LottiePlayerProps, keyof Omit<LottiePlayerProps, 'animationSrc' | 'animationData' | 'play' | 'stop'>>;
+
+export const defaultProps: DefaultProps = {
+    loopDisabled: false,
+    autoPlayDisabled: false,
+    speed: 1,
+    direction: directions[0],
```
xander-marjoram, line 57: I think this is easier to read
```suggestion
    direction: 'forward',
```

packages/components/pie-lottie-player/src/index.ts
```diff
@@ -0,0 +1,206 @@
...
+
+    set speed (value) {
+        this._speed = value;
+        if (!this._animationInstance) return;
+
+        this._animationInstance.setSpeed(value);
+    }
+
+    @property({ type: String, reflect: true })
+    get direction ():AnimationDirection {
```
xander-marjoram, line 147: This means we don't need the other import, and the direction type doesn't need to be exported from the defs file
```suggestion
    get direction (): LottiePlayerProps['direction'] {
```

apps/pie-storybook/package.json
fernandofranca: I don't see the point, since the it's only necessary for the build step.

packages/components/pie-lottie-player/src/defs.ts
jamieomaguire: All makes sense, thanks for the summary! 👍

apps/pie-storybook/package.json
jamieomaguire: That might also require adding turbo steps to the root `turbo.json` which I feel is a bit messy already. Especially if it's only relevant for this component.

jamieomaguire (dismissed):

packages/components/pie-lottie-player/src/index.ts
```diff
@@ -0,0 +1,215 @@
...
+        const shouldResumePlaying = this._loopDisabled === false && this._animationInstance.isPaused && this._animationInstance.isLoaded;
+        if (shouldResumePlaying) this.play();
+    }
+
+    @property({ type: Number, reflect: true })
+    get speed () {
+        return this._speed;
+    }
+
+    set speed (value) {
```
xander-marjoram, line 146: I'm curious if there was a specific reason you decided to go primarily with getters and setters rather than doing something like checking the changed properties in the `updated` hook?

packages/components/pie-lottie-player/src/index.ts
```diff
@@ -0,0 +1,215 @@
...
+// Valid values available to consumers
+export * from './defs';
+
+const componentSelector = 'pie-lottie-player';
+
+/**
+ * @tagname pie-lottie-player
+ */
+export class PieLottiePlayer extends LitElement implements LottiePlayerProps {
+    @query('div')
```
xander-marjoram, line 20: This evidently works well, but it feels perhaps a little more solid and reliable to hook onto an id or an attribute rather than `div`. Not the most important thing but thought I'd mention it.

packages/components/pie-lottie-player/test/component/pie-lottie-player.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+                animationData,
+            } as LottiePlayerProps;
+            await mount(PieLottiePlayer, {
+                props,
+            });
+
+            // Act
+            const component = page.locator(componentSelector);
+            const autoPlayDisabled = await component.evaluate((el:PieLottiePlayer) => el.autoPlayDisabled);
+            const loopDisabled = await component.evaluate((el:PieLottiePlayer) => el.autoPlayDisabled);
```
xander-marjoram, line 56:
```suggestion
            const loopDisabled = await component.evaluate((el:PieLottiePlayer) => el.loopDisabled);
```

xander-marjoram: I think this needs updating in a couple of other places too

apps/pie-storybook/stories/pie-lottie-player.stories.ts
```diff
@@ -0,0 +1,130 @@
...
+        <pie-button size="xsmall" @click=${handleStopAnimationClick}>Stop animation</pie-button>
+        <pie-button size="xsmall" @click=${handleStartAnimationClick}>Start animation</pie-button>
+    </div>
+    <pie-lottie-player
+        .animationSrc="${animationSrc === undefined ? animationSrcOptions[0] : animationSrc}"
+        .aria="${aria}"
+        ?loopDisabled="${loopDisabled}"
+        ?autoPlayDisabled="${autoPlayDisabled}"
+        .speed="${speed || 1}"
+        .direction="${direction || directions[0]}"
```
xander-marjoram, line 126: Can we just rely on (and also test) the default prop behaviour? `1` and `"forward"` are the default values
```suggestion
        .speed="${speed}"
        .direction="${direction}"
```

packages/components/pie-lottie-player/test/visual/pie-lottie-player.spec.ts
xander-marjoram, line 1: Something for a later PR but I wonder if we can emulate `prefers-reduced-motion` for a visual test 🤔

packages/components/pie-lottie-player/src/index.ts
```diff
@@ -0,0 +1,215 @@
...
+    private _animationSrc = '';
+    private _animationData?:object;
+    private _autoPlayDisabled = defaultProps.autoPlayDisabled;
+    private _loopDisabled = defaultProps.loopDisabled;
+    private _speed = defaultProps.speed;
+    private _direction = defaultProps.direction;
+
+    protected async firstUpdated (): Promise<void> {
+        // Check if the code is running in a browser environment
+        if (isServer) return;
```
xander-marjoram, line 35: In theory this should be unnecessary as `firstUpdated` is only called on the client: https://lit.dev/docs/components/lifecycle/#firstupdated

packages/components/pie-lottie-player/src/index.ts
fernandofranca: If not mistaken, the main thing I remember that made me go away from it was that it was rather inconvenient when it was necessary to override values.

Right now, if you look at each setter you'll recognize each one does a particular job. Shoving everything in the `updated` would make things harder to read.

apps/pie-storybook/stories/pie-lottie-player.stories.ts
fernandofranca: It bothered me a bit that setting 0 stops the animation :)

xander-marjoram: That sounds expected to me 😄

packages/components/pie-lottie-player/src/index.ts
xander-marjoram: Sure, I think other components are a little more specific like `input[type="checkbox"]` or `textarea` rather than just `div`. We can update later if it starts causing problems.

packages/components/pie-lottie-player/src/index.ts
```diff
@@ -0,0 +1,210 @@
...
+    @property({ type: String, reflect: true })
+    get direction ():LottiePlayerProps['direction'] {
+        return this._direction;
+    }
+
+    set direction (value) {
+        if (!value) return;
+        this._direction = value;
+
+        if (!this._animationInstance) return;
```
xander-marjoram, line 159: Nit: This is checked in this method and also in `_updateDirection` immediately after.

Happy for this to be updated in the next PR if it will exist 😄

packages/components/pie-webc/package.json
```diff
@@ -228,6 +238,8 @@
     "@justeattakeaway/pie-form-label": "0.14.1",
     "@justeattakeaway/pie-icon-button": "0.28.12",
     "@justeattakeaway/pie-link": "0.18.0",
+    "@justeattakeaway/pie-lottie-player": "0.0.0",
+    "@justeattakeaway/pie-link": "0.18.0",
```
xander-marjoram, line 242: pie-link seems to be duplicated

fernandofranca: 😄 Every time I try to solve conflicts in GH UI...

xander-marjoram (approved):

dandel10n (approved):

PR #1767, opened 2024-09-04
feat(pie-docs): DSW-2345 enhance contentLayout

leksaBoiko:

Feature url: https://pr1767-docs.pie.design/components/textarea/#auto-resize

[Updated] - Textarea overview page layout for "Behaviours" section
[Updated] - the `contentLayout` container to display the content of the specified number of columns

- [x] I have reviewed the `PIE Docs` PR preview

- [x] I have reviewed the `PIE Docs` PR preview

- [x] I have reviewed the `PIE Docs` PR preview


xander-marjoram:
Visually this looks great, and it's nicely done.

In terms of the code, I think it could be improved to be more flexible. A three-column layout should be the default for the larger viewports and I don't think we need a seperate class or a `columns` property to control this.

The card list on the docs homepage (For designers, For engineers, Our libraries) I think is quite a good example of how I'd prefer something like this to behave.

leksaBoiko:
> Visually this looks great, nicely done.
> 
> In terms of the code, I think it could be improved to be more flexible. A three-column layout should be the default for the larger viewports and I don't think we need a seperate class or a `columns` property to control this.
> 
> The card list on the docs homepage (For designers, For engineers, Our libraries) I think is quite a good example of how I'd prefer something like this to behave.

Based on feedback from the designers, it seems we need to include a separate 3-column layout, as there might be instances where both 2-column and 3-column layouts are displayed on the same screen resolution. However, it would be a good idea to confirm this with the designers for clarity.

tingting-lau:
Hey leksa is right, we use both 2-column and 3-column layout in the documentation

xander-marjoram (approved):

apps/pie-docs/src/assets/styles/components/_contentLayout.scss
```diff
@@ -13,6 +13,18 @@
...
+    @include f.media('>=huge') {
+      flex-flow: row wrap;
+
+      > * {
+        // stylelint-disable-next-line declaration-block-no-redundant-longhand-properties
+        flex: 0 1 calc(33.333% - var(--dt-spacing-e)); // Three equal-width columns
+      }
+    }
+  }
 }
```
jamieomaguire, line 28: I don't think this is ideal. We should be writing the CSS more fluidly so it renders 3 items on a single row. For example in a grid: `grid-template-rows: repeat(1fr, 3);`.

Why do we specifically need to write separate code for 3 items? 

If I have 2 or 3, but tell the container to display then on a row, why is it different for 3?

Furthermore, what if we want to display 4 or 5 in future?

jamieomaguire (approved):

PR #1859, opened 2024-09-13
feat(pie-radio): DSW-2307 add basic component functionality

xander-marjoram:
- [Added] - `checked`, `defaultChecked`, `disabled`, `name`, `required` and `value` props

Added tests covering `checked`, `disabled` and `required` props.

To prevent this PR from becoming too large, I will add the form reset behaviour and tests for the remaining props in a second PR.

In my opinion, visual tests should be added as part of the styling ticket.

---

### Reviewer 1 - Maryia

### Reviewer 2


packages/components/pie-radio/src/defs.ts
```diff
@@ -1,3 +1,45 @@
...
+}
+
+export type DefaultProps = ComponentDefaultProps<RadioProps, keyof Omit<RadioProps, 'name'>>;
+
+export const defaultProps: DefaultProps = {
+    checked: false,
+    defaultChecked: false,
+    disabled: false,
+    required: false,
+    value: 'on',
```
xander-marjoram, line 44: I'm a bit unsure about this one. This is technically the default value which will be submitted if one isn't provided. ([See MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio#data_representation_of_a_radio_group)).

I wonder if the better thing to do would be to make `value` a required prop to make sure that consumers specify this.

dandel10n: For radio MDN docs suggest not to forget to pass value to the radio even with the "on" being an html default, so lets make it a required prop instead 👍

packages/components/pie-radio/src/index.ts
```diff
@@ -12,9 +15,50 @@ const componentSelector = 'pie-radio';
...
+    public checked = defaultProps.checked;
+
+    @property({ type: Boolean })
+    public defaultChecked = defaultProps.defaultChecked;
+
+    @property({ type: Boolean })
+    public disabled = defaultProps.disabled;
+
+    @property({ type: String })
+    public name: RadioProps['name'];
```
dandel10n, line 29:
```suggestion
    public name?: RadioProps['name'];
```

packages/components/pie-radio/src/index.ts
```diff
@@ -12,9 +15,50 @@ const componentSelector = 'pie-radio';
...
+            ?disabled="${disabled}"
+            ?required="${required}">
+        <label for="radioId">
+            <slot></slot>
+        </label>`;
+    }
+
+    public get validity () : ValidityState {
+        return this.radio.validity;
     }
```
dandel10n, line 62: I am not sure how strictly we follow the files structure but for other components `render()`  part is the last one, so can we move this getter higher up?

packages/components/pie-radio/src/defs-react.ts
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
+import type React from 'react';
 /**
  * TODO: Verify if ReactBaseType can be set as a more specific React interface
  * Use the React IntrinsicElements interface to find how to map standard HTML elements to existing React Interfaces
```
dandel10n, line 4: Not sure if you plan to do it later but this comment can be removed and ReactBaseType can be set to `React.InputHTMLAttributes<HTMLInputElement>`

dandel10n (commented):
Looks good, I left only a few minor comments

packages/components/pie-radio/src/index.ts
xander-marjoram: This is not required in the index file because the `RadioProps` type already specifies the type as optional.

packages/components/pie-radio/src/index.ts
xander-marjoram: Interesting point. I would like it if we could have some linting rules about ordering of methods and properties. I remember Vue being quite opinionated with this sort of thing.

dandel10n (dismissed):

dandel10n (dismissed):

apps/pie-storybook/stories/pie-radio.stories.ts
```diff
@@ -1,19 +1,80 @@
...
+            description: 'The name of the radio (used as a key/value pair with `value`). This is required in order to work properly with forms.',
+            control: 'text',
+            defaultValue: {
+                summary: defaultArgs.name,
+            },
+        },
+
+        required: {
+            description: 'Same as native required attribute. If any radio button in a same-named group of radio buttons has the required attribute, a radio button in that group must be checked, although it doesn\'t have to be the one with the attribute applied.',
+            control: 'boolean',
```
jamieomaguire, line 59: Not a blocker to this PR, however this may benefit from a nice code example in the docs

apps/pie-storybook/stories/pie-radio.stories.ts
```diff
@@ -1,19 +1,80 @@
...
+            defaultValue: {
+                summary: defaultArgs.required,
+            },
+        },
+
+        slot: {
```
jamieomaguire, line 65: Out of interest why are we opting for a slot here?

packages/components/pie-radio/test/component/pie-radio.spec.ts
```diff
@@ -17,4 +31,197 @@ test.describe('PieRadio - Component tests', () => {
...
+                });
+
+                // Act
+                const radio = component.locator(inputSelector);
+
+                // Assert
+                expect((await radio.getAttribute('required'))).toBe(null);
+            });
+
+            test('should apply the required attribute to the input element', async ({ mount }) => {
```
jamieomaguire, line 124: Would it be possible to test that `valueMissing` is `true` instead of checking the attribute has been rendered? (assuming this is the expected ValidityState)

apps/pie-storybook/stories/pie-radio.stories.ts
xander-marjoram: This is a good idea. I'll see what I can do with the next PR on this ticket (I'll be adding more form functionality)

apps/pie-storybook/stories/pie-radio.stories.ts
xander-marjoram: This was mainly to match the checkbox component but maybe a prop would be better here.

Any strong feelings @dandel10n?

packages/components/pie-radio/test/component/pie-radio.spec.ts
xander-marjoram: Another good idea. That validity state only applies when there are multiple radios in the named group, and I'll be adding more tests like those in the next PR.

apps/pie-storybook/stories/pie-radio.stories.ts
dandel10n: For checkbox we went with a slot because in Figma component uses a Label child component that has its own options like trailing content etc. So to give consumers more freedom to customize the label we made it a slot.

For radio I don't see Label child component being used but in theory label can still have a link as part of it for example so slot makes it easier for consumers to implement.

dandel10n (approved):

jamieomaguire (approved):

PR #1862, opened 2024-09-13
feat(pie-toast): DSW-2204 implement auto-dismiss / persistent feature

thejfreitas:
In this Pull Request we are implementing the auto-dismiss feature after 5 seconds or when configurable via `duration` property and deactivating it if the `duration` property is null as well persistence when the user hover the mouse onto the component and when the user focus on the action button and close button using the keyboard avoiding closing the toast and reseting the auto-dismiss functionality.

### Extra
We've created additional documentation pages for each variant as well its combination with `isStrong` property.

### Screenshots and Videos
https://github.com/user-attachments/assets/d37dc0fb-5885-4005-9300-045104a25356

Screenshot 2024-09-13 at 2 39 29 PM

═══════════════════════════════════════════════════════════


apps/pie-storybook/stories/pie-toast.stories.ts
```diff
@@ -63,6 +64,10 @@ const toastStoryMeta: ToastStoryMeta = {
             description: 'The leading action configuration for the toast.',
             control: 'object',
         },
+        duration: {
+            description: 'It set the duration of the toast in milliseconds before it auto-dismiss.',
```
kevinrodrigues, line 68: `It sets the duration of the toast in milliseconds before it auto-dismisses.`?

packages/components/pie-toast/src/defs.ts
```diff
@@ -44,6 +46,13 @@ export interface ToastProps {
    * The leading action for the toast.
    */
   leadingAction?: ActionProps;
+
+  /**
+   * It set the duration of the toast in milliseconds before it auto-dismiss
```
kevinrodrigues, line 51: `It sets the duration of the toast in milliseconds before it auto-dismisses.`

packages/components/pie-toast/src/defs.ts
```diff
@@ -44,6 +46,13 @@ export interface ToastProps {
...
+   * If the value is null auto-dismiss is disabled
+   * If the value is not provided it auto-dismiss after 5 seconds (5000 milliseconds)
```
kevinrodrigues, line 53: `If the value is not provided it auto-dismisses after 5 seconds (5000 milliseconds)`

packages/components/pie-toast/src/index.ts
```diff
@@ -111,6 +164,52 @@ export class PieToast extends RtlMixin(LitElement) implements ToastProps {
...
+                }
+            }, { signal: abortSignal });
+            element.addEventListener(outEvent, () => {
+                this.setAutoDismiss();
+            }, { signal: abortSignal });
+        }
+    }
+
+    /**
+     * It created all event listeners to handle the auto-dismiss capability
```
kevinrodrigues, line 196: `Creates...`

kevinrodrigues (commented):
LGTM 🎉 We may need some component tests for some of the behaviour?

apps/pie-storybook/stories/pie-toast.stories.ts
```diff
@@ -2,7 +2,7 @@ import { html } from 'lit';
 import { ToastProps, defaultProps, variants } from '@justeattakeaway/pie-toast';
 import { action } from '@storybook/addon-actions';
 import { type StoryMeta } from '../types';
-import { createStory } from '../utilities';
+import { createStory, TemplateFunction } from '../utilities';
```
xander-marjoram, line 5:
```suggestion
import { createStory, type TemplateFunction } from '../utilities';
```
And also for `ToastProps` above, please

apps/pie-storybook/stories/pie-toast.stories.ts
```diff
@@ -76,34 +81,43 @@ const toastStoryMeta: ToastStoryMeta = {
...
     variant,
+    duration,
 }: ToastProps) => html`
     <pie-toast 
         ?isOpen="${isOpen}" 
         ?isDismissible="${isDismissible}" 
         ?isStrong="${isStrong}"
         variant="${variant}"
         message="${message}" 
+        .duration="${duration}"
```
xander-marjoram, line 103: I _think_ it should be this instead. You may need to wrap it in `ifDefined` to protect against `null`.
```suggestion
        duration="${duration}"
```
```suggestion
        .duration="${duration}"
```

packages/components/pie-toast/src/defs.ts
```diff
@@ -2,6 +2,8 @@ import { type ComponentDefaultProps } from '@justeattakeaway/pie-webc-core';
 
 export const variants = ['neutral', 'info', 'warning', 'success', 'error'] as const;
 
+export const defaultDuration = 5000;
```
xander-marjoram, line 5: I think this should be moved to `defaultProps` instead

packages/components/pie-toast/src/defs.ts
```diff
@@ -44,6 +46,13 @@ export interface ToastProps {
    * The leading action for the toast.
    */
   leadingAction?: ActionProps;
+
+  /**
+   * It set the duration of the toast in milliseconds before it auto-dismiss
+   * If the value is null auto-dismiss is disabled
+   * If the value is not provided it auto-dismiss after 5 seconds (5000 milliseconds)
+   */
+  duration?: number | null;
```
xander-marjoram, line 55: We don't tend to use `null` very often. What do you think about having `0` as the value for persistent toasts?

packages/components/pie-toast/src/index.ts
```diff
@@ -119,6 +218,14 @@ export class PieToast extends RtlMixin(LitElement) implements ToastProps {
     protected async updated (_changedProperties: PropertyValues<this>) {
         if (_changedProperties.has('isOpen') && this.isOpen) {
             dispatchCustomEvent(this, ON_TOAST_OPEN_EVENT, { targetNotification: this });
+
+            if (this.duration !== null) {
+                this.createAutoDismissEventListeners();
+            }
+        }
+
+        if (_changedProperties.has('isOpen') && !this.isOpen && this._abortController) {
```
xander-marjoram, line 227: The method already checks for the abort controller so we don't need to check it here too

packages/components/pie-toast/src/index.ts
```diff
@@ -137,7 +244,8 @@ export class PieToast extends RtlMixin(LitElement) implements ToastProps {
             _changedProperties.has('message') ||
             _changedProperties.has('isDismissible') ||
             _changedProperties.has('isMultiline') ||
-            _changedProperties.has('leadingAction')) {
+            _changedProperties.has('leadingAction') ||
+            _changedProperties.has('duration')) {
```
xander-marjoram, line 248: Can you remind me why this is necessary? If we still need to do it, I think it's worth adding a comment as a reminder

packages/components/pie-toast/src/index.ts
```diff
@@ -119,6 +218,14 @@ export class PieToast extends RtlMixin(LitElement) implements ToastProps {
     protected async updated (_changedProperties: PropertyValues<this>) {
         if (_changedProperties.has('isOpen') && this.isOpen) {
             dispatchCustomEvent(this, ON_TOAST_OPEN_EVENT, { targetNotification: this });
+
+            if (this.duration !== null) {
+                this.createAutoDismissEventListeners();
+            }
```
xander-marjoram, line 224: The duration check could probably be moved inside the method?

packages/components/pie-toast/src/defs.ts
thejfreitas: @xander-marjoram Interesting point. I will think about it with @kevinrodrigues

kevinrodrigues: Discussed with @thejfreitas.

@xander-marjoram when refining this ticket there were some comments around using `null` as opposed to `0` which I think i used in the first place but cant remember. @raoufswe also made a suggested to use `null` too we looked at this UI library as part of that. https://v2.chakra-ui.com/docs/components/toast (`null` representing `infinite` as opposed to using `0`).

apps/pie-storybook/stories/pie-toast.stories.ts
thejfreitas: Nope! Using the approach suggested and adding `ifDefined` to make it work properly, even all components has been set the `duration` as `null` in the storybook file the component will close after 5 seconds. So its better keeping the `.` dot in it.

packages/components/pie-toast/src/defs.ts
thejfreitas: Agreed!

packages/components/pie-toast/src/index.ts
thejfreitas: Agreed

packages/components/pie-toast/src/index.ts
thejfreitas: Without it the component ob storybook won't work properly since this piece of code make the component update if one of these properties changes.

packages/components/pie-toast/src/index.ts
```diff
@@ -64,15 +65,67 @@ export class PieToast extends RtlMixin(LitElement) implements ToastProps {
...
+    /**
+     * If the _abortController is set, it aborts all event
+     * listeners in this controller and the controller turns into null.
+     *
+     * @private
+     */
+    private abortAndCleanEventListeners () {
+        if (this._abortController) {
+            this._abortController.abort();
+            this._abortController = null;
```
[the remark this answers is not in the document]
thejfreitas: We don't set the controller to null every time. We set to null when the user closes the toast or when the component reaches the timeout ending its lifecycle. Once the controller aborts all event listeners it turns into null freeing up memory.

packages/components/pie-toast/src/index.ts
thejfreitas: It depends, IMHO leave the check out of it will help to understand that event listeners will be created only when the duration `is not null` at first glance.

kevinrodrigues (approved):

xander-marjoram:
A couple of comments about the storybook preview, I'm happy if you want to address these in a follow-up PR:

1. Please can you use `UseArgs` to link the value of `isOpen` to the current state of the toast? We use it in a few components to create a two-way binding between the control and its value (e.g., `checked` in `pie-radio`), it makes for quite a nice experience, in my opinion. In this case it makes it easier to reopen the toast once it has dismissed itself.
2. I noticed that if you load the toast story with `isDismissible = false`, the close button is still shown: https://pr1862-storybook.pie.design/?path=/story/toast--auto-dismiss&args=isDismissible:!false

packages/components/pie-toast/src/defs.ts
```diff
@@ -78,4 +87,5 @@ export const defaultProps: DefaultProps = {
     isStrong: false,
     isDismissible: true,
     isMultiline: false,
+    duration: defaultDuration,
```
xander-marjoram, line 90: We can probably delete the variable and assign the value directly here.
```suggestion
    duration: 5000,
```

packages/components/pie-toast/src/index.ts
xander-marjoram: Please can you add a comment to the code explaining this?

packages/components/pie-toast/src/toast.scss
```diff
@@ -25,6 +25,24 @@
   color: var(--toast-font-color);
   font-size: var(--toast-font-size);
   line-height: var(--toast-line-height);
+  transition-property: all;
+  transition-duration: var(--dt-motion-timing-100);
+  transition-timing-function: var(-dt-motion-easing-in);
```
xander-marjoram, line 30: Typo, I think?
```suggestion
  transition-timing-function: var(--dt-motion-easing-in);
```

packages/components/pie-toast/src/toast.scss
```diff
@@ -25,6 +25,24 @@
...
+}
+
+.c-toast--animate-in {
+  animation-duration: var(--dt-motion-timing-200);
+  animation-name: animate-in;
+  animation-timing-function: var(-dt-motion-easing-in);
```
xander-marjoram, line 36: And same here
```suggestion
  animation-timing-function: var(--dt-motion-easing-in);
```

packages/components/pie-toast/src/defs.ts
thejfreitas: @xander-marjoram I wanted to avoid "magic numbers" that's why I kept the variable.

apps/pie-storybook/stories/pie-toast.stories.ts
```diff
@@ -65,6 +65,10 @@ const toastStoryMeta: ToastStoryMeta = {
             description: 'The leading action configuration for the toast.',
             control: 'object',
         },
+        duration: {
+            description: 'It sets the duration of the toast in milliseconds before it auto-dismisses.',
```
ashleynolan, line 69:
```suggestion
            description: 'Sets the duration of the toast in milliseconds before it auto-dismisses.',
```

kevinrodrigues (dismissed):

packages/components/pie-toast/src/index.ts
```diff
@@ -64,15 +64,48 @@ export class PieToast extends RtlMixin(LitElement) implements ToastProps {
...
 
+    /**
+     * Create a timeout function and set its id into a private attribute.
+     *
+     * @private
+     */
+    private setAutoDismiss () {
+        this._timeoutId = setTimeout(() => {
+            this.closeToastComponent();
+        }, this.duration || 0 as never);
```
xander-marjoram, line 93: Does this mean that if duration is passed as null, it will fall back to 0 and therefore execute immediately? I know we have `as never` but that seems like something we either shouldn't need to specify, or should actively prevent.

thejfreitas: If the duration is null this method will NEVER be executed. Since TS complains if I don't provide a fallback here, I am adding zero as never since zero should never be a value here.

xander-marjoram: I think this might be another way around it?
```typescript
private setAutoDismiss () {
    if (this.duration === null) return;

    this._timeoutId = setTimeout(() => {
        this.closeToastComponent();
    }, this.duration);
}
```

thejfreitas:
> A couple of comments about the storybook preview, I'm happy if you want to address these in a follow-up PR:
> 
> 1. Please can you use `UseArgs` to link the value of `isOpen` to the current state of the toast? We use it in a few components to create a two-way binding between the control and its value (e.g., `checked` in `pie-radio`), it makes for quite a nice experience, in my opinion. In this case it makes it easier to reopen the toast once it has dismissed itself.
> 2. I noticed that if you load the toast story with `isDismissible = false`, the close button is still shown: https://pr1862-storybook.pie.design/?path=/story/toast--auto-dismiss&args=isDismissible:!false
> 3. If you set the duration to 10000 and refresh the page to restart the timer, then after 5 seconds close and immediately reopen the toast, it will close again once the _original_ 10 seconds is finished. For me the expected behaviour is that the timer should be reset when the toast is reopened.

1. has been solved. 2 and 3 needs more time to investigate which we can involve more people to fix these issues.

xander-marjoram (dismissed):
I'm happy to approve this because it all seems to work well and I don't want to stand in your way too much. I would appreciate if you could address my request to add that comment, but that can be as part of a follow-up PR if you like.

jamieomaguire (dismissed):

packages/components/pie-toast/src/index.ts
thejfreitas: I am sorry I missed this comment. I will add in a following PR for sure. Thank you!

jamieomaguire (approved):

xander-marjoram (approved):

jamieomaguire (approved):

PR #1927, opened 2024-09-19
feat(pie-radio): DSW-2307 add form functionality

xander-marjoram:
- Added form functionality
- Add tests for name and value props
- Add example form story for pie-radio

### Reviewer 1
- [x] I have reviewed the `PIE Storybook` PR preview - Fernando

### Reviewer 2


apps/pie-storybook/stories/pie-radio.stories.ts
```diff
@@ -88,16 +88,75 @@ export default radioStoryMeta;
...
+        <form id="testForm">
+            <pie-radio
+                .value="${value}"
+                name="${ifDefined(name)}"
+                ?checked="${checked}"
+                ?defaultChecked="${defaultChecked}"
+                ?disabled="${disabled}"
+                ?required="${required}"
+                @change="${onChange}">
+                ${sanitizeAndRenderHTML(slot)}
```
fernandofranca, line 138: I see that we don't use the sanitize function on every component with slotted content, is there a rule to determine when to know when use that or is it the case for wrapping the remaining ones in a follow-up?

packages/components/pie-radio/test/component/pie-radio.spec.ts
```diff
@@ -231,4 +289,238 @@ test.describe('PieRadio - Component tests', () => {
             });
         });
     });
+
+    test.describe('Events', () => {
+        test.describe('change', () => {
+            test('should dispatch a change event when radio is clicked that contains the original native event', async ({ mount }) => {
```
fernandofranca, line 295:
```suggestion
            test('when radio is clicked should dispatch a change event that contains the original native event', async ({ mount }) => {
```

packages/components/pie-radio/test/component/pie-radio.spec.ts
```diff
@@ -231,4 +289,238 @@ test.describe('PieRadio - Component tests', () => {
...
+                await page.click('button[type="reset"]');
+
+                const eventElements = await page.evaluate(() => {
+                    const events = document.querySelectorAll('#eventsContainer > div');
+                    return Array.from(events).map((el) => el.innerHTML);
+                });
+
+                // Assert
+                expect(eventElements).toHaveLength(1);
+                expect(eventElements[0]).toBe('change event fired');
```
fernandofranca, line 350: Interesting approach!

fernandofranca (approved):
🔥 Good stuff!!!

fernandofranca:
Sometimes Storybook does some funny stuff when we click the submit button

https://github.com/user-attachments/assets/2c6139ff-cf6c-46fa-9809-fa710686e9fd

packages/components/pie-radio/test/component/pie-radio.spec.ts
xander-marjoram: Thanks! I can't take the credit because these are essentially the same as the checkbox tests. There's quite a lot of duplication overall that makes me wonder if lots of the behaviour/functionality can be pulled out to a shared class...

apps/pie-storybook/stories/pie-radio.stories.ts
xander-marjoram: Good question. It seems to allow you to pass HTML-like text into the slot and it will be rendered as HTML, so I think we should be using that whenever we want the consumer to be able to do the same (which will basically be everywhere, otherwise we may as well just use a string prop instead of a slot!).

| `${sanitizeandRenderHTML(slot)}` | `${slot}` |
| --- | --- |
|  |  |

apps/pie-storybook/stories/pie-button.stories.ts
```diff
@@ -317,8 +317,9 @@ const FormTemplate: TemplateFunction<ButtonProps> = (props: ButtonProps) => html
     </form>
     <script>
         // Display a success message to the user when they submit the form
-        const form = document.querySelector('#testForm');
-        const formLog = document.querySelector('#formLog');
+        var form = document.querySelector('#testForm');
```
xander-marjoram, line 320: There's currently an issue where navigating from one form story to another (or indeed back to the same story) within the same session causes an error where the second form isn't registered because a variable called `form` (or `formLog`) has already been assigned, and storybook doesn't properly dispose of these.

Using `var` allows the variable to be reassigned, preventing the issue.

apps/pie-storybook/stories/pie-radio.stories.ts
fernandofranca: It makes all the sense! Thanks for the detailed answer!

apps/pie-storybook/stories/pie-button.stories.ts
fernandofranca, line 320: I see you replaced a few consts with `var` to avoid the redeclaration error, could it be better to use `let` to signal that this is intended and not a old syntax leftover?

xander-marjoram: I actually tried `let` first but you run into the same issue as with `const`, because the script is re-run from the beginning and you can't re-`let` a variable with the same name. `var` is different because you can use it to redeclare an existing variable, which is effectively what storybook is trying to do. 😞 

It is unfortunate but I thought this was the simplest solution. I could add a comment to explain the `var` usage though!

fernandofranca: Thanks Xander!

fernandofranca (dismissed):

dandel10n (approved):

packages/components/pie-radio/src/index.ts
```diff
@@ -18,19 +19,19 @@ const componentSelector = 'pie-radio';
...
 
-    @property({ type: Boolean })
+    @property({ type: Boolean, reflect: true })
     public disabled = defaultProps.disabled;
 
     @property({ type: String })
     public name: RadioProps['name'];
 
-    @property({ type: Boolean })
+    @property({ type: Boolean, reflect: true })
```
raoufswe, line 34: could you please share our current assumptions about when a property should reflect or not? From a quick search, I noticed that some form components have reflect: true for the required prop, while others do not 👀

PR #1930, opened 2024-09-20
fix(pie-card): DSW-2425 ensure the card is not interactive when it is disabled

raoufswe:

[Fixed] - Ensure the card is not interactive when `disabled` is passed

### Reviewer 1 @kevinrodrigues 

### Reviewer 2

### DIFF
--- a/packages/components/pie-card/src/index.ts
+++ b/packages/components/pie-card/src/index.ts
@@ -56,6 +56,13 @@ export class PieCard extends LitElement implements CardProps {
     @queryAssignedElements({ flatten: true })
     private assignedElements?: HTMLElement[];
 
+    private onClickHandler (event: Event) {
+        if (this.disabled) {
+            event.preventDefault();
+            event.stopPropagation();
+        }
+    }
+
     /**
      * Renders the card as an anchor element.
      *
@@ -63,21 +70,22 @@ export class PieCard extends LitElement implements CardProps {
      */
     private renderAnchor (classes: ClassInfo): TemplateResult {
         const paddingCSS = this.generatePaddingCSS();
+        const {
+            href, rel, target, disabled, aria,
+        } = this;
 
         return html`
             <a
                 class="${classMap(classes)}"
                 data-test-id="pie-card"
-                ?disabled=${this.disabled}
-                href=${this.href || ''}
-                target=${this.target || nothing}
-                rel=${this.rel || nothing}
+                href=${ifDefined(href && !disabled ? href : undefined)}
+                target=${target || nothing}
+                rel=${rel || nothing}
                 role="link"
-                aria-label=${this.aria?.label || nothing}
-                aria-disabled=${this.disabled ? 'true' : 'false'}
+                aria-label=${aria?.label || nothing}
+                aria-disabled=${disabled ? 'true' : 'false'}
                 style=${ifDefined(paddingCSS)}>
-                    <slot></slot>
-                </div>
+                    <slot  @slotchange=${this.handleSlotChange}></slot>
             </a>`;
     }
 
@@ -196,10 +204,11 @@ export class PieCard extends LitElement implements CardProps {
                     class="${classMap(classes)}"
                     data-test-id="pie-card"
                     role="button"
-                    tabindex="0"
+                    tabindex=${disabled ? '-1' : '0'}
                     aria-label=${aria?.label || nothing}
                     aria-disabled=${disabled ? 'true' : 'false'}
-                    style=${paddingCSS || ''}>
+                    style=${paddingCSS || ''}
+                    @click=${this.onClickHandler}>
                         <slot @slotchange=${this.handleSlotChange}></slot>
                     </div>
                 </div>`;
--- a/packages/components/pie-card/src/index.ts
+++ b/packages/components/pie-card/src/index.ts
@@ -58,6 +58,7 @@ export class PieCard extends LitElement implements CardProps {
 
     private onClickHandler (event: Event) {
         if (this.disabled) {
+            // needed to intercept/prevent click events when the card is disabled.
             event.preventDefault();
             event.stopPropagation();
         }
--- a/packages/components/pie-card/test/component/pie-card.spec.ts
+++ b/packages/components/pie-card/test/component/pie-card.spec.ts
@@ -43,41 +43,63 @@ test.describe('PieCard - Component tests', () => {
         await expect(renderedSlotContent).toBeVisible();
     });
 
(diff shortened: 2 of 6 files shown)


packages/components/pie-card/src/index.ts
```diff
@@ -53,28 +53,36 @@ export class PieCard extends LitElement implements CardProps {
     @validPropertyValues(componentSelector, paddingValues, undefined)
     public padding?: CardProps['padding'];
 
+    private onClickHandler (event: Event) {
+        if (this.disabled) {
+            event.preventDefault();
+            event.stopPropagation();
+        }
+    }
```
fernandofranca, line 61: Is this handler really necessary?

packages/components/pie-card/src/index.ts
```diff
@@ -53,28 +53,36 @@ export class PieCard extends LitElement implements CardProps {
...
 
         return html`
             <a
                 class="${classMap(classes)}"
                 data-test-id="pie-card"
-                ?disabled=${this.disabled}
-                href=${this.href || ''}
-                target=${this.target || nothing}
-                rel=${this.rel || nothing}
+                href=${ifDefined(href && !disabled ? href : undefined)}
```
kevinrodrigues, line 78: I think here you can just check if it not disabled i.e `!disabled ? href : undefined` because we only really care about the disabled state and not the href provided as it will be set after the `?` anyway.

packages/components/pie-card/test/component/pie-card.spec.ts
```diff
@@ -304,4 +284,80 @@ test.describe('PieCard - Component tests', () => {
             });
         });
     });
+
+    test.describe('Prop: `disabled" and tag="button"`', () => {
+        test('should have the correct attributes', async ({ mount, page }) => {
```
kevinrodrigues, line 289: Minor: Can you add the attribute the test is expecting to have in the description? e.g `should have the correct attributes: aria-disabled & tabindex`

packages/components/pie-card/test/component/pie-card.spec.ts
```diff
@@ -304,4 +284,80 @@ test.describe('PieCard - Component tests', () => {
...
+            });
+
+            // Assert
+            await expect(component).toBeDisabled();
+            expect(messages).toHaveLength(0);
+        });
+    });
+
+    test.describe('Prop: `disabled" and tag="a"`', () => {
+        test('should have the correct attributes', async ({ mount, page }) => {
```
kevinrodrigues, line 342: Similar to the other comment.

packages/components/pie-card/test/component/pie-card.spec.ts
```diff
@@ -304,4 +284,80 @@ test.describe('PieCard - Component tests', () => {
             });
         });
     });
+
+    test.describe('Prop: `disabled" and tag="button"`', () => {
```
kevinrodrigues, line 288: I would rearrange these tests so that you have one wrapper for `Prop disabled` My pr will have this but i can sort out conflicts. Then within those test you add `tag button and tag a`. That way they have one parent (Prop Disabled).

packages/components/pie-card/test/component/pie-card.spec.ts
```diff
@@ -304,4 +284,80 @@ test.describe('PieCard - Component tests', () => {
...
+                card?.shadowRoot?.querySelector('div')?.click();
+            });
+
+            // Assert
+            await expect(component).toBeDisabled();
+            expect(messages).toHaveLength(0);
+        });
+    });
+
+    test.describe('Prop: `disabled" and tag="a"`', () => {
```
kevinrodrigues, line 341: Place inside the main parent "disabled".

kevinrodrigues: Example: https://github.com/justeattakeaway/pie/pull/1946/files#diff-5375b8f692432c3f08dec3af2f84994cbdb7a8f84ba3123e182485876bd8472aR308

packages/components/pie-card/test/component/pie-card.spec.ts
raoufswe: considering we check for more than aria-disabled & tabindex and potentially more attributes in the future, i think it's  if we keep it genric and more conscious 👀  wdyt?

packages/components/pie-card/test/component/pie-card.spec.ts
raoufswe: done

packages/components/pie-card/src/index.ts
raoufswe: yes, we need to disable the component's click event if `disabled` is passed.
Since we use a div with the role of a button for the card, we have two options to disable the component: intercept the click event or use `pointer-events: none`. I opted in to intercept the event because disabling interactivity with CSS (pointer-events: none) doesn't prevent events triggered by JavaScript/programmability. For example, if you query the element,  and call it using `element,clikc()`.

packages/components/pie-card/test/component/pie-card.spec.ts
raoufswe: same as above

packages/components/pie-card/src/index.ts
fernandofranca: Good one! Would a short comment make sense before `event.preventDefault ` and `event.stopPropagation` so it becomes clear for others and they will not possibly risk removing these lines?

packages/components/pie-card/test/component/pie-card.spec.ts
kevinrodrigues: It's more for visual tests when they fail the description is useful. Currently the description does not tell you what has failed unless you look at the test itself to then see which attributes are the cause. Also not asking for change for this but usually one test suite should include a single expect (which makes the descriptions easier to handle)

packages/components/pie-card/test/component/pie-card.spec.ts
kevinrodrigues: https://github.com/justeattakeaway/pie/pull/1930#discussion_r1773644660

packages/components/pie-card/test/component/pie-card.spec.ts
raoufswe: makes sense, updated 💯

packages/components/pie-card/src/index.ts
raoufswe: done 💯

kevinrodrigues (approved):

fernandofranca (approved):

PR #1951, opened 2024-09-24
refactor(pie-cookie-banner): DSW-2280 small type / linting changes

xander-marjoram:
This was supposed to be a PR updating pie-cookie-banner to use classes instead of attributes for styling, but we weren't using attributes in the first place, so I took the opportunity to do a little refactoring.

### Reviewer 1 @leksaBoiko 

### Reviewer 2


packages/components/pie-cookie-banner/src/localisation-utils.ts
```diff
@@ -1,11 +1,6 @@
-import {
-    TemplateResult,
-} from 'lit';
-
-import {
-    type CookieBannerLocale,
-    type CustomTagEnhancers,
-} from './defs';
+import { type TemplateResult } from 'lit';
```
leksaBoiko, line 1: Looks good to me! Just to clarify, you removed return types and `type TemplateResult` in this file:`packages/components/pie-cookie-banner/src/index.ts`, but didn't remove here, for `enhanceCustomTags` and `localiseRichText` functions. What's the reason for this?

xander-marjoram: Good question! I've taken a look and I guess the short answer is that the return type wasn't providing any value in those other files, because the functions were very simple and the return type is easily inferred because they just return HTML strings.

In this file, the functions are a bit more complex and the types are also used inside the JSDoc comments. They could probably be removed but I don't have a deep enough understanding of the functions at the moment to improve the documentation. If I just removed them we would be losing some potentially helpful comments.

leksaBoiko (approved):

PR #1970, opened 2024-10-07
feat(generator-pie-component): DSW-2423 improve pie generator

fernandofranca:
### pie-monorepo

[Added] - github action for automating the creation of component labels
[Changed] - update ci.yml syntax to match the one the "yaml" package writes

### @justeattakeaway/generator-pie-component

[Added] - automation to update project-labeler.yml
[Added] - automation to add placeholder Percy token to ci.yml
[Added] - automation to update new package dependencies automatically
[Added] - automation to update Storybook with new package
[Changed] - ensure to update `package.json` exports in alphabetical order
[Changed] - updated documentation instructions

## How to test

```
yarn build --filter=generator-pie-component
npx yo @justeattakeaway/pie-component
```


.github/workflows/add-missing-labels/add-missing-labels.js
```diff
@@ -0,0 +1,29 @@
...
+                    name,
+                    color: 'ededed',
+                    owner: context.repo.owner,
+                    repo: context.repo.repo,
+                });
+            }
+        }
+    } catch (error) {
+        console.error(`${scriptPrefix} Failed`);
+        console.error(error.message);
```
fernandofranca, line 27: Catching the error so the whole workflow doesn't fail just because of this step

.github/workflows/add-missing-labels/add-missing-labels.js
```diff
@@ -0,0 +1,29 @@
...
+            console.error(`${scriptPrefix} No components were added, nothing to update.`);
+        } else {
+            for await (const componentName of addedComponentPackages) {
+                const name = componentName.replace(packagePrefix, '');
+
+                console.info(`${scriptPrefix} Adding label for the package "${name}"...`);
+
+                await github.rest.issues.createLabel({
+                    name,
+                    color: 'ededed',
```
jamieomaguire, line 19: I wonder if there is a way to randomise the colour 🤣

ashleynolan (dismissed):

.github/workflows/add-missing-labels/add-missing-labels.js
fernandofranca: Might be rather easy... I thought all components were supposed to follow the same gray for consistence 🤔

xander-marjoram: It's possible, but I'm glad we're going with the consistent colour because all of the other components use this one 😄

xander-marjoram:
@siggerzz I'm pretty sure it fails silently (other than logging it) because I didn't realise for at least a week that the token for pie-radio was missing. I reckon this could be fixed in another PR.

.github/workflows/labeler.yml
```diff
@@ -10,6 +10,22 @@ jobs:
     - name: Checkout
       uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
 
+    - name: Fetch main branch
+      run: git fetch origin main:main
+
+    - name: List local branches
+      run: git branch
```
xander-marjoram, line 17: Why are these steps needed?

siggerzz: I think this can be simplified by doing:

```
- name: Checkout
      uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
      with:
        fetch-depth: 0
```

.github/workflows/add-missing-labels/get-added-component-packages.js
```diff
@@ -0,0 +1,67 @@
...
+    }
+}
+
+/**
+ * Compares the entries of two objects and returns the added keys
+ * @param {object} baseObj dependencies at base brach
+ * @param {object} currentObj dependencies at current branch
+ * @returns Array with the names of added dependencies
+ */
+function getNewDependencies (baseObj = {}, currentObj = {}) {
```
xander-marjoram, line 27: When we upgrade to Node 22 I think we can use a native Set method for this:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/difference

No changes required for now 😄

packages/tools/generator-pie-component/src/app/index.ts
```diff
@@ -45,11 +51,82 @@ export default class extends Generator {
...
+
+    /**
+     * Add placeholder Percy token environment variable to ci.yml
+     */
+    _addPercyTokenEnvVar () {
+        const { percyComponentName } = this.props;
+        const yamlFilePath = this.destinationPath('.github/workflows/ci.yml');
+
+        // Read file
+        const yamlDoc:any = this._readAndParseYaml(yamlFilePath);
```
xander-marjoram, line 81: Can `any` be replaced by the `Document` type from the `yaml` package?

packages/tools/generator-pie-component/README.md
```diff
@@ -42,45 +42,17 @@ An interactive prompt should now be displayed asking for a component name.
...
 
 - Create a Percy project for the component on the Percy website.
-- Update the `test:visual` command in package.json to the following:
-
-```sh
-"test:visual": "run -T cross-env-shell PERCY_TOKEN=${PERCY_TOKEN_PIE_COMPONENT_NAME} percy exec --allowed-hostname cloudfront.net -- npx playwright test -c ./playwright-lit-visual.config.ts",
-```
-
-- Add the token environment variable to `github/workflows/ci.yml` as follows:
+- Update the component token environment variable in `github/workflows/ci.yml` as follows:
```
xander-marjoram, line 48: I don't think this file will need manually updating, I think you've fully automated it 😄

.github/workflows/labeler.yml
jamieomaguire: @siggerzz just curious if using that action is really beneficial? It seems like it adds a dependency that presumably might need to be updated from time to time whereas the commands Fernando added will never change

packages/tools/generator-pie-component/src/app/index.ts
xander-marjoram: Actually I've noticed you're already using that type above so this should be an easy fix 👍🏻

packages/components/pie-webc/src/componentService.js
```diff
@@ -1,3 +1,4 @@
+import path from 'path';
```
xander-marjoram, line 1: I don't think we need this import, we can use `this.path` instead

jamieomaguire (approved):

.github/workflows/labeler.yml
xander-marjoram: @siggerzz I think what Fernando has is good because we don't actually need to check out the main branch, we're just using it to compare to the current branch, unless I'm misunderstanding what the action does

siggerzz: Ah, that's totally valid! I thought we were already using checkout, but I misread. My bad!

.github/workflows/add-missing-labels/get-added-component-packages.js
fernandofranca: Didn't knew that one! Pretty useful!!!

fernandofranca: Added a todo comment

packages/tools/generator-pie-component/README.md
fernandofranca: Updated the docs

packages/components/pie-webc/src/componentService.js
fernandofranca: Good point

xander-marjoram (approved):
There's still one more `any` but you can fix that in a follow-up if you like

jamieomaguire (dismissed):

ashleynolan (approved):

packages/components/pie-webc/test/componentService.spec.js
```diff
@@ -1,3 +1,4 @@
+import path from 'path';
```
xander-marjoram, line 1: Is this import needed in the test?

xander-marjoram (approved):

packages/components/pie-webc/test/componentService.spec.js
fernandofranca: It's there to simplify the "mock" implementation for `path.parse`, otherwise its usage would break tests.

PR #1982, opened 2024-10-10
feat(pie-icon-button): DSW-000 add aria prop to btn and enable browse…

jamieomaguire:
- Adds an `aria` property to the icon button
- Adds a missing `test:browsers` command to the `package.json`
- Fixes the broken a11y tests

### Reviewer 1 - Maryia

### Reviewer 2


siggerzz (approved):

packages/components/pie-icon-button/src/index.ts
```diff
@@ -53,17 +57,23 @@ export class PieIconButton extends LitElement implements IconButtonProps {
...
                 size="${size || 'medium'}"
                 variant="${variant || 'primary'}"
                 ?disabled="${disabled}"
-                ?isLoading="${isLoading}">
+                ?isLoading="${isLoading}"
+                aria-label="${ifDefined(aria?.label)}"
+                aria-labelledby="${ifDefined(aria?.labelledby)}"
+                aria-describedby="${ifDefined(aria?.describedby)}"
+                aria-expanded="${ifDefined(aria?.expanded)}"
+                aria-controls="${ifDefined(aria?.controls)}">
```
raoufswe, line 76: I see the value/use case of adding label and expanded to the icon button, but I'm having a hard time figuring out the use cases of the labelledby, describedby, and controls and was wondering if you can share more about the reasoning ? 🥇

packages/components/pie-icon-button/test/accessibility/pie-icon-button.spec.ts
```diff
@@ -5,26 +5,45 @@ import type {
...
+        label: 'qux',
+    },
 };
 
 const componentPropsMatrix: WebComponentPropValues[] = getAllPropCombinations(props);
 const componentPropsMatrixByVariant: Record<string, WebComponentPropValues[]> = splitCombinationsByPropertyValue(componentPropsMatrix, 'variant');
 const componentVariants: string[] = Object.keys(componentPropsMatrixByVariant);
 
+test.beforeEach(async ({ mount }, testInfo) => {
+    testInfo.setTimeout(testInfo.timeout + 40000);
```
fernandofranca, line 29: I see there other occurrences of this increased setTimeout, do we have plans to tidy up with some shared constant for the cases where the test might take longer?

packages/components/pie-icon-button/test/accessibility/pie-icon-button.spec.ts
```diff
@@ -5,26 +5,45 @@ import type {
...
+
+    // This ensures the icon-button and icon components are registered in the DOM for each test.
+    // It appears to add them to a Playwright cache which we understand is required for the tests to work correctly.
+    const iconButtonComponent = await mount(PieIconButton);
+    await iconButtonComponent.unmount();
+    const iconComponent = await mount(IconClose);
+    await iconComponent.unmount();
```
fernandofranca, line 36: I assume there are other occurrences of this workaround in other test related files.

It would be nice to have a shared function that abstracts this workaround with something less verbose, that could be easily removed if and when it becomes unnecessary.

Something along these lines...
```
async function mountComponentsForPlaywright(componentsClassArray) {
  for await (const component of componentsClassArray) {
    const mountedComponent = await mount(component);
    await mountedComponent.unmount();
  }
}
```

fernandofranca (commented):
Looks good, the main question is Raouf's one about the aria props that might be unnecessary.

packages/components/pie-icon-button/test/accessibility/pie-icon-button.spec.ts
jamieomaguire: I agree and think that would be a nice change. Would you be happy for me to raise a separate PR after this that applies that change to all existing test files @fernandofranca ?

packages/components/pie-icon-button/test/accessibility/pie-icon-button.spec.ts
jamieomaguire: I think this would be good to discuss offline and we can include @JoshuaNg2332 as I am less familiar with why we are doing this

packages/components/pie-icon-button/src/index.ts
jamieomaguire: Sure - this was a direct port of the aria attributes exposed to consumers in Snacks. It was a suggestion from @dandel10n 😃 

Screenshot 2024-10-11 at 09 11 23

raoufswe: I took a look at Snacks and CW, and it seems they aren't used anywhere, so I don’t think they are useful for our use cases and my concern is that copying things over might not be the ideal solution and we should only add what’s needed until we explicitly need to do otherwise.

just for context, I'm bringing this up because I realized we have a couple of unnecessary/incorrect assignments of aria attributes in the codebase, and I will try ti refine a ticket for it next week 🤝

dandel10n: It is used, in snacks iconButton in Stepper for example, the attribute is also used in custom buttons in snacks so I believe it is useful to have.

packages/components/pie-icon-button/test/accessibility/pie-icon-button.spec.ts
```diff
@@ -5,26 +5,45 @@ import type {
...
 
 const props: PropObject = {
     variant: variants,
     disabled: [true, false],
+    aria: {
+        labelledby: 'foo',
+        describedby: 'bar',
+        controls: 'baz',
+        expanded: false,
+        label: 'qux',
```
dandel10n, line 20: I would leave only label here since other attributes are situational

packages/components/pie-icon-button/test/accessibility/pie-icon-button.spec.ts
fernandofranca: That would be great! 👍

packages/components/pie-icon-button/test/accessibility/pie-icon-button.spec.ts
fernandofranca: Created a discussion in our "messaging app".

packages/components/pie-icon-button/src/index.ts
jamieomaguire: @raoufswe is there any harm in supporting these attributes? (provided they are correct usage!)

fernandofranca (approved):

raoufswe (approved):

PR #1989, opened 2024-10-14
fix(pie-monorepo): DSW-000 update the labeler workflow

fernandofranca:

This is a follow-up for https://github.com/justeattakeaway/pie/pull/1970 and https://github.com/justeattakeaway/pie/pull/1983

After merging the previous PRs it was noticed that the labeler workflow was failing and this PR solves the root problem, the context in which the workflow runs.

Since this action runs on "pull_request_target", the current context is the "main" branch ([reference](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_target)), therefore we need to pull the actual branch PR before performing the comparison in the "Add Missing Labels" step

### DIFF
--- a/.changeset/sweet-garlics-eat.md
+++ b/.changeset/sweet-garlics-eat.md
@@ -1,0 +1,7 @@
+---
+"@justeattakeaway/pie-tag": minor
+"pie-storybook": patch
+"pie-docs": patch
+---
+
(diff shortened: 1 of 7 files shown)


raoufswe (approved):
🏅

xander-marjoram (approved):

jamieomaguire (approved):

PR #2000, opened 2024-10-17
format(pie-monorepo): DSW-000 update prop definitions to align with conventions

xander-marjoram:
- Use component props interface for all prop type definitions (e.g., `CardProps['href']` instead of just `string`).
  - Don't use `?` when declaring props, this comes from the interface anyway.
- If the prop has a default value, use that from `defaultProps`, and let TS infer the type.
- Make sure all uses of `@property()` have a `type`.
- `@state` properties should be private and prefixed with an underscore.
- Use `!` when a `@query` property is guaranteed to exist, reducing the complexity of the code.


packages/components/pie-card/src/defs.ts
```diff
@@ -8,7 +8,7 @@ export const paddingValues = spacingTokens.flatMap((first) => [first, ...spacing
 
 type PaddingValue = typeof spacingTokens[number];
 
-export type AriaProps = {
```
xander-marjoram, line 11: We don't export the `AriaProps` types to avoid name collisions across components

packages/components/pie-webc/package.json
```diff
@@ -235,8 +235,6 @@
   "scripts": {
     "lint:scripts": "run -T eslint .",
     "lint:scripts:fix": "yarn lint:scripts --fix",
-    "lint:style": "echo \"Error: no scss / css to lint\" && exit 0",
```
xander-marjoram, line 238: There are no styles in this package so this was logging a warning

packages/components/pie-webc/package.json
```diff
@@ -235,8 +235,6 @@
...
-    "lint:style:fix": "yarn lint:style --fix",
```
xander-marjoram, line 239: There are no styles in this package so this was logging a warning

raoufswe (approved):

dandel10n (approved):

PR #2001, opened 2024-10-17
fix(stylelint-config-pie): DSW-000 add js extension to rule imports for ESM compat

ashleynolan:

- [Changed] - Add file extensions to rule imports

Fixes issue in native ESM environments, where without the prefix, it throws an error.

### DIFF
--- a/packages/tools/stylelint-config-pie/index.js
+++ b/packages/tools/stylelint-config-pie/index.js
@@ -1,6 +1,7 @@
-import baseRules from './rules/base';
-import strictRules from './rules/strict';
-import orderingRules from './rules/ordering';
+/* eslint-disable import/extensions */
+import baseRules from './rules/base.js';
+import strictRules from './rules/strict.js';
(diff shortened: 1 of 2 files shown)


packages/tools/stylelint-config-pie/index.js
```diff
@@ -1,6 +1,7 @@
-import baseRules from './rules/base';
-import strictRules from './rules/strict';
-import orderingRules from './rules/ordering';
+/* eslint-disable import/extensions */
```
jamieomaguire, line 1: Does this imply we should be updating our eslint config?

ashleynolan: Perhaps – I didn't atm just as we currently have quite a few of the same exemption in PIE, so I figure it's probably a wider conversation than doing git as part of this.

jamieomaguire (approved):

xander-marjoram (approved):

PR #2005, opened 2024-10-18
Proof of Concept: Monitor Component bundle sizes

jamieomaguire:

Allows us to set max limits (in KB) for each of our components, and runs a CI check that fails if any exceed the limits. If they do, all components will be mentioned in a PR comment which contains the failures in a table. It also logs out the gzip and brotli compressed sizes.

I wrote [the package](https://www.npmjs.com/package/compsizer) to prototype it quickly however there's no reason we can't fork it and move into the monorepo if we wish.

CI Behaviour:
- Lists all failing components
- Adds a comment on failiure
- Updates the existing comment if previous runs added one
- Deletes the existing comment if it previously failed and now passes

#### The new CI steps (only run on the Ubuntu Node 20 build)
Screenshot 2024-11-01 at 10 53 24

#### When all components pass
Screenshot 2024-11-01 at 10 20 18

#### PR Comment left when components fail
Screenshot 2024-11-01 at 10 36 58

#### CI logs when a component fails
Screenshot 2024-11-01 at 10 59 14


maledr5 (commented):
Looking good thanks for putting this together. My only nitpick is I wonder if we could have all the config in 1 file per component and not 2. But other than that looks like a good approach 👍

packages/components/pie-modal/component-bundle-sizes.json
```diff
@@ -0,0 +1,5 @@
+{
+  "pie-modal/index.js": 25075,
```
jamieomaguire, line 2: this file is auto generated by the plugin to store the bundle sizes to check against the next time the command is run

maledr5: Thanks for clarifying. No issues from me then

jamieomaguire: Cheers! Worth nothing we don't necessarily need this file as the script will still just check if the size exceeds the threshold defined in the json config. 

This generated file is just to help understand exactly what the previous/current sizes are :)

package.json
```diff
@@ -22,6 +22,7 @@
     "generate-component-statuses": "./utilities/component-statuses/generate-component-statuses.js"
   },
   "scripts": {
+    "analyse-component-bundles": "turbo run analyse-component-bundles --filter=\"./packages/components/**\" --filter=!pie-webc --filter=!pie-webc-testing --filter=!pie-webc-core",
```
siggerzz, line 25: No need for these filters, as turbo by default only executes tasks if present in a package.json

siggerzz (commented):
minor comment!

packages/components/pie-modal/compsizer.config.json
```diff
@@ -0,0 +1,20 @@
...
+    "pie-modal": {
+      "maxSize": "50 KB",
+      "warnOnIncrease": "10%",
+      "distFolderLocation": "./dist"
+    }
+  },
+  "defaults": {
+    "warnOnIncrease": "5%"
+  }
+}
```
jamieomaguire, line 20: going to look to making most of these default values in the tool when not provided. it'll make the config object much smaller

siggerzz:
Closing as the changes to Bundlewatch have been merged

PR #2018, opened 2024-10-31
fix(pie-textarea): DSW-2539 replace label prop with a pie form label

raoufswe:

This PR is a draft version of how we can replace the label prop in the `pie-textarea` component with an approach similar to the `pie-text-field`, where consumers are expected to import and use the `pie-form-label` component separately to render the label.

The following are some key points to consider outlined in the spike ticket with answers: 

- How should maxLength work if the label is no longer inside the textarea?
The presentation layer can be offloaded to the consumer, where they need to pass the text for displaying the total remaining characters to the `trailing` prop of the `form-label` component. The validation logic should be handled by browsers using the `maxlength` attribute ([MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/maxlength)) when it's passed as a prop. 

- Is it realistic to expect consumers to handle this including formatting and any RTL logic? (e.g., 0/50 should still appear the same in RTL, rather than 50/0).
I believe yes, consumers just need to listen to the change or input event and use the value prop's length to determine the current character count similar to how it's handled in the label story in this PR. About RTL, browsers automatically maintain the order of the text and won’t swap it

Screenshot 2024-11-01 at 15 45 53

- Should maxLength be removed from pie-textarea?
no, because we need to offload the max length validation to the browser by relying on the [maxlength](https://www.w3schools.com/tags/att_textarea_maxlength.asp) attribute 

Overall, the approach seems feasible and will simplify the component while providing a consistent API similar to other form elements.
 
### Reviewer 1

### Reviewer 2

### DIFF
--- a/apps/pie-storybook/stories/pie-textarea.stories.ts
+++ b/apps/pie-storybook/stories/pie-textarea.stories.ts
@@ -1,4 +1,4 @@
-import { html } from 'lit';
+import { html, nothing } from 'lit';
 import { ifDefined } from 'lit/directives/if-defined.js';
 import { action } from '@storybook/addon-actions';
 import { useArgs as UseArgs } from '@storybook/preview-api';
@@ -28,7 +28,6 @@ const Template = ({
     name,
     autocomplete,
     autoFocus,
-    label,
     maxLength,
     assistiveText,
     status,
@@ -67,7 +66,6 @@ const Template = ({
             ?readonly="${readonly}"
             ?required="${required}"
             maxLength="${ifDefined(maxLength)}"
-            label="${ifDefined(label)}"
             @input="${onInput}"
             @change="${onChange}"
             assistiveText="${ifDefined(assistiveText)}"
@@ -167,15 +165,8 @@ const textareaStoryMeta: TextareaStoryMeta = {
                 summary: 'off',
             },
         },
-        label: {
-            description: 'The label for the textarea field.',
-            control: 'text',
-            defaultValue: {
-                summary: defaultProps.label,
-            },
-        },
         maxLength: {
-            description: 'The maximum number of characters allowed in the textarea field. To apply a length restriction, you must also provide label text.',
+            description: 'The maximum number of characters allowed in the textarea field.',
             control: 'number',
             defaultValue: {
                 summary: 0,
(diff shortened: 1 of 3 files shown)


packages/components/pie-textarea/src/index.ts
```diff
@@ -133,16 +128,6 @@ export class PieTextarea extends FormControlMixin(RtlMixin(LitElement)) implemen
         this._throttledResize();
     }
 
-    private restrictInputLength () {
```
raoufswe, line 136: If we decide to keep the current `label` prop approach, I think we should consider leveraging max-length as browsers will do that natively.

jamieomaguire:
Thanks for this!

I don't know if passing the maxLength calculations on to the consumer is the right thing to do. It feels like something we should be able to provide out of the box.

Out of interest, what do our friends in the mobile team do?

apps/pie-storybook/stories/pie-textarea.stories.ts
```diff
@@ -234,15 +225,24 @@ const ExampleFormTemplate: TemplateFunction<TextareaProps> = () => html`
     </form>
 `;
 
+const WithLabelTemplate: TemplateFunction<TextareaProps> = (props: TextareaProps) => html`
+        <p>Please note, the label is a separate component. See <pie-link href="/?path=/story/form-label">pie-form-label</pie-link>.</p>
+        <pie-form-label for="${ifDefined(props.name)}" trailing="${props.maxLength ? `${props.value.length}/${props.maxLength}` : nothing}">Label</pie-form-label>
```
raoufswe, line 230: > I don't know if passing the maxLength calculations on to the consumer is the right thing to do. It feels like something we should be able to provide out of the box.

@jamieomaguire, just in case it wasn't very clear. the actual validation will be done using the max length attribute by browsers. users just need to construct the text and I don't think that's complex considering why we are doing the spike

raoufswe:
@RLumleyJET, @xander-marjoram, and I reviewed the max-length behavior for the textarea component and we've decided to let consumers handle the max-length pattern directly instead of internally handing it. The reasons are:

- based on the PIE specs, the max-length behavior doesn’t prevent users from typing beyond the limit, so the validation logic currently in the component is no longer necessary and can be removed. 
- When users exceed the max length, the component’s status should change to error, and a message should display the remaining character count, similar to how it’s implemented in the Android library. since we don't control the character limit text copy, and users will need to handle the logic to set the status to error anyway, it's more feasible to be documented as a pattern instead of built-in and.

https://github.com/user-attachments/assets/198681aa-dae7-4769-a792-9b4fd6bc4dee


Follow-up tickets will be created for:
- Adding a story example that demonstrates the max-length pattern.
- Cleaning up label related code.

PR #2023, opened 2024-11-04
refactor(pie-tag): DSW-2467 update variant names (breaking change)

raoufswe:
## Describe your changes (can list changeset entries if preferable)

[Updated] - (Breaking change) rename the following variants:
    - blue to information
    - green to success
    - red to error
    - yellow to brand-05
    - brand to brand-02 

📓 docs will updated later in DSW-2324

## Author Checklist (complete before requesting a review)
- [x] I have performed a self-review of my code
- [x] I have reviewed the `PIE Storybook`/`PIE Docs` PR preview
- [x] I have reviewed visual test updates properly before approving
- [x] If changes will affect consumers of the package, I have created a changeset entry.
- [x] If a changeset file has been created, I have used the `/snapit` functionality to test my changes in a consuming application

### Reviewer 1
- [x] I have reviewed the `PIE Storybook`/`PIE Docs` PR preview
- [x] If there are visual test updates, I have reviewed them

### Reviewer 2
- [x] I have reviewed the `PIE Storybook`/`PIE Docs` PR preview
- [x] If there are visual test updates, I have reviewed them

PR #2030, opened 2024-11-07
feat(pie-monorepo): DSW-000 bump some components to v1

jamieomaguire:
- Bumps some components to version `1.0.0`

- [x] I have reviewed the `PIE Docs` PR preview

### Reviewer 2 @raoufswe 

### DIFF
--- a/packages/components/pie-tag/package.json
+++ b/packages/components/pie-tag/package.json
@@ -13,7 +13,7 @@
     "**/*.d.ts"
   ],
   "pieMetadata": {
-    "componentStatus": "beta"
+    "componentStatus": "stable"
   },
   "scripts": {
     "build": "run -T vite build",
--- a/packages/components/pie-tag/package.json
+++ b/packages/components/pie-tag/package.json
@@ -13,7 +13,7 @@
     "**/*.d.ts"
   ],
   "pieMetadata": {
-    "componentStatus": "stable"
+    "componentStatus": "beta"
   },
   "scripts": {
     "build": "run -T vite build",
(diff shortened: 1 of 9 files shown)


siggerzz (approved):

raoufswe (approved):

siggerzz (approved):

xander-marjoram:
This should also trigger a docs deployment because the component statuses are changing.
For now I've checked locally and the statuses do seem to have been updated, so I'm happy with this PR.

xander-marjoram (dismissed):

packages/components/pie-tag/package.json
```diff
@@ -13,7 +13,7 @@
     "**/*.d.ts"
   ],
   "pieMetadata": {
-    "componentStatus": "beta"
+    "componentStatus": "stable"
```
raoufswe, line 16: we are introducing a breaking change for the tag component on this PR
https://github.com/justeattakeaway/pie/pull/2023, should we update it there instead so we don't have to release two majors?

xander-marjoram: Good point. I think we can make the "major" change in both places but we should make sure to release them as part of the same changeset.

xander-marjoram: Although actually I would argue that `pie-tag` is not stable if we're things like this...

jamieomaguire: great spot @raoufswe - reverted tag

raoufswe (approved):

xander-marjoram (approved):

ashleynolan (approved):

siggerzz (approved):

PR #2037, opened 2024-11-08
fix(pie-monorepo): DSW-2579 update labeler workflow

fernandofranca:

- Reverts the `labeler.yml` workflow to perform only basic labeling
- Splits the automation in two new workflows:
  - `labeler-get-labels.yml`
    - runs without extra permissions as it runs with untrusted code
    - uploads artifact with the PR details: new label(s) and pr number
  - `labeler-set-labels.yml` 
    - uses the artifact generated by the `labeler-get-labels.yml` workflow
    - creates and assigns labels to the original PR
    - uses extra permissions to perform the steps aforementioned


jamieomaguire (approved):

.github/workflows/labeler/set-labels.js
```diff
@@ -0,0 +1,63 @@
+const fs = require('fs');
+const { execSync } = require('child_process');
+
+module.exports = async ({ github, context }) => {
+    const { repo : { owner, repo }} = context;
+
+    const allArtifacts = await github.rest.actions.listWorkflowRunArtifacts({
```
siggerzz, line 7: Nice! I was unsure how we would manage the download artifacts for the correct workflow! 🎉

siggerzz (approved):

.github/workflows/labeler/set-labels.js
fernandofranca: It's nice that the `workflow_run` has the implicit id of the original workflow.

PR #2080, opened 2024-11-19
ci(pie-monorepo): DSW-2581 update changeset config for dependabot

siggerzz:
- Updates Dependabot commit messages to have a nicer prefix.
- Fixes an issue with the automatic changeset generation.


.github/dependabot.yml
```diff
@@ -10,7 +10,7 @@ updates:
         update-types: ['version-update:semver-major']
     open-pull-requests-limit: 3
     commit-message:
-      prefix: "deps(npm): DSW-000 update"
+      prefix: "deps(npm): DSW-000"
```
jamieomaguire, line 13: before it was clear it was a dependency update, how come you want to remove the work?

siggerzz: It looks like dependabot naming is like so: `deps(github-actions): DSW-000 update: Bump slackapi/slack-github-action from 1.27.0 to 1.27.1 in /.github/workflows`, so the `update` isn't needed IMO

raoufswe (approved):

jamieomaguire (approved):

PR #2098, opened 2024-11-27
feat(pie-toast-provider): DSW-2222 toast provider basic functionality

raoufswe:
- Adds the ability to create a toast via the provider
- Adds the ability to clear toasts via the provider
- The provider should handle the order priority when tests are added
- [Docs (Draft)](https://pr2104-docs.pie.design/components/toast/toast-provider/code/)

---

## Testing
[How do I test my changes?](https://github.com/justeattakeaway/pie/wiki/PIE-Aperture)

| Task                   | Link                             |
|------------------------|----------------------------------|
| Aperture PR            | [🔗](https://github.com/justeattakeaway/pie-aperture/pull/228) |
| NextJS 14 deployment   | [🔗](https://pr228-aperture-nextjs-app-v14.pie.design) |
| Nuxt 3 deployment      | [🔗](https://pr228-aperture-nuxt-app.pie.design/components/toast-provider) |
| Vanilla deployment     | [🔗](#) |

### Reviewer 1

### Reviewer 2

### DIFF
--- a/apps/pie-storybook/stories/pie-toast-provider.stories.ts
+++ b/apps/pie-storybook/stories/pie-toast-provider.stories.ts
@@ -1,19 +1,39 @@
 import { html } from 'lit';
 import { type Meta } from '@storybook/web-components';
+import { action } from '@storybook/addon-actions';
 
-import '@justeattakeaway/pie-toast-provider';
-import { type ToastProviderProps } from '@justeattakeaway/pie-toast-provider';
+import { toaster } from '@justeattakeaway/pie-toast-provider';
+import { type ToastProviderProps, defaultProps } from '@justeattakeaway/pie-toast-provider';
+import '@justeattakeaway/pie-button';
+import '@justeattakeaway/pie-tag';
 
 import { createStory } from '../utilities';
 
 type ToastProviderStoryMeta = Meta<ToastProviderProps>;
 
-const defaultArgs: ToastProviderProps = {};
+const defaultArgs: ToastProviderProps = {
+    ...defaultProps,
+    options: {
+        duration: 3000,
+        isDismissible: true,
+        onPieToastOpen: action('onPieToastOpen'),
+        onPieToastClose: action('onPieToastClose'),
+        onPieToastLeadingActionClick: action('onPieToastLeadingActionClick'),
+    },
+};
 
 const toastProviderStoryMeta: ToastProviderStoryMeta = {
     title: 'Toast Provider',
     component: 'pie-toast-provider',
-    argTypes: {},
+    argTypes: {
+        options: {
+            description: 'Default options for all toasts; accepts all toast props.',
+            control: 'object',
+            defaultValue: {
+                summary: defaultProps.options,
+            },
+        },
+    },
     args: defaultArgs,
     parameters: {
         design: {
@@ -25,10 +45,85 @@ const toastProviderStoryMeta: ToastProviderStoryMeta = {
 
 export default toastProviderStoryMeta;
 
-// TODO: remove the eslint-disable rule when props are added
-// eslint-disable-next-line no-empty-pattern
-const Template = ({}: ToastProviderProps) => html`
-    <pie-toast-provider></pie-toast-provider>
+const Template = ({ options }: ToastProviderProps) => {
+    const onQueueUpdate = (event: CustomEvent) => {
+        const queueLength = document.querySelector('#queue-length-tag') as HTMLElement;
+        if (queueLength) {
+            queueLength.textContent = `Toast Queue Length: ${event.detail.length}`;
+        }
+    };
+
+    return html`
+    <pie-toast-provider 
+        .options=${options} 
+        @pie-toast-provider-queue-update=${onQueueUpdate}>
+    </pie-toast-provider>
+
+    <pie-tag id="queue-length-tag" variant="information" style="margin-top: 16px;">
+        Toast Queue Length: 0
+    </pie-tag>
+
+    <div style="margin-top: 16px; display: flex; gap: 16px; flex-wrap: wrap;">
+        <pie-button 
+            @click=${() => {
+        toaster.create({
+            message: 'Low Priority Info',
+            variant: 'info',
+        });
+    }}>
+            Trigger Info Toast (Low Priority)
+        </pie-button>
+
+        <pie-button
+            @click=${() => {
+        toaster.create({
+            message: 'Medium Priority Warning Toast',
+            variant: 'warning',
+        });
+    }}>
+            Trigger Warning Toast (Medium Priority)
+        </pie-button>
+
+        <pie-button
+            @click=${() => {
+        toaster.create({
+            message: 'High Priority Error Toast',
+            variant: 'error',
+        });
+    }}>
+            Trigger Error Toast (High Priority)
+        </pie-button>
+
+        <pie-button
+            @click=${() => {
+        toaster.create({
+            message: 'Actionable Info Toast',
+            variant: 'info',
+            leadingAction: { text: 'Retry' },
+        });
+    }}>
+            Trigger Actionable Info Toast
+        </pie-button>
+
+        <pie-button
+            @click=${() => {
+        toaster.create({
+            message: 'Persistent Toast',
+            duration: null,
+        });
+    }}>
+            Trigger Persistent Toast
+        </pie-button>
+
+        <pie-button
+            variant="secondary"
+            @click=${() => {
+        toaster.clearAll();
+    }}>
+            Clear All Toasts
+        </pie-button>
+    </div>
 `;
+};
 
 export const Default = createStory<ToastProviderProps>(Template, defaultArgs)();
--- a/apps/pie-storybook/stories/testing/pie-chip.test.stories.ts
+++ b/apps/pie-storybook/stories/testing/pie-chip.test.stories.ts
@@ -1,6 +1,5 @@
 import { html, nothing } from 'lit';
 import { ifDefined } from 'lit/directives/if-defined.js';
-import { action } from '@storybook/addon-actions';
 import { type Meta } from '@storybook/web-components';
 
 import '@justeattakeaway/pie-chip';
@@ -88,11 +87,11 @@ const chipStoryMeta: ChipStoryMeta = {
 export default chipStoryMeta;
 
 const clickAction = () => {
-  console.log('pie-chip clicked');
-}
+    console.log('pie-chip clicked');
+};
 const closeAction = () => {
-  console.log('pie-chip-close clicked');
-}
+    console.log('pie-chip-close clicked');
+};
 
 const Template: TemplateFunction<ChipProps> = ({
     aria,
--- a/packages/components/pie-toast-provider/package.json
+++ b/packages/components/pie-toast-provider/package.json
@@ -41,6 +41,7 @@
     "cem-plugin-module-file-extensions": "0.0.5"
   },
   "dependencies": {
+    "@justeattakeaway/pie-toast": "0.5.0",
     "@justeattakeaway/pie-webc-core": "0.24.2"
   },
   "volta": {
--- a/packages/components/pie-toast-provider/src/defs.ts
+++ b/packages/components/pie-toast-provider/src/defs.ts
@@ -1,3 +1,54 @@
-// TODO - please remove the eslint disable comment below when you add props to this interface
-// eslint-disable-next-line @typescript-eslint/no-empty-interface
-export interface ToastProviderProps {}
+import { type ToastProps } from '@justeattakeaway/pie-toast';
+
+import { type ComponentDefaultProps } from '@justeattakeaway/pie-webc-core';
+
+export const PRIORITY_ORDER: { [x: string]: number } = {
+    'error-actionable': 1,
+    error: 2,
+    'warning-actionable': 3,
+    'success-actionable': 4,
+    'info-actionable': 5,
+    'neutral-actionable': 6,
+    warning: 7,
+    success: 8,
+    info: 9,
+    neutral: 10,
+};
+
+export interface ExtendedToastProps extends ToastProps {
+    /**
+     * Triggered when the user interacts with the close icon or when the toast auto dismiss.
+     */
+    onPieToastClose?: () => void;
+
+    /**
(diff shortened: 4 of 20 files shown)


apps/pie-storybook/stories/pie-toast-provider.stories.ts
```diff
@@ -25,10 +45,85 @@ const toastProviderStoryMeta: ToastProviderStoryMeta = {
...
+        Toast Queue Length: 0
+    </pie-tag>
+
+    <div style="margin-top: 16px; display: flex; gap: 16px; flex-wrap: wrap;">
+        <pie-button 
+            @click=${() => {
+        toaster.create({
+            message: 'Low Priority Info',
+            variant: 'info',
+        });
```
fernandofranca, line 72: Would it be ok to move the indentation of these buttons click handlers content?

```suggestion
        <pie-button 
            @click=${() => {
              toaster.create({
                  message: 'Low Priority Info',
                  variant: 'info',
              });
```

packages/components/pie-toast-provider/src/index.ts
```diff
@@ -1,20 +1,144 @@
...
 export * from './defs';
+export { toaster } from './toaster';
 
 const componentSelector = 'pie-toast-provider';
 
 /**
  * @tagname pie-toast-provider
+ * @event {CustomEvent} pie-toast-provider-queue-update - when a toast is added or removed from the queue.
  */
 export class PieToastProvider extends RtlMixin(LitElement) implements ToastProviderProps {
```
fernandofranca, line 35: What happens when a consumer mistakenly adds a second Toast Provider to an app?

apps/pie-storybook/stories/pie-toast-provider.stories.ts
raoufswe: it's auto-formatted on save for me 🙄 are you getting the same thing locally? 

https://github.com/user-attachments/assets/9c34bdb3-4013-474b-8017-0a49ee6ea25a

packages/components/pie-toast-provider/src/index.ts
fernandofranca: I understand that is expected to have a single provider at the moment, that's why I ask what happens if the user does a mistake. Everything still works perfectly when more than one provider is in the DOM? Should we help the user to avoid this mistake?

raoufswe: added some error logs and validation to ensure only one toast provider is used in the apps.

packages/components/pie-toast-provider/src/index.ts
```diff
@@ -1,20 +1,143 @@
...
+            return priorityA - priorityB;
+        });
+
+        if (!this._currentToast) {
+            this._showNextToast();
+        }
+    }
+
+    /**
+   *
```
fernandofranca, line 111:
```suggestion
   /**
```

packages/components/pie-toast-provider/src/toaster.ts
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Singleton toaster interface for global access.
+ */
+export const toaster = {
+    _getToastProvider (): PieToastProvider | null {
+        const toastProviders = document.querySelectorAll('pie-toast-provider');
+
+        if (toastProviders.length === 0) {
+            console.error('ToastProvider is not initialized.');
```
fernandofranca, line 12:
```suggestion
            console.error('The pie-toast component requires a pie-toast-provider element present in the DOM.');
```
I believe this would be more helpful, as it informs what is missing or expected.

packages/components/pie-toast-provider/src/index.ts
```diff
@@ -1,20 +1,143 @@
...
+        return PRIORITY_ORDER[key];
+    }
+
+    /**
+     * Handles the dismissal of the current toast and displays the next one in the queue (if any).
+     */
+    private _dismissToast () {
+        this._currentToast?.onPieToastClose?.();
+        this._currentToast = null;
+        requestAnimationFrame(() => { this._showNextToast(); });
```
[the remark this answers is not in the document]
raoufswe: without `requestAnimationFrame`, the `_dismissToast` method will trigger immediate synchronous updates to both the dismissal of the current toast and the display of the next toast in the queue and nothing will show up in the UI cause they happened instantly. it also ensures  the in / out animation of the toast has finished before rendering the next one based on the duration prop to avoid any glitch.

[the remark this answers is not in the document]
fernandofranca: Sounds reasonable! Thanks!

fernandofranca (dismissed):
Good stuff!!!

packages/components/pie-toast-provider/test/component/pie-toast-provider.spec.ts
```diff
@@ -1,5 +1,10 @@
 import { test, expect } from '@sand4rt/experimental-ct-web';
 import { PieToastProvider } from '../../src/index.ts';
+import {
```
siggerzz, line 3: Would it be possible to follow the new approach like we discussed earlier? Doing so will mean we have one less component to refactor 🙏🏼

siggerzz (changes requested):
Small comment regarding testing, but looks good otherwise!

siggerzz (dismissed):

fernandofranca (dismissed):

apps/pie-storybook/stories/testing/pie-chip.test.stories.ts
```diff
@@ -88,11 +87,11 @@ const chipStoryMeta: ChipStoryMeta = {
...
 const clickAction = () => {
-  console.log('pie-chip clicked');
-}
+    console.log('pie-chip clicked');
+};
 const closeAction = () => {
-  console.log('pie-chip-close clicked');
-}
+    console.log('pie-chip-close clicked');
+};
```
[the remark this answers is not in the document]
raoufswe: yea sure thing! i will address it in my follow up PR if you don't mind 🙇

siggerzz (approved):

fernandofranca (approved):

packages/components/pie-toast-provider/package.json
```diff
@@ -41,6 +41,7 @@
     "cem-plugin-module-file-extensions": "0.0.5"
   },
   "dependencies": {
+    "@justeattakeaway/pie-toast": "0.5.0",
```
xander-marjoram, line 44: I think this should also include button and tag if they're being used in the stories. However this does add a dependency when building, do we need to use our components for this?

packages/components/pie-toast-provider/src/defs.ts
```diff
@@ -1,3 +1,54 @@
...
+    success: 8,
+    info: 9,
+    neutral: 10,
+};
+
+export interface ExtendedToastProps extends ToastProps {
+    /**
+     * Triggered when the user interacts with the close icon or when the toast auto dismiss.
+     */
+    onPieToastClose?: () => void;
```
xander-marjoram, line 22: Is there a reason we decided to use function props here rather than emitting events? Also what happens if there is more than one toast?

apps/pie-storybook/stories/testing/pie-chip.test.stories.ts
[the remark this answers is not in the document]
xander-marjoram: How about using the storybook actions addon?

packages/components/pie-toast-provider/package.json
raoufswe: button and tag aren't part of the component code itself so I don't think we should include storybook-related dependencies to the component dependency tree. at least we didn't do it for all other components. 

https://github.com/justeattakeaway/pie/blob/main/packages/components/pie-toast-provider/src/index.ts#L15

packages/components/pie-toast-provider/src/defs.ts
raoufswe: yes, users will interface with the component using the [toaster singleton](https://github.com/justeattakeaway/pie/blob/main/apps/pie-storybook/stories/pie-toast-provider.stories.ts#L89) which accepts an object for all toast props.

we are only passing props to the toast component which does emit the event internally so no new events were created.

raoufswe: this is an exmaple in aperture if it helps 
https://github.com/justeattakeaway/pie-aperture/pull/228/files#diff-b30a824bb84fa94b7a399a009b0bc5d1e67556bddec47fa26dc717e2de07a0faR21-R24 

options is the global configuration of all toasts controlled by the provider but individual calls of the toaster supports that too.

packages/components/pie-toast-provider/package.json
xander-marjoram: I think you're right actually, given that Storybook already depends on all the components. I was thinking of some components which have extra dependencies because of other components being imported in test files

packages/components/pie-toast-provider/test/component/pie-toast-provider.spec.ts
```diff
@@ -1,17 +1,147 @@
...
+                });
+            });
+
+            // Assert
+            const queueVariants = toastsQueue.map((toast: ExtendedToastProps) => `${toast.variant}${toast.leadingAction ? '-actionable' : ''}`);
+            for (let i = 1; i < queueVariants.length; i++) {
+                const prevPriority = PRIORITY_ORDER[queueVariants[i - 1]];
+                const currPriority = PRIORITY_ORDER[queueVariants[i]];
+                expect(currPriority).toBeGreaterThanOrEqual(prevPriority); // Ensure the current has a higher priority
+            }
```
xander-marjoram, line 75: I wonder if a cleaner way might be to `sort` by priority and check that the ordering is the same?

packages/components/pie-toast-provider/src/index.ts
```diff
@@ -1,20 +1,143 @@
...
+} from '@justeattakeaway/pie-webc-core';
+import '@justeattakeaway/pie-toast';
 import styles from './toast-provider.scss?inline';
-import { type ToastProviderProps } from './defs';
+import {
+    defaultProps,
+    PRIORITY_ORDER,
+    type ToastProviderProps,
+    type ExtendedToastProps,
+    ON_TOAST_PROVIDER_QUEUE_UPDATE_EVENT,
```
xander-marjoram, line 22: Nit: I don't think we need `_EVENT` as that's already implied by `ON_`

packages/components/pie-toast-provider/src/defs.ts
```diff
@@ -1,3 +1,54 @@
...
+    error: 2,
+    'warning-actionable': 3,
+    'success-actionable': 4,
+    'info-actionable': 5,
+    'neutral-actionable': 6,
+    warning: 7,
+    success: 8,
+    info: 9,
+    neutral: 10,
+};
```
xander-marjoram, line 16: I just tried this out locally and it seems to give us better intellisense and avoids using `string` which feels too generic. I was considering asking you to add a string union of the possible priorities too but actually if we use this `as const` we can instead get a string union from `keyof typeof PRIORITY_ORDER` if needed...
```typescript
export const PRIORITY_ORDER = {
    'error-actionable': 1,
    error: 2,
    'warning-actionable': 3,
    'success-actionable': 4,
    'info-actionable': 5,
    'neutral-actionable': 6,
    warning: 7,
    success: 8,
    info: 9,
    neutral: 10,
} as const;
```

packages/components/pie-toast-provider/src/index.ts
```diff
@@ -1,20 +1,143 @@
...
 /**
  * @tagname pie-toast-provider
+ * @event {CustomEvent} pie-toast-provider-queue-update - when a toast is added or removed from the queue.
  */
 export class PieToastProvider extends RtlMixin(LitElement) implements ToastProviderProps {
+    @state()
+    private _toasts: ExtendedToastProps[] = [];
+
+    @state()
+    private _currentToast: ExtendedToastProps | null = null;
```
xander-marjoram, line 40: If we still need these `ExtendedToastProps`, should they live inside the toast component instead of the provider?

xander-marjoram: It feels like this line should say
```typescript
private _currentToast: PieToast | null = null;
```
and same for `_toasts`; it feels like that should probably be of type `PieToast[]`.

packages/components/pie-toast-provider/test/visual/pie-toast-provider.spec.ts
```diff
@@ -1,10 +1,15 @@
...
 
 test.describe('PieToastProvider - Visual tests`', () => {
-    test('should display the PieToastProvider component successfully', async ({ page, mount }) => {
-        await mount(PieToastProvider);
+    test('should display the PieToastProvider component successfully', async ({ page }) => {
+        const basePage = new BasePage(page, 'toast-provider--default');
+
+        basePage.load();
+
+        // Follow up to remove in Jan
```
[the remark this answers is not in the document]
raoufswe: i have discussed this with @siggerzz and @JoshuaNg2332 and they have asked to leave it for now

packages/components/pie-toast-provider/src/index.ts
raoufswe: that current convention in the codebase is to postfix them with _EVENT so I would stick with it for consistency. 

Screenshot 2025-01-08 at 15 10 34

packages/components/pie-toast-provider/src/index.ts
raoufswe: we need to extend the `PieToast` interface to include the events so they can be passed to the `createToast` method, otherwise the events won't be recognised by typescript and we don't have a way to trigger/pass the events to the toast component. 

https://github.com/justeattakeaway/pie/blob/main/packages/components/pie-toast-provider/src/index.ts#L134-L136

```
createToast({
    onPieToastClose: () => log(),
    onPieToastOpen: () => log(),
    onPieToastLeadingActionClick: () => log()
})

```

packages/components/pie-toast-provider/src/defs.ts
raoufswe: will be addressed in the docs PR 🤝

PR #2151, opened 2025-01-13
format(pie-modal): DSW-000 fix linting

siggerzz:


apps/pie-storybook/stories/testing/pie-modal.test.stories.ts
```diff
@@ -18,138 +18,138 @@ type ModalProps = SlottedComponentProps<ModalPropsBase>;
...
-  },
 };
 
 export default modalStoryMeta;
 
 const supportingClickAction = () => {
-  /* eslint-disable-next-line no-console */
-  console.log('supporting-click');
+    /* eslint-disable-next-line no-console */
+    console.log('supporting-click');
```
fernandofranca, line 126: Could you replace the console.log with console.info, so we can get rid of the eslint override comments?

fernandofranca (approved):

PR #2168, opened 2025-01-21
feat(pie-thumbnail): DSW-2583 2584 2585 Add disabled, has padding, backgroundColor and placeholder props

bntsv:
Adding disabled, hasPadding, backgroundColor, and placeholder behaviour to the Thumbnail component

- [X] I have performed a self-review of my code.
- [X] I have added thorough tests where applicable (unit / component / visual).
- [X] If changes will affect consumers of the package, I have created a changeset entry.

---

## Testing
[How do I test my changes?](https://github.com/justeattakeaway/pie/wiki/PIE-Aperture)

| Task                   | Link                             |
|------------------------|----------------------------------|
| Aperture PR            | [🔗](https://github.com/justeattakeaway/pie-aperture/pull/239) |
| NextJS 14 deployment   | [🔗](https://pr239-aperture-nextjs-app-v14.pie.design/components/thumbnail) |
| Nuxt 3 deployment      | [🔗](https://pr239-aperture-nuxt-app.pie.design/components/thumbnail) |
| Vanilla deployment     | [🔗](https://pr239-aperture-vanilla-app.pie.design/components/thumbnail.html) |

### Reviewer 1 @raoufswe 


apps/pie-storybook/stories/pie-thumbnail.stories.ts
```diff
@@ -40,6 +44,34 @@ const thumbnailStoryMeta: ThumbnailStoryMeta = {
...
+            control: 'boolean',
+            defaultValue: {
+                summary: defaultArgs.hasPadding,
+            },
+        },
+        backgroundColor: {
+            description: 'Set the backgroundColor attribute of the thumbnail.',
+            control: 'text',
+            defaultValue: {
+                summary: defaultArgs.backgroundColor,
```
raoufswe, line 65: we use the `defaultProps` values from the `defs` file for the summary field here to show the default values of the components. otherwise, it won't be very clear to users thinking the default values of the story vs component. 

Screenshot 2025-01-22 at 12 29 43

apps/pie-storybook/stories/testing/pie-thumbnail.test.stories.ts
```diff
@@ -40,6 +44,34 @@ const thumbnailStoryMeta: ThumbnailStoryMeta = {
                 summary: defaultArgs.alt,
             },
         },
+        disabled: {
+            description: 'Set the disabled attribute of the thumbnail.',
+            control: 'boolean',
+            defaultValue: {
+                summary: defaultArgs.disabled,
```
raoufswe, line 51: this will need updating as mentioned above

apps/pie-storybook/stories/testing/pie-thumbnail.test.stories.ts
```diff
@@ -50,17 +82,29 @@ const Template: TemplateFunction<ThumbnailProps> = ({
...
     </pie-thumbnail>`;
 
 // Define the prop options for the matrix
 const sharedPropOptions = {
-    src: ['https://www.pie.design/assets/img/jet-logo-narrow.svg'],
-    alt: ['JET logo'],
+    src: [defaultArgs.src],
+    alt: [defaultArgs.alt],
+    disabled: [defaultArgs.disabled],
+    hasPadding: [defaultArgs.hasPadding],
```
raoufswe, line 105: we need to generate a visual snapshot for both when the component true and false so the values should be

 ```suggestion
    disabled: [true, false],
    hasPadding: [true, false],
```

File: packages/components/pie-thumbnail/src/thumbnail.scss [hunk 4]
```diff
@@ -11,15 +13,26 @@
   border-radius: var(--thumbnail-border-radius);
   border: 1px solid var(--thumbnail-border-color);
   background-color: var(--thumbnail-bg-color);
+  opacity: var(--thumbnail-opacity);
```
raoufswe, line 16: i think we don't need to set the opacity here and only set it when the component is disabled. 
as by default, the component shouldn't care about opacity until it's needed and browsers will apply the default values anyway. I also think we don't need a variable for it since it's only being used once so for simplicity we can do: 

 &.c-thumbnail--disabled {
    opacity: 50%
  }

packages/components/pie-thumbnail/src/thumbnail.scss
```diff
@@ -11,15 +13,26 @@
...
   .c-thumbnail-img {
+    box-sizing: border-box;
     width: 100%;
     height: 100%;
     object-fit: contain;
     display: block;
+    padding: var(--thumbnail-img-padding);
+
+    &.c-thumbnail--padding {
+      --thumbnail-img-padding: calc(var(--dt-spacing-a) / 2);
```
raoufswe, line 35: is there a reason why we didn't apply the padding in the container itself? 

something like 


```
  --thumbnail-padding: calc(var(--dt-spacing-a) / 2);
    &.c-thumbnail--padding {
    padding: var(--thumbnail-img-padding);
}

```

packages/components/pie-thumbnail/test/component/pie-thumbnail.spec.ts
```diff
@@ -1,7 +1,24 @@
-import { test, expect } from '@playwright/test';
+import { test, expect } from '@sand4rt/experimental-ct-web';
```
raoufswe, line 1: we are moving away from '@sand4rt/experimental-ct-web' so we should keep using '@playwright/test'; please.

packages/components/pie-thumbnail/src/thumbnail.scss
bntsv: If we apply padding to the container itself, the container will have a distance from other elements around it but there will be no padding between the outline/border and the image. I thought the idea is to have 2px between the borders of the element and the image. Besides, users can apply padding to the whole pie-thumbnail component externally is they want with as much px they like, why would we provide functionality that does that internally and limits the padding to 2 px?

packages/components/pie-thumbnail/test/component/pie-thumbnail.spec.ts
bntsv: when I try to use mount it doesnt allow me unless I am using `test` function from `@sand4rt/experimental-ct-web`. Should I also avoid using `mount` or is there a way around this? I could use `page` and access the component through it. I was only wondering how to pass desired props to it but I see in other examples `page.load(props)` so I guess this should work.

raoufswe: mount shouldn't be used anymore so we need to use page and access the component through it.

packages/components/pie-thumbnail/src/thumbnail.scss
bntsv: Sorry just realised what you mean, my bad, will adjust this!

packages/components/pie-thumbnail/src/thumbnail.scss
```diff
@@ -16,7 +17,16 @@
...
+  &.c-thumbnail--disabled {
+    opacity: 0.5;
+  }
+
+  &.c-thumbnail--padding {
+    padding: var(--thumbnail-padding);
+  }
+
   .c-thumbnail-img {
+    box-sizing: border-box;
```
raoufswe, line 29: is this needed?

packages/components/pie-thumbnail/test/component/pie-thumbnail.spec.ts
```diff
@@ -17,4 +33,74 @@ test.describe('PieThumbnail - Component tests', () => {
...
+                backgroundColor,
+            };
+            const pieThumbnailPage = new BasePage(page, 'thumbnail--default');
+            await pieThumbnailPage.load(props);
+
+            const thumbnailComponent = page.locator(componentSelector);
+
+            const [currentBgStyle, expectedBgStyle] = await getShadowElementStylePropValues(thumbnailComponent, componentSelector, ['background-color', backgroundColor as string]);
+
+            expect(currentBgStyle).toBe(expectedBgStyle);
```
siggerzz, line 49: Rather than using `getShadowElementStylePropValues` (which we've been trying to reduce usage of and eventually remove once our test refactor is complete), you may be able to leverage:

`await expect(thumbnailComponent).toHaveCSS('background-color, expectedBgStyle);`

packages/components/pie-thumbnail/test/component/pie-thumbnail.spec.ts
```diff
@@ -1,7 +1,23 @@
 import { test, expect } from '@playwright/test';
 import { BasePage } from '@justeattakeaway/pie-webc-testing/src/helpers/page-object/base-page.ts';
+import { getShadowElementStylePropValues } from '@justeattakeaway/pie-webc-testing';
+import { type ThumbnailProps } from '../../src/index.ts';
 
 const componentSelector = '[data-test-id="pie-thumbnail"]';
+const componentImgSelector = '[data-test-id="pie-thumbnail-img"]';
```
siggerzz, line 7: Can these be added to a `selectors.ts` file in this directory like the other components?

Having a single definition will help to improve reusability of selectors, as eventually, we'll look to expose these to our consumers 👍🏼 

https://github.com/justeattakeaway/pie/blob/main/packages/components/pie-card/test/helpers/page-object/selectors.ts

apps/pie-storybook/stories/pie-thumbnail.stories.ts
```diff
@@ -23,21 +27,49 @@ const thumbnailStoryMeta: ThumbnailStoryMeta = {
...
         alt: {
             description: 'Set the alt attribute for the underlying image tag.',
             control: 'text',
             defaultValue: {
-                summary: defaultArgs.alt,
+                summary: defaultProps.alt,
+            },
+        },
+        disabled: {
+            description: 'Set the disabled attribute of the thumbnail.',
```
raoufswe, line 48: images don't have a disabled attribute so i think we should say the following instead. 
```suggestion
            description: 'Applies the disabled styling of the thumbnail.',
```

apps/pie-storybook/stories/pie-thumbnail.stories.ts
```diff
@@ -23,21 +27,49 @@ const thumbnailStoryMeta: ThumbnailStoryMeta = {
...
+            control: 'boolean',
+            defaultValue: {
+                summary: defaultProps.disabled,
+            },
+        },
+        hasPadding: {
+            description: 'Set the hasPadding attribute of the thumbnail.',
```
raoufswe, line 55:
```suggestion
            description: 'Adds extra space around the thumbnail container. ',
```

apps/pie-storybook/stories/pie-thumbnail.stories.ts
```diff
@@ -23,21 +27,49 @@ const thumbnailStoryMeta: ThumbnailStoryMeta = {
...
+            control: 'boolean',
+            defaultValue: {
+                summary: defaultProps.hasPadding,
+            },
+        },
+        backgroundColor: {
+            description: 'Set the backgroundColor attribute of the thumbnail.',
```
raoufswe, line 62:
```suggestion
            description: 'Applies a background color to the thumbnail container.',
```

apps/pie-storybook/stories/pie-thumbnail.stories.ts
```diff
@@ -23,21 +27,49 @@ const thumbnailStoryMeta: ThumbnailStoryMeta = {
...
+            control: 'text',
+            defaultValue: {
+                summary: defaultProps.backgroundColor,
+            },
+        },
+        placeholder: {
+            description: 'Set the placeholder attribute of the thumbnail.',
```
raoufswe, line 69:
```suggestion
            description: 'If an image is unavailable, the placeholder  prop can be used to ensure there is always something visible to users.',
```

apps/pie-storybook/stories/testing/pie-thumbnail.test.stories.ts
```diff
@@ -23,21 +27,49 @@ const thumbnailStoryMeta: ThumbnailStoryMeta = {
...
         alt: {
             description: 'Set the alt attribute for the underlying image tag.',
             control: 'text',
             defaultValue: {
-                summary: defaultArgs.alt,
+                summary: defaultProps.alt,
+            },
+        },
+        disabled: {
+            description: 'Set the disabled attribute of the thumbnail.',
```
raoufswe, line 48: we need to update the description values here to match the above suggestions

packages/components/pie-thumbnail/test/component/pie-thumbnail.spec.ts
```diff
@@ -1,7 +1,23 @@
 import { test, expect } from '@playwright/test';
 import { BasePage } from '@justeattakeaway/pie-webc-testing/src/helpers/page-object/base-page.ts';
+import { getShadowElementStylePropValues } from '@justeattakeaway/pie-webc-testing';
+import { type ThumbnailProps } from '../../src/index.ts';
```
siggerzz, line 4: I think this is causing your test issues for some reason. Can you try and import this directly from `../../src/defs.ts`

packages/components/pie-thumbnail/src/index.ts
```diff
@@ -27,21 +28,48 @@ export class PieThumbnail extends LitElement implements ThumbnailProps {
...
 
         const wrapperClasses = {
             'c-thumbnail': true,
             [`c-thumbnail--${variant}`]: true,
+            'c-thumbnail--disabled': disabled,
+            'c-thumbnail--padding': hasPadding,
+        };
+
+        const wrapperStyles = {
+            backgroundColor,
```
ashleynolan, line 67: would be interesting to know why `backgroundColor` is so open, rather than just having a set number of background colors to pick from? Can see this getting misused potentially if any colour can be specified.

packages/components/pie-thumbnail/src/thumbnail.scss
```diff
@@ -3,6 +3,7 @@
   --thumbnail-border-radius: var(--dt-radius-rounded-b);
   --thumbnail-bg-color: var(--dt-color-container-default);
   --thumbnail-border-color: transparent;
+  --thumbnail-padding: calc(var(--dt-spacing-a) / 2);
```
ashleynolan, line 6: More feedback to design – should we have a token for `2px` if we're going to be using it in components as a spacing value?

cc @RLumleyJET

apps/pie-storybook/stories/testing/pie-thumbnail.test.stories.ts
```diff
@@ -23,21 +27,49 @@ const thumbnailStoryMeta: ThumbnailStoryMeta = {
...
+        },
+        backgroundColor: {
+            description: 'Applies a background color to the thumbnail container.',
+            control: 'text',
+            defaultValue: {
+                summary: defaultProps.backgroundColor,
+            },
+        },
+        placeholder: {
+            description: 'If an image is unavailable, the placeholder  prop can be used to ensure there is always something visible to users.',
```
ashleynolan, line 69:
```suggestion
            description: 'If an image is unavailable, the placeholder prop can be used to ensure there is always something visible to users.',
```

packages/components/pie-thumbnail/src/index.ts
raoufswe: the background color of the thumbnail component can be any background color as specified by design specs.

packages/components/pie-thumbnail/src/defs.ts
```diff
@@ -4,16 +4,32 @@ export const variants = [
...
-export type DefaultProps = ComponentDefaultProps<ThumbnailProps, 'variant' | 'src' | 'alt'>;
+export type DefaultProps = ComponentDefaultProps<ThumbnailProps, 'variant' | 'src' | 'alt' | 'disabled' | 'hasPadding' | 'backgroundColor' | 'placeholder'>;
 
 export const defaultProps: DefaultProps = {
     variant: 'default',
     src: '',
     alt: '',
+    disabled: false,
+    hasPadding: false,
+    backgroundColor: 'var(--dt-color-container-default)',
```
jamieomaguire, line 30: how do we ensure this name stays in date?

packages/components/pie-thumbnail/src/index.ts
jamieomaguire: agreed

packages/components/pie-thumbnail/src/defs.ts
raoufswe: I am not sure I am following. the design specs mentioned the background color of the component should be container-default by default and can be any color. if it ever changes, then they will probably add a changelog and we need to update it.

packages/components/pie-thumbnail/src/index.ts
raoufswe: could you please share how you think we should proceed with this? 
I don't think manually listing all background color tokens is maintainable.
this is the designs note

``
When using transparent PNGs the background colour is set to $container-default, and can be changed as need to any container colour token.
``

packages/components/pie-thumbnail/src/index.ts
```diff
@@ -27,21 +28,48 @@ export class PieThumbnail extends LitElement implements ThumbnailProps {
...
         };
 
         return html`
-            <div data-test-id="pie-thumbnail" class="${classMap(wrapperClasses)}">
-                <img data-test-id="pie-thumbnail-img" src="${src}" class="c-thumbnail-img" alt="${alt}">
+            <div data-test-id="pie-thumbnail" class="${classMap(wrapperClasses)}" style="${styleMap(wrapperStyles)}">
```
jamieomaguire, line 71: Sorry but I don't think we can allow this. It looks like a direct injection of raw CSS into the style tag. I think a user could put anything in here and that poses a security risk.

I think we should only let them select from a predefined list of options. Or something else more restricted.

packages/components/pie-thumbnail/src/defs.ts
bntsv: we can have a [fallback value](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties#custom_property_fallback_values) like var(--dt-color-container-default, inherit). If you make any changes to the design tokens names overall this would cause a lot of code to not work properly since we are already using these in many apps though.

packages/components/pie-thumbnail/src/index.ts
raoufswe: we updated to follow the variants approach as discussed offline.

packages/components/pie-thumbnail/src/defs.ts
raoufswe: we updated to follow the variants approach as discussed offline.

packages/components/pie-thumbnail/src/index.ts
raoufswe: we updated to follow the variants approach as discussed offline.

packages/components/pie-thumbnail/src/defs.ts
```diff
@@ -4,16 +4,36 @@ export const variants = [
     'default', 'outline',
 ] as const;
 
+export const backgroundColors = [
+    'default', 'subtle', 'strong', 'dark', 'inverse', 'inverse-alternative'
+];
```
raoufswe, line 9: we forgot to add as const here

packages/components/pie-thumbnail/src/thumbnail.scss
raoufswe: i will raise this with designers on Moday's standup

packages/components/pie-thumbnail/test/component/pie-thumbnail.spec.ts
```diff
@@ -1,20 +1,55 @@
...
-        // Act
-        const thumbnail = page.locator(componentSelector);
+        await basePage.load();
 
         // Assert
-        expect(thumbnail).toBeVisible();
+        await expect(page.getByTestId(thumbnail.selectors.container.dataTestId)).toBeVisible();
+    });
+
+    test('should set placeholder values if a placeholder is provided and img src is NOT valid', async ({ page }) => {
```
jamieomaguire, line 16: is this testing the onerror function? Can't quite tell

jamieomaguire: if it is, probably worth specifying the test is when an image load errors

bntsv: This is testing the public interface for setting a placeholder.  So it is also indirectly testing the private method that is used as a callback for the error event of the img tag.

bntsv: I have adjusted the title

packages/components/pie-thumbnail/src/defs.ts
```diff
@@ -4,16 +4,36 @@ export const variants = [
     'default', 'outline',
 ] as const;
 
+export const backgroundColors = [
+    'default', 'subtle', 'strong', 'dark', 'inverse', 'inverse-alternative'
+]as const;
```
ashleynolan, line 9: Super NIT 😄 

```suggestion
] as const;
```

packages/components/pie-thumbnail/src/thumbnail.scss
```diff
@@ -16,6 +17,34 @@
...
+
+  &.c-thumbnail--background-dark {
+    --thumbnail-bg-color: var(--dt-color-container-dark);
+  }
+
+  &.c-thumbnail--background-inverse {
+    --thumbnail-bg-color: var(--dt-color-container-inverse);
+  }
+
+  &.c-thumbnail--background-inverse-alternative {
```
ashleynolan, line 40: To fit with our CSS namign scheme:

```suggestion
  &.c-thumbnail--background-inverseAlt {
```

apps/pie-storybook/stories/testing/pie-thumbnail.test.stories.ts
```diff
@@ -50,28 +85,67 @@ const Template: TemplateFunction<ThumbnailProps> = ({
...
+})();
+
+export const InvalidPlaceholder = createStory<ThumbnailProps>(Template, {
+    placeholder: {
+        src: 'https://www.pie.design/assets/img/404_narrow.png',
+        alt: 'Placeholder Alt',
+    },
+    src: 'https://www.pie.design/assets/img/jet-logo-narrow.svg',
+    alt: 'JET Logo',
+})();
```
raoufswe, line 147: the storyname might be wrong. it mentions `InvalidPlaceholder` but all of the props are valid and all images will load as expected.

packages/components/pie-thumbnail/src/thumbnail.scss
ashleynolan, line 40: As dashes are used to indicate child elements, can just use camelCase for the background classes:

```suggestion
  &.c-thumbnail--backgroundSubtle {
    --thumbnail-bg-color: var(--dt-color-container-subtle);
  }

  &.c-thumbnail--backgroundStrong {
    --thumbnail-bg-color: var(--dt-color-container-strong);
  }

  &.c-thumbnail--backgroundDark {
    --thumbnail-bg-color: var(--dt-color-container-dark);
  }

  &.c-thumbnail--backgroundInverse {
    --thumbnail-bg-color: var(--dt-color-container-inverse);
  }

  &.c-thumbnail--backgroundInverseAlt {
```

ashleynolan (commented):
Couple of v small naming things, otherwise looks good!

packages/components/pie-thumbnail/src/index.ts
```diff
@@ -27,21 +29,46 @@ export class PieThumbnail extends LitElement implements ThumbnailProps {
...
+    @validPropertyValues(componentSelector, backgroundColors, defaultProps.backgroundColor)
+    public backgroundColor = defaultProps.backgroundColor;
+
+    @property({ type: Object })
+    public placeholder: ThumbnailProps['placeholder'];
+
+    private handleImageError = () => {
+        if (this.placeholder?.src) this.src = this.placeholder.src as string;
+        if (this.placeholder?.alt) this.alt = this.placeholder.alt as string;
+    };
```
raoufswe, line 48: the placeholder props don't seem to render as expected in aperture apps 👀 
https://pr239-aperture-nuxt-app.pie.design/components/thumbnail
https://pr239-aperture-vanilla-app.pie.design/components/thumbnail.html

might be the case that we need to swap the values using a state to ensure the component updates correctly?

packages/components/pie-thumbnail/src/thumbnail.scss
raoufswe: i think having them as `camelCase` will make it harder to construct the class name.
for example, the prop value is `inverse-alt` (all of our prop values are snake-case), to create the className `backgroundInverseAlt`, we need a transformer and I don't think we follow this approach in the code base.

raoufswe: how about ` &.c-thumbnail-background--subtle `

raoufswe: updated to use the last suggestion

raoufswe (approved):

jamieomaguire (approved):

siggerzz (approved):

packages/components/pie-thumbnail/src/thumbnail.scss
```diff
@@ -16,6 +17,34 @@
...
+
+  &.c-thumbnail-background--dark {
+    --thumbnail-bg-color: var(--dt-color-container-dark);
+  }
+
+  &.c-thumbnail-background--inverse {
+    --thumbnail-bg-color: var(--dt-color-container-inverse);
+  }
+
+  &.c-thumbnail-background--inverse-alternative {
```
ashleynolan, line 40: 🙃 
```suggestion
  &.c-thumbnail-background--inverseAlternative {
```

ashleynolan (commented):
One NIT naming thing – then LGTM

ashleynolan (approved):
Approving based on discussion – can you make a ticket that covers the camelCase classname issue?

packages/components/pie-thumbnail/src/index.ts
```diff
@@ -27,21 +29,46 @@ export class PieThumbnail extends LitElement implements ThumbnailProps {
...
+            disabled,
+            hasPadding,
+            backgroundColor,
+            _handleImageError,
         } = this;
 
         const wrapperClasses = {
             'c-thumbnail': true,
             [`c-thumbnail--${variant}`]: true,
+            [`c-thumbnail-background--${backgroundColor}`]: true,
```
jamieomaguire, line 64: quick thing - just create a map here of the prop value to the camelCase class name. Then everything works as Ash has asked and its hardly any extra work/code as far as I can tell

raoufswe: i will do that in a follow up PR shortly

PR #2251, opened 2025-02-28
feat(pie-tag): DSW-2550 pie tag icon only mode (v1)

fernandofranca:
### @justeattakeaway/pie-tag

[Added] - supporting code for icon only mode
[Updated] - fixed padding in icon only mode

### pie-storybook

[Added] - regular story for Tag icon only mode
[Added] - test stories for Tag icon only mode

### pie-docs

[Added] - page for Tag icon only mode
[Updated] - content for Tag page regarding icon only mode

### Reviewer 1
- [ ] I have reviewed the Docs site preview
- [ ] If new links have been added, I have tested them in the preview

### Reviewer 2
- [ ] I have reviewed the Docs site preview
- [ ] If new links have been added, I have tested them in the preview


dandel10n (approved):

packages/components/pie-tag/src/index.ts
```diff
@@ -45,10 +45,50 @@ export class PieTag extends LitElement implements TagProps {
...
 
+    private isIconOnly = false;
+
+    updated (changedProperties: PropertyValues<this>) {
+        if (changedProperties.has('size')) this.checkIfIsIconOnly();
+    }
+
+    private checkIfIsIconOnly () {
+        // The size must be large
+        const isLargeSize = this.size === 'large';
```
raoufswe, line 56: We already have a check on `renderIconSlot` where the icon slot is only shown for the large size. Do you think we still need to revalidate here?

packages/components/pie-tag/src/index.ts
```diff
@@ -45,10 +45,50 @@ export class PieTag extends LitElement implements TagProps {
...
+
+        // The default slot must be empty
+        const defaultSlotText = this.textContent?.trim();
+        const isDefaultSlotEmpty = defaultSlotText === '';
+
+        // The icon slot must have some content
+        const iconSlot = this.shadowRoot?.querySelector('slot[name="icon"]') as HTMLSlotElement | null;
```
raoufswe, line 63: Do you think we could be more declarative and use Lit's `queryAssignedNodes` decorator here?

We usually use these decorators to query the slotted children in the codebase.

https://lit.dev/docs/components/shadow-dom/#query-assigned-nodes

packages/components/pie-tag/src/index.ts
```diff
@@ -45,10 +45,50 @@ export class PieTag extends LitElement implements TagProps {
...
+
+        if (isLargeSize && isDefaultSlotEmpty && iconSlot) {
+            const iconSlotNodes = iconSlot.assignedNodes({ flatten: true });
+
+            // The icon slot content must be an icon
+            if (iconSlotNodes && iconSlotNodes.length === 1) {
+                const firstNode = (iconSlotNodes[0] as Element);
+                const isIcon = firstNode.tagName.startsWith('ICON-');
```
raoufswe, line 71: I believe we also allow users to pass a custom SVG as an icon for all components. That's why we include the `::slotted(svg)` styles for all components that accept a named icon slot. Do you think we should remove the validation as the slot name is already called `icon` or extend it to match SVG as well?

fernandofranca:
Closing the PR per deployment and test issues. Opened a new one here: https://github.com/justeattakeaway/pie/pull/2253

packages/components/pie-tag/src/index.ts
fernandofranca: It could test for the SVG tag. Do you predict any issue with testing for icon and svg?

packages/components/pie-tag/src/index.ts
fernandofranca: Even if its a bit redundant, it's not expensive or consuming, and having it aggregated in a single place helps to explain the requirements for "icon only". I don't see a problem with that.

raoufswe: Can we create a getter and reuse it in both places then?

something like: 

```
   private get _shouldShowIcon (): string {
    return this.size === 'large';
    }
```

PR #2316, opened 2025-04-07
feat(pie-docs): DSW-2914 add the select component code section

raoufswe:

- Added the code section of the select component

- [x] I have filled out the DS Review Tracker checklist (Moving PR to "Ready to Review")

### Reviewer 1 - @ashleynolan 
- [x] If new links have been added, I have tested them in the preview

### Reviewer 2
- [x] If new links have been added, I have tested them in the preview


ashleynolan (dismissed):

thejfreitas (dismissed):

apps/pie-docs/src/components/select-input/code/props.json
```diff
@@ -0,0 +1,110 @@
...
+    ],
+    [
+      "disabled",
+      {
+        "type": "code",
+        "item": [
+          "boolean"
+        ]
+      },
+      "Same as the HTML disabled attribute - indicates whether the select is disabled.",
```
fernandofranca, line 36: Should we omit `Same as the HTML disabled attribute` as it was done for the options props?

apps/pie-docs/src/components/select-input/code/optgroup-props.json
```diff
@@ -0,0 +1,75 @@
...
+  "rows": [
+    [
+      "tag",
+      {
+        "type": "code",
+        "item": [
+          "\"optgroup\""
+        ]
+      },
+      "Must be `optgroup` to identify this as an option group.",
```
fernandofranca, line 18: This is slightly different from what we have in the TSDocs.

In the `defs.ts` file it suggests that it can either `option` or `optgroup`.

Should we keep everything consistent or there is some benefit in having differences for props descriptions?

apps/pie-docs/src/components/select-input/code/code.md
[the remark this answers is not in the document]
raoufswe: The component is called 'select-input' on design, and that will break both the doc site navigation and its content. 

Let's check with design if it's possible to rename in a follow up ticket.

apps/pie-docs/src/components/select-input/code/optgroup-props.json
raoufswe: I believe it's the latter, since we explicitly have a tables prop for the `optgroup` tag, and this suggests that these props are only available when the tag is set to "optgroup".

ashleynolan (approved):

PR #2551, opened 2025-09-05
fix(pie-modal): DSW-3317 allow SRs to announce heading/aria label

jamieomaguire:
- Allow an aria labe to be passed directly to the dialog element
- Falls back to heading text if none provided
- Remove `aria-busy="false"` if `isLoading` is false

### DIFF
--- a/packages/components/pie-modal/src/defs.ts
+++ b/packages/components/pie-modal/src/defs.ts
@@ -10,6 +10,7 @@ type AriaProps = {
     close?: string;
     back?: string;
     loading?: string;
+    label?: string;
 };
 
 type ActionProps = {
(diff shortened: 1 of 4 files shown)


packages/components/pie-modal/src/defs.ts
```diff
@@ -161,7 +162,6 @@ export const defaultProps: DefaultProps = {
     isDismissible: false,
     isFooterPinned: true,
     isFullWidthBelowMid: false,
-    isLoading: false,
```
xander-marjoram, line 164: Just wondering why you wanted/needed to remove this, `false` seems like a sensible default value to me

jamieomaguire: because there's no reason to announce it is NOT loading if it never is expected to load. 

If it loads: announce its busy
Loading finishes: its not busy

Never was loading: why mention it at all?

JoshuaNg2332 (approved):

PR #2854, opened 2026-05-14
chore(pie-monorepo): DSW-123 add new Axe `best-practice` rule

siggerzz:


apps/pie-docs/test/accessibility/accessibility.spec.js
```diff
@@ -16,6 +16,7 @@ expectedRoutesJson.forEach((route) => {
         const results = await makeAxeBuilder()
             .include(`[data-test-id=${selector}]`)
             .exclude('iframe')
+            .disableRules(['landmark-unique', 'color-contrast-enhanced', 'heading-order', 'empty-heading', 'landmark-main-is-top-level', 'empty-table-header'])
```
jamieomaguire, line 20: could we add a comment just saying why - I imagine future us will be confused :D

jamieomaguire (approved):

PR #2890, opened 2026-05-28
chore(pie-monorepo): DSW-123 upgrade yarn to stop dependabot PR's failing

siggerzz:
Currently Dependabot PR's fail as it runs in it's own isolated runner using yarn v4, which currently we don't use. This PR will enable Dependabot PR's to run without issue, and also modernise and secure PIE, as a number of new features have been added to Yarn 4 as a result of the increasing number of npm supplychain attacks

- [ ] I have added thorough tests where applicable (unit / component / visual).
- [ ] I have reviewed visual test updates properly before approving.
- [ ] If a changeset file has been created, I have tested these changes in [PIE Aperture](https://github.com/justeattakeaway/pie-aperture/) using the `/test-aperture` command.
---


.yarnrc.yml
```diff
@@ -1 +1,10 @@
+approvedGitRepositories:
+  - "**"
+
+compressionLevel: mixed
+
+enableScripts: false
```
siggerzz, line 6: Prevents `postinstall` scripts of our dependencies from running.

.yarnrc.yml
```diff
@@ -1 +1,10 @@
...
+
 nodeLinker: node-modules
+
+npmMinimalAgeGate: 0
```
siggerzz, line 10: This was added automatically, but will enable us to prevent the installation of packages that have been newly published to npm

fernandofranca: I'd be happy if we increase that amount to something like a month.

fernandofranca (approved):

.yarnrc.yml
siggerzz: Agreed, however the wording suggests it could impact our changeset snapshots, so perhaps something we need to explore later.

>Minimum age of a package version according to the publish date on the npm registry to be considered for installation.

If a package version is newer than the minimal age gate, it will not be considered for installation. This can be used to reduce the likelihood of installing compromised packages, or to avoid relying on packages that could still be unpublished (e.g. the npm registry has specific rules for packages less than 3 days old).

PR #2906, opened 2026-06-04
Dsw 3992 light dom

jamieomaguire:
ignore for now please

### DIFF
--- a/packages/tools/pie-css/scss/mixins/components/_radio.scss
+++ b/packages/tools/pie-css/scss/mixins/components/_radio.scss
@@ -194,25 +194,24 @@
     }
 }
 
-// =============================================================================
-// SLOTTED VARIANTS
-//
-// When styling a slotted <input type="radio"> inside a shadow DOM host,
-// pseudo-classes (:checked, :disabled, :focus-visible, etc.) must go INSIDE
-// the ::slotted() functional notation, while pseudo-elements (::before, ::after)
-// must go AFTER it.
-//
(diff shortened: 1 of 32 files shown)


packages/tools/pie-css/scss/mixins/components/_radio.scss
```diff
@@ -193,3 +193,222 @@
...
+/// =============================================================================
+
+/// Slotted variant of radio-input-base.
+/// Applies base styles, pseudo-elements, and state styles to a slotted radio input.
+/// @param {String} $selector - The selector to place inside ::slotted() (e.g. 'input[type="radio"]')
+/// @example scss - Basic usage
+///   :host {
+///     @include radio-slotted-input-base('input[type="radio"]');
+///   }
+@mixin radio-slotted-input-base($selector) {
```
ashleynolan, line 223: This is a really nice pattern 😄 

Q – could this use a default selector if it were to be radio specific?

```scss
@mixin radio-slotted-input-base($selector: 'input[type="radio"]') {
```

and then called just with:

```scss
@include radio-slotted-input-base();
```

Or could even set it as an optional variable, so that all the mixins can make use of it, and it be overridden by the user if they needed to:

```scss
$pRadioSelector: 'input[type="radio"]' !default;

@mixin radio-slotted-input-base() {
	::slotted(#{$pRadioSelector}) {
        …
    }
}
```

jamieomaguire: yeah you totally could add a default value! both are quite smooth ideas

PR #2917, opened 2026-06-05
chore(pie-monorepo): DSW-4032 update deps + enable corepack

siggerzz:
- Updates a few dependencies that weren't covered by the last PR.
- Update CI to use the official mise github action
- Enable corepack to fix Dependabot PRs

This PR also fixes an issue where Dependabot PR's are unable to run as a result of our switch from Volta > Mise.

https://github.com/justeattakeaway/pie/actions/runs/26785797996/job/78961073411

This broke because we wanted a single source of truth for tool versions, and removed `packageManager` from the package.json.

This is problematic because Dependabot uses `packageManager` to determine which yarn version to use, and when omitted, it attempts to guess the version > picks the wrong version and errors due to `.yarnrc.yml` config options that aren't present in the version it resolved.

Unfortunately, Dependabot runs in its own isolated runners, so we have no control over the logic / way to override the version it resolves.

The fix is to enable `packageManager` again, but tell `mise` that we have corepack enabled in the repo, so a `mise install` will still correctly install `yarn`, it just means when we update yarn, it's done in package.json

- [ ] I have added thorough tests where applicable (unit / component / visual).
- [ ] If a changeset file has been created, I have tested these changes in [PIE Aperture](https://github.com/justeattakeaway/pie-aperture/) using the `/test-aperture` command.
- [ ] I have reviewed visual test updates properly before approving.
---

### DIFF
--- a/apps/pie-docs/package.json
+++ b/apps/pie-docs/package.json
@@ -55,11 +55,10 @@
     "cssnano": "7.1.7",
     "dree": "5.1.5",
     "eleventy-plugin-toc": "1.1.5",
-    "js-cookie": "2.2.1",
     "markdown-it": "14.1.1",
     "markdown-it-anchor": "9.2.0",
     "node-html-parser": "7.1.0",
-    "postcss": "8.5.12",
+    "postcss": "8.5.15",
     "postcss-cli": "11.0.1",
     "slugify": "1.6.9"
   },
--- a/apps/pie-storybook/package.json
+++ b/apps/pie-storybook/package.json
@@ -35,7 +35,7 @@
     "@types/dompurify": "3.2.0",
     "react": "18.3.1",
     "react-dom": "18.3.1",
-    "react-syntax-highlighter": "15.6.6",
+    "react-syntax-highlighter": "16.1.1",
     "remark-gfm": "4.0.1",
     "storybook": "10.3.5",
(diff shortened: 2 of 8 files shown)


apps/pie-docs/package.json
```diff
@@ -55,11 +55,10 @@
     "cssnano": "7.1.7",
     "dree": "5.1.5",
     "eleventy-plugin-toc": "1.1.5",
-    "js-cookie": "2.2.1",
```
siggerzz, line 58: This was a missing peer dep for something we no longer use in the repo

apps/pie-storybook/package.json
```diff
@@ -35,7 +35,7 @@
     "@types/dompurify": "3.2.0",
     "react": "18.3.1",
     "react-dom": "18.3.1",
-    "react-syntax-highlighter": "15.6.6",
+    "react-syntax-highlighter": "16.1.1",
```
siggerzz, line 38: No breaking changes that affect us

jamieomaguire (dismissed):

raoufswe (dismissed):

JoshuaNg2332 (dismissed):

xander-marjoram (approved):

raoufswe (approved):

PR #2963, opened 2026-06-23
feat(pie-switch): DSW-4019 bind the checked prop correctly and add missing form behaviours

raoufswe:
- [Added] - Wrapped the checked value with the [lit live directive ](https://lit.dev/docs/api/directives/#live) to ensure the DOM and component state stay in sync.
- [Added] - Missing `defaultChecked` prop and form rest logic
- [Changed] - Refactored the toggle between leading and trailing labels to be handled via CSS.
- [Changed] - Updated the component to use a `<div>` as the parent wrapper when the label prop isn’t provided. This avoids nested `<label>` elements when consumers wrap the component with [an external label](https://webc.pie.design/?path=/docs/components-switch--overview#external-labels), which is an [invalid](https://forum.freecodecamp.org/t/nesting-label/462466) HTML pattern

- [-] I have reviewed visual test updates properly before approving.

---

### Reviewer 1

### Reviewer 2 @maledr5 


.changeset/small-mugs-mix.md
```diff
@@ -0,0 +1,9 @@
+---
+"@justeattakeaway/pie-switch": minor
+---
+
+- [Removed] - (Breaking Change) `aria.describedby` support from pie-switch. It was misused as `aria-describedby` expects element IDs to reference existing descriptive elements, which isn’t feasible across Shadow DOM boundaries and being text conflicted with aria-label, internal labels, and external label handling.
```
raoufswe: I am not sure if this deserves to be a major change. Mainly because the prop isn't used anywhere  from a quick GitHub search in our org. Also, if it's used, it isn't doing what it's supposed to do and will be ignored with change as the property is part of the `aria` object prop and nothing should be really break

raoufswe: reverted the change

packages/components/pie-switch/test/component/pie-switch.spec.ts
```diff
@@ -386,10 +386,54 @@ test.describe('Component: `Pie switch`', () => {
...
+            const switchEl = page.locator('#pie-switch');
+            await expect(switchEl).not.toHaveAttribute('checked');
+
+            // Act
+            await page.$eval('#testForm', (form) => {
+                (form as HTMLFormElement).reset();
+            });
+
+            // Assert
+            await expect(switchEl).toHaveAttribute('checked', '');
```
siggerzz, line 413: Think this should work given the other examples in this spec

```suggestion
            await expect(switchEl).not.toHaveAttribute('checked');
```

siggerzz (commented):
small comment

packages/components/pie-switch/test/component/pie-switch.spec.ts
raoufswe: updated!

raoufswe: I reverted the changes again because the test expects the switch to be checked. We’re using `toHaveAttribute('checked', '')` since it’s a boolean attribute, so its actual value is an empty string but it should still be there

siggerzz (approved):

packages/components/pie-switch/src/index.ts
```diff
@@ -166,6 +182,23 @@ export class PieSwitch extends FormControlMixin(DelegatesFocusMixin(PieElement))
...
+        if (this.checked === this.defaultChecked) {
+            return;
+        }
+
+        this.checked = this.defaultChecked;
+
+        const changeEvent = new Event('change', { bubbles: true, composed: true });
+        this.dispatchEvent(changeEvent);
+
+        this.handleFormAssociation();
```
maledr5, line 199: Do you know if `this.checked = this.defaultChecked;` would trigger the `updated` lifecycle method? because that is already calling `this.handleFormAssociation();` so it might be called twice. 

I see it is used like this in the `handleChange` method as well, so I don't think is a blocker, but something to maybe consider.

maledr5 (approved):

packages/components/pie-switch/src/index.ts
raoufswe: it does seem to be called 👀 but it's currently a pattern in multiple components already so i will address it in a follow up PR

PR #3041, opened 2026-07-20
feat(pie-monorepo): DSW-4184 Add aperture-props-test skill 

KatarinaNeskovic:
## Describe your changes

Adds the `aperture-props-test` agent skill, which streamlines manual verification of a newly added or changed PIE component prop in Aperture. This skill is user-invoked, as opposed to model-invoked so explicit call is required for agent to load it. Ticket is [here](https://justeattakeaway.atlassian.net/browse/DSW-4184?atlOrigin=eyJpIjoiMDAzMzk0MzlhMmVkNGZkNmI3OWJlZmIzOThjMjdiMDAiLCJwIjoiaiJ9). 

After `/test-aperture` has run on a PIE PR, the skill:
- pulls the matching Aperture branch (by branch name),
- temporarily adds a prop example across all 4 framework apps (Nuxt, Next.js v14/v15, Vanilla),
- opens Aperture in a separate IDE window for review, then starts all dev servers once user confirms the edits,
- prompts visual verification and, on confirmation, reverts the scratch edits and stops the servers.

Note: The skill assumes we don't commit props for every props test as per team's usual practice. 

Passes the repo `skill-reviewer` checklist and was tested end-to-end against `pie-textarea`'s `aria` and `size`prop in both Claude and OpenCode.

## Acceptance criteria

- A skill exists with clear trigger guidance for testing a newly added PIE Web Component prop in Aperture.
- The skill accepts the component name, prop name, and prop value as inputs. If prop value isn’t provided it suggest the value and confirms with the user before continuing
- It supports adding temporary test usage in all four Aperture applications.
- The skill starts the development server for each Aperture application concurrently.
- Once the servers are available, the skill prompts the user to verify the visual output in each application.
- The skill waits for explicit user confirmation before removing temporary test changes.
- After confirmation, the skill removes only the temporary changes it created and reports the files are cleaned up and servers stopped running.
- It does not commit, push, or create pull requests.
- It handles an unsupported component or missing example location by stopping and asking for user input rather than guessing. (if some component pages are missing for specific apps it will report that and continue with the workflow)
- The skill is tested against one representative component and prop (tested against `aria` and `size`). 

- [-] I have added thorough tests where applicable (unit / component / visual). — N/A: markdown-only agent skill, no runtime code.
- [-] I have reviewed the `PIE Storybook`/`PIE Docs` PR preview. — N/A: no Storybook/Docs changes.
- [-] I have reviewed visual test updates properly before approving. — N/A: no visual tests affected.
- [-] If changes will affect consumers of the package, I have created a changeset entry. — N/A: repo tooling only, not a published package change.
- [-] If a changeset file has been created, I have tested these changes in PIE Aperture using the `/test-aperture` command. — N/A: no changeset; not a component change.


jamieomaguire:
@KatarinaNeskovic I am not fully understanding the purpose of this skill. Perhaps we could chat offline about what problems we're trying to solve here?

.agents/skills/aperture-props-test/SKILL.md
```diff
@@ -0,0 +1,161 @@
+---
+name: aperture-props-test
+description: Verify a newly added or changed PIE component prop in Aperture across all 4 framework apps. Invoke as `/aperture-props-test <component> <prop> [value]` after `/test-aperture` has run on the PIE PR.
```
fernandofranca: "Verify a newly added or changed PIE component prop" seems a bit abstract.

I think the description can be more straightforward about when this skill is needed.

Something like "Assists the user in testing added or changed component props with the PIE Aperture repo".

"Users will invoke as `/aperture-props-test <component> <prop> [value]`".

Knowing if the `/test-aperture`  has already might require more context that probably only the user knows.

A good thing to have in mind is that the description is the first thing the AI models will look into for deciding if the skill instructions should be used or not, so a short and concise description with the least possible amount of information to be inferred is better.

.agents/skills/README.md
```diff
@@ -35,3 +35,13 @@ Reviews a `SKILL.md` file against required frontmatter rules and best-practice g
...
+
+---
+
+### aperture-props-test
+
+Verifies a newly added or changed PIE component prop in Aperture. After `/test-aperture` has run on the PIE PR, it pulls the matching Aperture branch, temporarily adds prop examples across all 4 framework apps (Nuxt, Next.js v14, Next.js v15, Vanilla), starts the dev servers, prompts visual verification, then reverts the scratch edits on confirmation. Nothing is committed.
+
+```
+npx skills add git@github.com:justeattakeaway/pie.git --skill aperture-props-test
+```
```
fernandofranca, line 47: Do you think users might want to use and start this skill in the Aperture repo?

.agents/skills/aperture-props-test/SKILL.md
KatarinaNeskovic: Hey @fernandofranca Thank you for your feedback. I will make the description more straightforward as you suggested!

.agents/skills/README.md
KatarinaNeskovic: Hey @fernandofranca I thought about that after our chat :)  I think it feels like a natural trigger if the user is already in their PIE branch, added props and triggered /test-aperture in the PIE PR - then the skill just redirects automatically to Aperture, takes relevant info from PIE and adds props. Happy to demo how it flows now and chat if you think it'd better flow that it runs from Aperture? :)

.agents/skills/aperture-props-test/SKILL.md
KatarinaNeskovic: @fernandofranca Just a small follow-up: this skill will be user invoked , so the model will not auto-select the skill based on the description but it will be loaded on user trigger. I made it more straightforward as you suggested, and it will surely help the agent if we change the model invocation type! Additionally, as the purpose of the description is at the moment for user to read it (and sometimes we all forget to run `/test-aperture`) I'd recommend having that reminder in the description. Would you agree? :)

fernandofranca: The reason for the short and concise description is that it will be read every time an llm needs to evaluate how to approach what the user is asking. It's encouraged not to have anything that is not explicitly needed for the AIs in that field.

> Run after `/test-aperture` has been triggered on the PIE PR.

A good place for adding the context for us humans would be the `.agents/skills/README.md`, as we describe each of the repo skills in that file.

.agents/skills/README.md
fernandofranca: Makes sense. No need to change this! 👍

.agents/skills/aperture-props-test/SKILL.md
fernandofranca: https://platform.claude.com/docs/en/agents-and-tools/agent-skills/best-practices#writing-effective-descriptions

> The description is critical for skill selection: Claude uses it to choose the right Skill from potentially 100+ available Skills. Your description must provide enough detail for Claude to know when to select this Skill, while the rest of SKILL.md provides the implementation details.

fernandofranca: Found this content in a PDF file :)

image

fernandofranca (approved):

PR #3071, opened 2026-07-29
fix(pie-radio-group): DSW-4118 fix render issues of radio label

maledr5:
## Describe your changes
This fixes rendering issues of the radio-group in inline mode, in some cases it will render all in one line, instead of the label expanding the whole space.

Previously, the `<legend>` was only rendered when `_hasLabel` was true, which caused a layout issue: when a label was added dynamically or the slot was populated after initial render, the legend would not appear in the correct position within the flex layout. In inline mode this was especially noticeable as the legend needs `flex-basis: 100%` to span the full row. It was discovered that most of the times `<legend>` was never loading, and only the `<slot>` was present. 

This solution simplifies the component by removing the `_hasLabel` state, the `_handleLabelSlotChange` handler, and the conditional render logic. The trade off is that the `<legend>` element will always render, but when empty, screenreaders ignore it. 

This can be tested in this Aperture PR: https://github.com/justeattakeaway/pie-aperture/pull/540

**Expected:**

01-vanilla

**Issue:**

02-nuxt

### Reviewer 1


packages/components/pie-radio-group/src/radio-group.scss
```diff
@@ -16,14 +16,24 @@
     flex-flow: column wrap;
 
     > legend {
-        margin-block-end: var(--dt-spacing-b);
+        margin: 0;
+        padding: 0;
+
+        ::slotted(*) {
```
jamieomaguire: why `*`? is that safe and does it need to be that general?

packages/components/pie-radio-group/src/index.ts
```diff
@@ -462,7 +437,7 @@ export class PieRadioGroup extends FormControlMixin(RtlMixin(PieElement)) implem
                 data-test-id="pie-radio-group-fieldset"
                 aria-describedby=${hasAssistiveText ? assistiveTextId : nothing}
                 class="${classMap(classes)}">
-                    ${this._renderWrappedLabel()}
+                    <legend><slot name='label'></slot></legend>
```
jamieomaguire, line 440: I see a whitespace change in your before/after screenshot between label and radios. Is that expected on the vertical aligned version?

packages/components/pie-radio-group/src/index.ts
```diff
@@ -148,16 +144,6 @@ export class PieRadioGroup extends FormControlMixin(RtlMixin(PieElement)) implem
         this.dispatchEvent(changedEvent);
     }
 
-    /**
-     * Updates the `_hasLabel` state when content is added to the label slot.
-     * @param {Event} e - The slotchange event.
-     * @private
-     */
-    private _handleLabelSlotChange (e: { target: HTMLSlotElement }): void {
```
jamieomaguire, line 156: I think leveraging `updated` or similar might be the fix here instead of always rendering a `legend` regardless of labels. I have certainly seen problems like this arise from relying on `slotChange` soley as I think in frameworks this isn't a reliable lifecycle to hook into on initial render. So this would remain but potentially called in `updated` too (or whatever lifecycle method is most appropriate).

What do you think?

packages/components/pie-radio-group/src/radio-group.scss
maledr5: good point!

packages/components/pie-radio-group/src/index.ts
maledr5: Regarding the extra space, it's meant to be there, it was the old style added to the `legend` element, is that that element was never rendering before: 
`> legend {
        margin-block-end: var(--dt-spacing-b);
    }`
    
 now that space was moved to the slotted content, I believe doing the same thing intended initially.

packages/components/pie-radio-group/src/index.ts
maledr5: This was the issue indeed. As the label is actually rendering in SSR (in Nuxt at least), no lifecycle method gets called, including `slotchange`, because nothing is changing. The `hasLabel` in Nuxt has always been false. 

I took another look to see if other lifecycle methods could pick this up but is not working. I have asked 2 different AIs and both agree with the same solution: As SSR can't know what the slot has, we need to use CSS to always render the legend but hide it. 

Can you clarify why is a problem to always render the `legend`? From what I researched and tested with voice over, if it is empty screen readers ignore it. Regarding the extra space, I explained that in the other comment, but it is, or was intended at least originally.

jamieomaguire (approved):

packages/components/pie-radio-group/src/index.ts
jamieomaguire: Sorry @maledr5 I completely misread some of what the code is doing. Your changes make perfect sense. No issue with the legend being rendered as we can't query during SSR like you said 👍

PR #3089, opened 2026-08-04
feat(pie-icon-with-background): DSW-4171 add isDimmed state

nindyahapsari:

[Added] - `isDimmed` prop for `pie-icon-with-background`

isDimmed is a disable state that apply only visually not functionally because the icon itself is not interactive

---

### Reviewer 1
- [-] I have reviewed the Aperture changes (if added)

### Reviewer 2
- [-] I have reviewed the Aperture changes (if added)


packages/components/pie-icon-with-background/src/defs.ts
```diff
@@ -36,6 +36,11 @@ export interface IconWithBackgroundProps {
      * When true, applies a stronger colour emphasis. Has no effect on the `neutral-alternative` variant.
      */
     isStrong?: boolean;
+
+    /**
+     * When true, applies disabled colours to the component.
+     */
+    isDisabled?: boolean;
```
jamieomaguire: I know this is a draft but I think we generally just name it `disabled`. Breaks from our boolean is/has convention but I believe thats our pattern

nindyahapsari: Good catch. I'll change this

raoufswe (dismissed):

packages/components/pie-icon-with-background/README.md
```diff
@@ -37,6 +37,7 @@ Ideally, you should install the component using the **`@justeattakeaway/pie-webc
 | `size`     | `small`, `medium`, `large`, `xlarge`                                                                                                              | The size of the component, sizing both the container and slotted icon.  | `medium`  |
 | `variant`  | `neutral`, `neutral-alternative`, `information`, `success`, `error`, `warning`, `brand-02`, `brand-03`, `brand-04`, `brand-05`, `brand-06`, `brand-08` | The background colour variant of the component.                         | `neutral` |
 | `isStrong` | `true`, `false`                                                                                                                                   | When true, applies a stronger colour emphasis (no effect on `neutral-alternative`). | `false`   |
+| `isDimmed` | `true`, `false`                                                                                                                                   | When true, applies a dimmed styling to the component.                               | `false`   |
```
jamieomaguire: I haven't checked what we say in Tag, but there is always room for improvement! I would extend this description to explain its for styling it a bit like it is disabled. It is especially useful if the component is used inside another component that _can_ be disabled. This component is not interactive and therefore it is not named disabled to avoid misunderstanding.

You could add some of this to the prop description, and potentially a small section to the readme specifically to mention how to use as part of a more complex component that can be disabled.

I would consider if any of this could be re-used in the pie-tag readme too (and any other components that use this prop too). That would only warrant a patch version update in my opinion so the new readme is published for consumers.

Let me know what you think.

nindyahapsari: Thanks Jamie! I agree on the explanation because the name _isDimmed_ perhaps doesn't automatically explain that, it is more on styling/visual rather than functionality. 

I checked on pie-tag and the explanation for _isDimmed_ is the same (one-liner explanation). I could also update that in separate PR so that changes doesn't get mix up with this one.

Let me know if that is a good idea.

ashleynolan (dismissed):

packages/components/pie-icon-with-background/test/component/pie-icon-with-background.spec.ts
```diff
@@ -85,4 +85,32 @@ test.describe('PieIconWithBackground - Component tests', () => {
...
+            const basePage = new BasePage(page, 'icon-with-background--default');
+
+            await basePage.load({ isDimmed: false });
+
+            // Act
+            const iconWithBackground = page.locator(componentSelector);
+
+            // Assert
+            await expect(iconWithBackground).not.toHaveClass(/is-dimmed/);
+        });
```
raoufswe: I think we don't need this test as it's already covered by the visual / Percy test. What matters the most is the component renders in a correct way regardless of its implementation details.

Also, the test will still pass if we accidentally renamed or removed the class because it never existed still so it's better to rely on visual tests for visual things and leave component tests for things like interaction or logic.

nindyahapsari: good catch! thank you! 
removed.

raoufswe (dismissed):

packages/components/pie-icon-with-background/README.md
jamieomaguire: I am happy for readme updates to go into this PR

packages/components/pie-icon-with-background/README.md
```diff
@@ -37,6 +37,7 @@ Ideally, you should install the component using the **`@justeattakeaway/pie-webc
...
 | `variant`  | `neutral`, `neutral-alternative`, `information`, `success`, `error`, `warning`, `brand-02`, `brand-03`, `brand-04`, `brand-05`, `brand-06`, `brand-08` | The background colour variant of the component.                         | `neutral` |
 | `isStrong` | `true`, `false`                                                                                                                                   | When true, applies a stronger colour emphasis (no effect on `neutral-alternative`). | `false`   |
+| `isDimmed` | `true`, `false`                                                                                                                                   | When true, applies a dimmed visual styling to indicate a disabled context. Useful when this component is placed inside a parent component that has a disabled state (e.g. a disabled card or form section). Named `isDimmed` rather than `disabled` as this component is non-interactive. | `false`   |
```
jamieomaguire, line 40: please add to tag readme too

packages/components/pie-icon-with-background/test/component/pie-icon-with-background.spec.ts
```diff
@@ -85,4 +85,19 @@ test.describe('PieIconWithBackground - Component tests', () => {
             await expect(iconWithBackground).toHaveClass(/c-iconWithBackground--strong/);
         });
     });
+
+    test.describe('isDimmed prop', () => {
+        test('should apply the is-dimmed class when isDimmed is true', async ({ page }) => {
```
jamieomaguire: How come you added this test when there is a visual test as well?

packages/components/pie-icon-with-background/test/component/pie-icon-with-background.spec.ts
jamieomaguire: I know it isnt in this PR but same question as below for all the classes. Why are you testing these?

jamieomaguire (changes requested):

packages/components/pie-icon-with-background/test/component/pie-icon-with-background.spec.ts
nindyahapsari: Hey, so I dont have a specific reason. I just thought it was a good use case to add another layer of test (not just visual test). But i understand it's unnecessary (from the feedback)

packages/components/pie-icon-with-background/test/component/pie-icon-with-background.spec.ts
nindyahapsari: The answer in the previous comment. Since it's not necessary, i will remove these test

packages/components/pie-icon-with-background/test/component/pie-icon-with-background.spec.ts
nindyahapsari: since it's unnecessary i will remove the test

packages/components/pie-icon-with-background/README.md
nindyahapsari: Yes, I will do this one

nindyahapsari: done. push the changes to this PR

apps/pie-storybook/stories/pie-icon-with-background.stories.ts
```diff
@@ -55,6 +55,13 @@ const iconWithBackgroundStoryMeta: IconWithBackgroundStoryMeta = {
                 summary: String(defaultProps.isStrong),
             },
         },
+        isDimmed: {
+            description: 'When true, applies a purely visual dimmed appearance. Use this when the component is placed inside an interactive parent component that has a disabled state, such as a disabled card or button.',
+            control: 'boolean',
+            defaultValue: {
+                summary: String(defaultProps.isDimmed),
```
raoufswe: do we need to cast it to string? just checked other boolean values in the codebase and it seems it might not be needed

nindyahapsari: Oh great catch. I miss this one.

raoufswe (approved):

fernandofranca (approved):

jamieomaguire (approved):

PR #3156, opened 2026-08-27
feat(pie-toast): DSW-3673 toast stacking vertically

nindyahapsari:

Multiple toast only available when `isStack` is `true`, otherwise the default behaviour is one toast at a time

[Added] - support for displaying up to 3 toasts simultaneously in a stacked layout

Preview:

(updated demo)

https://github.com/user-attachments/assets/786e13d8-7561-4091-863a-3936e9b947bc

 
---

### Reviewer 1
- [ ] I have reviewed the Aperture changes (if added)

### Reviewer 2
- [ ] I have reviewed the `PIE Storybook`/`PIE Docs` PR preview.
- [ ] I have verified that all acceptance criteria for this ticket have been completed.
- [ ] I have reviewed the Aperture changes (if added)
- [ ] If there are visual test updates, I have reviewed them.


raoufswe:
i think the unmounting of expired toasts is a bit off. When three toasts are shown with different expiring duration (wait a second before adding a new toast to the queue), the expired toast should gracefully unmounts, and the next toast should take its place and go down. Currently they all just disappear and get replaced with one.

https://github.com/user-attachments/assets/10413ea6-35d6-4563-8c7a-9ef1dd0bb700

I also think after the first render, the stacking is maxed to two only not three, and then only one is shown each time. 

https://github.com/user-attachments/assets/ea5c1dc8-1618-430f-9820-38bafc1f0762



do you also think if we should keep the current max to 1, then users can opt in if they want to show the stacking?

nindyahapsari:
> i think the unmounting of expired toasts is a bit off. When three toasts are shown with different expiring duration (wait a second before adding a new toast to the queue), the expired toast should gracefully unmounts, and the next toast should take its place and go down. Currently they all just disappear and get replaced with one.
> 
>  Screen.Recording.2026-08-31.at.13.54.12.mov 
> I also think after the first render, the stacking is maxed to two only not three, and then only one is shown each time.
> 
>  Screen.Recording.2026-08-31.at.13.57.37.mov 
> do you also think if we should keep the current max to 1, then users can opt in if they want to show the stacking?

@raoufswe Hey hey! thanks for the reviews. I took some times to figure out about the animations and i think now is better in terms of the flow (the slide out animation) not just vanishing in thin air. 

other than that, what I found when doing the animation is that the toast can be a lot when you displaying 3 at the same time. It could be confusing. 
I would agree and learn towards your input of making the 3 stack toast an opt.

packages/components/pie-toast-provider/test/component/pie-toast-provider.spec.ts
```diff
@@ -166,21 +154,22 @@ test.describe('PieToastProvider - Component tests', () => {
...
+                const visibleToasts = await page.evaluate(() => {
+                    const tp = document.querySelector('pie-toast-provider') as unknown as { _visibleToasts: ExtendedToastProps[] };
+                    return tp._visibleToasts;
+                });
+
+                expect(visibleToasts.length).toBeGreaterThan(0);
+                visibleToasts.forEach((toast) => {
                     expect(toast.isDismissible).toBeTruthy();
                     expect(toast.variant).toBe('neutral');
                 });
```
jamieomaguire: are you certain this is synchronus?

packages/components/pie-toast-provider/test/component/pie-toast-provider.spec.ts
```diff
@@ -166,21 +154,22 @@ test.describe('PieToastProvider - Component tests', () => {
...
-                // Assert — iterate every toast that passed through the queue.
-                // Note: the first toast is immediately moved to _currentToast by the component
-                // and does not appear in the queue snapshots.
-                const seenToasts = (await getQueueSnapshots(page)).flat();
-                expect(seenToasts.length).toBeGreaterThan(0);
-                seenToasts.forEach((toast) => {
+                });
+
+                // Assert — read visible toasts directly; global options must be applied to each
+                const visibleToasts = await page.evaluate(() => {
```
jamieomaguire: could this be reusable?

nindyahapsari: Good catch. Created a helper function for this

packages/components/pie-toast-provider/test/component/pie-toast-provider.spec.ts
nindyahapsari: you are right. The state update is sync but the render isn't. The helper now awaits the updateComplete

packages/components/pie-toast-provider/src/defs.ts
```diff
@@ -53,13 +53,20 @@ export interface ToastProviderProps {
      * When set to `default`, the toasts will be positioned at bottom-left for RTL languages and bottom-right for LTR languages.
      */
     position?: typeof positions[number];
+
+    /**
+     * When true, up to `MAX_VISIBLE_TOASTS` toasts are displayed at the same time in a vertical
```
raoufswe, line 58: might be clearer to mention the value of MAX_VISIBLE_TOASTS

packages/components/pie-toast-provider/src/index.ts
```diff
@@ -31,6 +34,20 @@ export { toaster } from './toaster';
...
+ *
+ * Dismissal runs in two phases, each driven by a motion event: the toast slides out
+ * (`animationend`), then its row collapses (`transitionend`). Neither event fires when motion is
+ * disabled — a consumer setting `animation: none` or `transition: none`, or a hidden element.
+ * Without a fallback the toast would stay mounted forever and no queued toast would ever be
+ * promoted into its slot.
+ *
+ * Comfortably longer than the 200ms slide-out and 200ms collapse, so it never pre-empts real
+ * motion.
+ */
```
raoufswe, line 48: i might be missing something, but this feels like implementation details that doesn't bring much benefits as a comment and not sure what we are trying to document here?

packages/components/pie-toast-provider/src/index.ts
```diff
@@ -75,106 +142,179 @@ export class PieToastProvider extends PieElement implements ToastProviderProps {
...
+                ${renderAnnouncements(politeAnnouncements)}
+            </div>
+            <div
+                class="c-toast-provider-announcer"
+                role="alert"
+                aria-live="assertive"
+                aria-atomic="false"
+                data-test-id="pie-toast-provider-announcer-assertive">
+                ${renderAnnouncements(assertiveAnnouncements)}
             </div>
```
raoufswe, line 295: shouldn't aria-atomi  be set to true in live regions?

apps/pie-storybook/stories/testing/pie-toast-provider.test.stories.ts
```diff
@@ -253,3 +262,23 @@ const AutoResolveProviderTemplate = () => html`
...
+    ...createStory<ToastProviderProps>(PositionTemplate, { ...defaultArgs, isStacked: true })(),
+    play: () => {
+        toaster.create({ message: 'First toast message', duration: null });
+        toaster.create({ message: 'Second toast message', duration: null });
+        toaster.create({ message: 'Third toast message', duration: null });
+    },
+};
+
+// Same three toasts as `Stacked`, but with the default `isStacked: false`, so only the first
+// is displayed and the other two wait in the queue.
```
raoufswe, line 276: i think the comment can be omitted as the Unstacked const name is clear enough to document that

packages/components/pie-toast-provider/src/toast-provider.scss
```diff
@@ -43,11 +45,39 @@
...
     }
 
+    .c-toast-provider-item {
+        display: grid;
+        grid-template-rows: 1fr;
+
+        // The inter-toast spacing lives on the item rather than as a `gap` on the container so it
+        // can collapse alongside the toast height. A `gap` cannot be animated per item, so it would
+        // stay at full size throughout the collapse and then disappear when the item is removed,
+        // making the surviving toasts jump.
```
raoufswe, line 55: This looks more of an implementation details and perhaps a decision log so i would remove it.

packages/components/pie-toast-provider/src/index.ts
```diff
@@ -75,106 +142,179 @@ export class PieToastProvider extends PieElement implements ToastProviderProps {
...
-                    @pie-toast-leading-action-click="${_currentToast.onPieToastLeadingActionClick}">
-                </pie-toast>
-            `}
+            ${repeat(_visibleToasts, (toast) => toast, (toast) => html`
+                <div
+                    class="c-toast-provider-item${this._collapsingToasts.includes(toast) ? ' c-toast-provider-item--collapsing' : ''}"
+                    @transitionend="${() => { if (this._collapsingToasts.includes(toast)) this._finalizeCollapse(toast); }}">
+                    <pie-toast
+                        class="${this._getToastClass(toast)}"
+                        .isOpen="${live(true)}"
```
raoufswe, line 302: mind sharing why we need live(true) here?
same goes for the (.) binding, maybe it could work by simply passing 


```suggestion
                        isOpen
```

packages/components/pie-toast-provider/src/index.ts
```diff
@@ -75,106 +142,179 @@ export class PieToastProvider extends PieElement implements ToastProviderProps {
...
-     * Clears all toasts from the queue and dismisses the currently visible toast.
+     * Clears all toasts from the queue and dismisses all visible toasts.
      */
     public clearToasts () {
+        this._clearMotionFallbacks();
         this._toasts = [];
-        this._currentToast = null;
+        this._visibleToasts = [];
+        this._dismissingToasts = [];
+        this._collapsingToasts = [];
```
raoufswe, line 245: do you think we could have use the same array but added a status property in the toast to figure out if it's dismissing | visible | collapsing?

packages/components/pie-toast-provider/src/index.ts
nindyahapsari: Good question. We need live(true) because when a user closes the toast, pie-toast sets its own isOpen to false immediately. Without live(), Lit's _dirty check_ sees the bound value is still true (in Lit's memory) and skips writing it back, leaving the toast's shadow content empty while we're still animating out. live() forces Lit to read the actual DOM property, see it's false, and set it back to true so the content stays visible for the 400ms slide-out animation. Without it, the animation runs on an invisible element.

We can't use just isOpen without live() because then the animation would be invisible. The alternatives (stagger renders, use a separate dismissing prop, or prevent the toast from closing itself) are all more complex and don't reduce bugs, they just move the complexity around.

But I'm open for suggestions :)

packages/components/pie-toast-provider/src/index.ts
nindyahapsari: Good question, aria-atomic="false" is intentional here. With "true" the screen reader reads the entire region as one atomic block, so if 3 toasts become visible at once all 3 messages get merged into a single announcement. With "false" each new item appended to the region is announced individually which is what we need for the stacking case. The component tests verify this in the all stacked toasts announced when multiple become visible in one render test.

packages/components/pie-toast-provider/src/index.ts
nindyahapsari: That is a reasonable design. The main reason I kept them as separate **state()** arrays is that each one drives a different part of the render, _visibleToasts controls which toasts are in the DOM, _dismissingToasts controls the slide out animation class and _collapsingToasts controls the height collapse transition. With separate arrays Lit can re render only the affected subset when any one changes. A single merged array with a status field would also work but any status transition would trigger a re render of the entire array since Lit compares arrays by reference. Happy to refactor in a follow up if you prefer the simpler data model, just wanted to flag the tradeoff.

apps/pie-storybook/stories/testing/pie-toast-provider.test.stories.ts
nindyahapsari: yeah I agree

packages/components/pie-toast-provider/src/toast-provider.scss
nindyahapsari: agreed and done

packages/components/pie-toast-provider/src/index.ts
nindyahapsari: agreed and removed

packages/components/pie-toast-provider/src/defs.ts
nindyahapsari: added

packages/components/pie-toast-provider/src/index.ts
raoufswe: ah nice

raoufswe (approved):

PR #3157, opened 2026-08-27
fix(pie-toast-provider): WAI-558 announce toast messages to SR

dandel10n:
Fixes the accessibility issue behind [WAI-558](https://justeattakeaway.atlassian.net/browse/WAI-558): toast messages are not announced by screen readers (WCAG 4.1.3 Status Messages).

**Root cause:** `pie-toast` renders `role="status"`/`role="alert"` inside its own (nested) shadow root, and the provider only renders `<pie-toast>` once a toast exists. A live region created together with its content is not reliably announced — the region must already be in the DOM when the content changes.

**Changes**
- **`pie-toast-provider`** — renders a single, persistent, visually-hidden ARIA live region inside the always-present provider container. When a toast is created its message is written into that region, so it is announced automatically without focus moving. `error` toasts use an assertive region (`role="alert"` / `aria-live="assertive"`); all other variants use a polite one (`role="status"` / `aria-live="polite"`).
- **`pie-toast`** — the `aria` prop gains a `live` option (`"off" | "polite" | "assertive"`) that overrides the live region politeness. The provider passes `aria.live: "off"` to the toast it renders, so the toast keeps its `role` for semantics but its own announcement is switched off, preventing the message being announced twice. Default behaviour is unchanged (politeness is implied by the role), and a standalone `pie-toast` used without the provider keeps its own live region — so this is backwards compatible.

- [-] N/A I have reviewed visual test updates properly before approving.

Prod:

https://github.com/user-attachments/assets/f072a990-28fa-48f0-aee6-728a8a0c6964

Feature branch:

https://github.com/user-attachments/assets/e7cce4fd-63b3-471d-9700-f14a2ddd040d

### Reviewer 1 @raoufswe 

### Reviewer 2


packages/components/pie-toast/src/index.ts
```diff
@@ -338,7 +357,7 @@ export class PieToast extends PieElement implements ToastProps {
 
         return html`
             <div
-                role="${variant === 'error' ? 'alert' : 'status'}"
+                role="${ifDefined(this.getLiveRegionRole())}"
```
raoufswe: do you think we can need to add the following here as well like the provider 

```
 aria-live="${isError ? 'assertive' : 'polite'}"
 aria-atomic="true"
```

packages/components/pie-toast/src/defs.ts
```diff
@@ -60,6 +60,13 @@ export interface ToastProps {
    * The ARIA labels used for various parts of the toast.
    */
   aria?: AriaProps;
+
+  /**
+   * When true, the toast does not expose its own ARIA live region (`role`).
+   * Used by `pie-toast-provider`, which owns a single persistent live region
+   * for announcements, to prevent the same message being announced twice.
+   */
```
raoufswe: what do you think of?
```suggestion
  /**
   * When true, the toast does not set its own ARIA live region (`role`).
   * Used for cases like `pie-toast-provider`, which owns its own live region
   * for announcements, to prevent the same message being announced twice.
   */
```

packages/components/pie-toast-provider/README.md
```diff
@@ -55,6 +56,12 @@ This component does not have any slots. All content is controlled through proper
 |-----------------------------------|--------------------------------------------------|
 | `pie-toast-provider-queue-update` | Triggered when a toast is added or removed from the queue. |
 
+### Accessibility
+
+The provider renders a single persistent ARIA live region that is always present in the DOM, even when no toast is visible. When a toast is created, its message is written into this region so screen readers announce it automatically, without focus needing to move (WCAG 4.1.3 Status Messages). `error` toasts use an assertive region (`role="alert"`); all other variants use a polite region (`role="status"`).
+
+Because the provider owns this announcement, the toast it renders has its own live region suppressed (via `pie-toast`'s `suppressLiveRegion` prop) to avoid the message being announced twice. A standalone `pie-toast` used without the provider keeps its own live region.
```
raoufswe: i think this is more of implementation details and might not be very useful to consumers and the comments we have the code is sufficient imo

apps/pie-storybook/stories/testing/pie-toast.test.stories.ts
```diff
@@ -78,6 +78,13 @@ const toastStoryMeta: ToastStoryMeta = {
             description: 'The ARIA labels used for various parts of the toast.',
             control: 'object',
         },
+        suppressLiveRegion: {
+            description: 'When true, the toast does not expose its own ARIA live region (used by pie-toast-provider).',
```
raoufswe: i think we can omit used by pie-toast provider here so it isn't tied to specific use cases

fernandofranca (approved):

raoufswe (approved):

PR #3171, opened 2026-09-02
feat(pie-tooltip): DSW-4306 basic tooltip styles and markup

jamieomaguire:
- Adds the initial tooltip component. No interactivity yet, just the styles and markup rendering.
- The bulk of the tooltip logic is related to anchoring it relative to the trigger element
- Some initial a11y testing has been done, but there will be a dedicated ticket later on to fully ensure quality

https://github.com/user-attachments/assets/190d7615-a4ac-4df7-b69b-9a7d00bd8dcb

Screenshot 2026-09-04 at 17 55 52

Screenshot 2026-09-04 at 17 56 08

Screenshot 2026-09-04 at 17 56 23

Screenshot 2026-09-04 at 17 56 46

- [ ] If a component was added or its props were updated, I provided the necessary updates for the Figma Code Connect file
---

### Reviewer 1
- [-] I have reviewed the `PIE Storybook`/`PIE Docs` PR preview.
- [-] I have verified that all acceptance criteria for this ticket have been completed.
- [-] I have reviewed the Aperture changes (if added)


packages/components/pie-tooltip/src/index.ts
```diff
@@ -1,28 +1,407 @@
...
+        const containerInlineSize = container ? container.getBoundingClientRect().width : width;
+
+        this.style.setProperty('--tooltip-container-inline-size', `${containerInlineSize}px`);
+        this._isPositioned = true;
+    }
+
+    private handleCloseButtonClick (): void {
+        // String literal required: CEM cannot resolve constants, so a named constant here would
+        // generate a spurious event entry and a broken React prop name.
+        this.dispatchEvent(new Event('pie-tooltip-close', { bubbles: true, composed: true }));
```
fernandofranca: For consistency, perhaps use the `ON_TOOLTIP_CLOSE_EVENT` constant from defs

fernandofranca: Should we use `dispatchCustomEvent` since this is not a native event?

In AGENTS.md there's a recommendation:
> If you need a PIE custom event API, use `dispatchCustomEvent`...

packages/components/pie-tooltip/src/tooltip.scss
```diff
@@ -1,7 +1,369 @@
...
+    @each $alignment in $tooltip-alignments {
+        .c-tooltip--left#{$alignment} .c-tooltip-arrow,
+        .c-tooltip--right#{$alignment} .c-tooltip-arrow {
+            transform: scaleX(-1);
+        }
+    }
+}
+
+.c-tooltip-heading {
+    @include p.font-theme('font-body-s-strong');
```
fernandofranca: I believe the correct one is `font-body-strong-s`

packages/components/pie-tooltip/src/index.ts
```diff
@@ -1,28 +1,407 @@
...
+        }
+    }
+
+    public disconnectedCallback (): void {
+        this.stopTrackingTrigger();
+        super.disconnectedCallback();
+    }
+
+    // Listens for scroll (capture — catches any ancestor), resize, and dir-attribute changes to
+    // re-anchor the panel. Multiple events per frame coalesce into one rAF.
```
fernandofranca: > Multiple events per frame coalesce into one rAF.

For clarity sake, it might be worth expanding `rAF` to `requestAnimationFrame`

packages/components/pie-tooltip/src/index.ts
jamieomaguire: @fernandofranca I wrote in the comment there but I couldn't get it to work when using the constant. For some reason it just didn't work 🤔 It seemed to cause an issue in the `react.ts` file. I think its because its a native rather than custom event.

I noticed in accordion we also use a raw string value and I wondered if it was related to the same issue? 
Screenshot 2026-09-08 at 11 04 47

jamieomaguire: I think we should stick with native events unless there is a real need to contain some kind of data in the event itself. We have discussed this previously in the team and I think that AGENTS.md file might need updating.

packages/components/pie-tooltip/src/tooltip.scss
jamieomaguire: Good spot - updating

packages/components/pie-tooltip/src/index.ts
jamieomaguire: Ok - I took a look at our agents file and I think we're fine to use a native event here:

```
#### Event Handling

Prefer native platform events where possible (for example `Event`/`InputEvent`) and avoid custom events unless there is a concrete requirement for a custom payload shape or cross-boundary behaviour.

- If a native event is sufficient, dispatch/forward the native event.
- If you need a PIE custom event API, use **`dispatchCustomEvent(element, 'pie-event-name', detail)`** from `pie-webc-core` rather than constructing `CustomEvent` manually.
- If you need to forward a non-composed native event out of shadow DOM, use **`wrapNativeEvent(event)`** from `pie-webc-core`.
```

I'll resolve this but if you have any preference for a custom event please let me know and I can unresolve!

jamieomaguire: Ok so this was interesting. Its a bug in the CEM analyser package. It currently doesn't expect events to be passed as variables, just raw strings. 

I have raised a PR with a proposed fix [here](https://github.com/open-wc/custom-elements-manifest/pull/341) but for now, we can add an `@ignore` jsdoc comment above the dispatch call. This allows us to pass the constant and it works just fine. The manifest file will generate the event name from the jsdoc comment at the top of the class instead of inferred via this call 🤝

fernandofranca (approved):

packages/components/pie-tooltip/src/tooltip.scss
```diff
@@ -1,7 +1,369 @@
...
+// Grid so panel alignment uses logical start/center/end. A panel wider than its track overflows
+// evenly from center, keeping it centred on the trigger.
+.c-tooltip-layer {
+    position: absolute;
+
+    display: grid;
+    grid-template-columns: 100%;
+    grid-template-rows: 100%;
+    visibility: hidden;
+    transition: visibility 0s linear 150ms;
```
raoufswe, line 66:
```suggestion
    transition: visibility 0s linear var(--dt-motion-timing-150);
```

packages/components/pie-tooltip/src/tooltip.scss
```diff
@@ -1,7 +1,369 @@
...
+    --tooltip-inline-size: 280px;
+    --tooltip-max-inline-size: calc(100vw - var(--dt-spacing-d) * 2);
+    --tooltip-background-color: var(--dt-color-container-inverse);
+    --tooltip-content-color: var(--dt-color-content-inverse);
+    --tooltip-border-radius: var(--dt-radius-rounded-b);
+    --tooltip-padding-block: var(--dt-spacing-b);
+    --tooltip-padding-inline: var(--dt-spacing-c);
+    --tooltip-close-size: 32px;
+    --tooltip-motion-duration: 150ms;
+    --tooltip-motion-easing: cubic-bezier(0.45, 0, 0.55, 1);
```
raoufswe, line 136: i believe we already have pie motion tokens for these cases

PR #3172, opened 2026-09-03
ci(pie-monorepo): DSW-123 remove unused changeset code

siggerzz:
When we first introduced Changesets, logic was introduced to handle publishing of `beta` / `next` tagged npm packages when changes were merged into a `beta-*` or `feature-*` branch.

This never got adopted as part of our process, partially due to the introduction of `/snapit`.

As you can see from this screenshot of all our Version Package merges, all these steps get skipped and have never been executed.
Screenshot 2026-09-03 at 10 39 22

No change in behaviour expected as part of this change


.github/workflows/changeset-release.yml
```diff
@@ -69,64 +68,18 @@ jobs:
         EOF
     # Build Packages
     - name: Build
-      if: github.event_name != 'pull_request' &&
```
siggerzz, line 72: Not needed since we gate execution at the job-level

.github/workflows/changeset-release.yml
```diff
@@ -69,64 +68,18 @@ jobs:
...
-      with:
-        files: ".changeset/pre.json"
-    - name: exit prerelease mode
-      # If .changeset/pre.json does not exist and we did not recently exit
-      # prerelease mode.
-      if: steps.check_files.outputs.files_exists == 'true' && contains(github.ref_name, 'main')
-      run: yarn exec changeset pre exit
     - name: Create latest release PR
       id: changesets-main
-      if: contains(github.ref_name, 'main')
```
siggerzz, line 89: ditto

ashleynolan:
Does this remove beta and feature releases in the future? Main reason I ask is that we have used that workflow for tokens for rebrands, so just wanted to check how we would do that in future, or if the capability is still there?

raoufswe (approved):

siggerzz:
@ashleynolan It does yeah. However, in practice, this would never have worked in the repo's current state, because the job was always gated to only publish on `main`. All this logic is leftover implementation from the introduction of Changesets that we never adopted as part of our process.

I think the best option for scenarios like you mentioned, would be for all PR's to merge into a temporary branch, and use `/snapit` as new changes get merged into it. We could even think about updating `/snapit` to provide a tag override, but it's so rare we do this, i'd rather not solutionise

JoshuaNg2332 (approved):

ashleynolan (approved):

